PR #2, opened 2015-11-25
Project submitted

jakebrady5:


bin/tictactoe
```diff
@@ -1,3 +1,54 @@
 #!/usr/bin/env ruby
-
+require 'pry'
```
aviflombaum: You don't need to require pry, it's in the Gemfile and it's being loaded by `config/environment`

bin/tictactoe
```diff
@@ -1,3 +1,54 @@
...
+    player1 = Computer.new("X")
+    player2 = Human.new("O")
+    game = Game.new(player1, player2)
+    game.play
+  else
+    puts "Invalid entry. Please type either human or computer."
+    who_first?
+  end
+end
+
```

`TicTacToeCLI.new`

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```
aviflombaum: Why not just call `reset!` here and use that logic instead of repeating what "resetting" a game does?

Try to rely on your abstractions like methods instead of literal logic.

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    cells[spot.to_i - 1]
+  end
+
+  def full?
+    !(cells.include?(" "))
+  end
+
+  def turn_count
+    turns = cells.delete_if{|a| a == " "}
+    turns.count
```
aviflombaum: [`Array#count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) accepts a block for unique counts and you could write:

```ruby
cells.size - cells.count{|c| c == " "}
```

or

```ruby
cells.count{|c| ["X", "O"].include?(" ")}
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def taken?(spot)
+    index = spot.to_i - 1
+    cells[index] != " "
```
aviflombaum: I would do this within the `cells` index as `cells[spot.to_i - 1] != " "`

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def valid_move?(spot)
+    ok = ["1","2","3","4","5","6","7","8","9"]
+    if ok.include?(spot) && (taken?(spot) == false)
+      return true
+    else
+      false
+    end
+  end
+
```
aviflombaum: I'd re-write as:

```ruby
def valid_move?(spot)
  (1..9).to_a.include?(spot.to_i) && !taken?(spot)
end
```

Or

```ruby
def valid_move?(spot)
  spot.to_i.between?(1,9) && !taken?(spot)
end
```

Try to use the fact that a method with no code executed will return nil by default to not have to explicitly return false. 

And then if you can avoid a local variable, that's better. We want to maintain as little state as possible.

And then, use booleans that implicitly return true/false and avoid if statements. This style makes the method more declarative, it is the definition of the logic, not the implementation.

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+      @player_1 = player_1
+      @player_2 = player_2
+      @board = board
+      @winner = nil
+    end
+
+    def current_player
+      turn_counter = board.cells.count{|a| a != " "}
```
aviflombaum: Why not use `Board#turn_count` here instead of re-implementing your logic? Ask the objects questions, that's why we build methods, so our other code is simpler as the object is an encapsulation of the logic.

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+      if turn_counter % 2 == 0
+        return @player_1
```
aviflombaum: Try not to use the `return keyword`

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+      cb = board.cells
+      WIN_COMBINATIONS.each do |a|
+        if (cb[a[0]] != " ") && (cb[a[0]] == cb[a[1]] && cb[a[0]] == cb[a[2]])
+          @winner = cb[a[0]]
+          return true
+        end
+      end
+      return false
+    end
+
```
aviflombaum: This is good, I'd consider this:

```ruby
def won?
  WIN_COMBINATIONS.each do |a|
    @winner = board.cells[a[0]] if (board.cells[a[0]] != " ") && 
                                   (board.cells[a[0]] == board.cells[a[1]] && 
                                    board.cells[a[0]] == board.cells[a[2]])
  end

  return @winner
end
```

1. Don't create local variables, use `board.cells`. One reason is you could simplify your `Board` instances by building something like `Board#position` or even `Board#[]` that creates a short hand accessor for the value of the board's cell at a position.

```ruby
class Board
  def [](index)
    self.cells[index]
  end

  def position(index)
    self.cells[index]
  end
end
```

Then you could write:

```ruby
def won?
  WIN_COMBINATIONS.each do |a|
    @winner = board[a[0]] if (board[a[0]] != " ") && 
                                   (board[a[0]] == board[a[1]] && 
                                    board[a[0]] == board[a[2]])
  end

  return @winner
end
```

Pretty cool that you can implement your own `Object#[]` and make your own objects have an array like interface, right? :heart: ruby.

And then instead of returning true or false, I'm returning `@winner`. I, and I alone, like a `return` keyword for the last line of the method when I'm returning a somewhat unrelated value like an instance variable or a local variable. But the idea is that either `@winner` was set correctly or it wasn't set at all in which case the method returns nil.

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+    def draw?
+      board.cells.count{|a| a != " "} == 9 && won? == false
```
aviflombaum: Why not use `Board#full?` here?

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    def winner
+      won?
```
aviflombaum: See if `won?` returned @winner it would either be the token or nil. Return objects as they are implicitly true and can be used in conditionals.

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+      @winner
+    end
+
+    def turn
+      board.display
+      spot = current_player.move(board)
+      if valid_move?(spot)
+        board.cells[spot.to_i - 1] = current_player.token
```
aviflombaum: Why not use `Board#update` here?

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+    def valid_move?(spot)
+      spot = spot.to_i - 1
+      if position_taken?(spot) == false && spot.between?(0,8) == true
+        true
+      else
+        false
+      end
+    end
+
+    def position_taken?(spot)
```
aviflombaum: Don't reimplement - we have `Board#taken?` no?

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+      else
+        puts "invalid entry"
+        turn
+      end
+
+    end
+
+    def valid_move?(spot)
+      spot = spot.to_i - 1
+      if position_taken?(spot) == false && spot.between?(0,8) == true
```
aviflombaum: Why not use `Board#valid_move?`

Use the methods from your other objects :-)

lib/players/computer.rb
```diff
@@ -0,0 +1,91 @@
...
+      spot = go_here.select{|a| board.cells[a] == " "}
+      spot = spot[0] + 1
+      return spot
+    else
+      nil
+    end
+  end
+
+  def default_move(board)
+  possible_moves = ["1","2","3","4","5","6","7","8","9"]
```
aviflombaum: Should be indented.

lib/players/computer.rb
```diff
@@ -0,0 +1,91 @@
...
+  spot
+  end
+end
+
+
+
+
+
+
+
```
aviflombaum: I'd like to see you refactor this class. Try to break things into smaller methods.

aviflombaum:
@jakebrady5 I left some notes on your code, let's schedule a time to chat about it! Great job!

PR #3, opened 2015-11-26
Done

onyoo:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+require_all '../lib'
+require_all '../lib/players'
```
aviflombaum: These lines shouldn't be needed, they should be covered by the `config/environment`.

bin/tictactoe
```diff
@@ -1,3 +1,30 @@
...
+
+elsif no_players == 2
+  Game.new(Human.new("X"), Human.new("O")).play
+
+else
+  p "You've selected an AI only game."
+  Game.new(Player::Computer.new("X"), Player::Computer.new("O")).play
+
+end
+
```

`TicTacToeCLI.new`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+
+class Board
+  attr_accessor :cells
+
+
+  def initialize
+    @cells = [" "] * 9
```
aviflombaum: Why not just call the `reset!` method you define below here? It's always better to rely on an abstraction like a method call than literal logic.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def reset!
+    @cells.clear
+    @cells = [" "] * 9
```
aviflombaum: I think you only need LOC12, not `@cells.clear`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
+    @cells.all?{|x| x != " "}
+  end
+
+  def turn_count
+    @cells.select{|x| x != " "}.length
```
aviflombaum: You can just use the block format of `count` here.

`@cells.count{|x| x != " "}`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def taken?(pos_str)
+    @cells[pos_str.to_i - 1] == " " ? false : true
+  end
+
+  def valid_move?(pos)
+    !taken?(pos) && 0 < pos.to_i && pos.to_i < 10
```
aviflombaum: You can use `pos.to_i.between?(1,9)` too

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def update(pos, player)
+    if valid_move?(pos)
+      @cells[pos.to_i - 1] = player.token
+    end 
+  end
```
aviflombaum: I think you can use an end-of-line if statement to make this method read a bit more declaratively. 

```ruby
def update(pos, player)
  @cells[pos.to_i - 1] = player.token if valid_move?(pos)
end
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+    self.board.turn_count % 2 == 0 ? player_1 : player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def won? 
+    WIN_COMBINATIONS.detect do |win_array|
+############ Try to insert PROC?? ###########
```
aviflombaum: Not sure what this comment is about but your code is great. Good use of `detect` and `all`

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+    if draw?
+      nil
+############ Try to insert PROC?? ###########
+    elsif WIN_COMBINATIONS.detect{|win_array| win_array.all?{|num| self.board.cells[num] == "X"}}
+      "X"
+    elsif WIN_COMBINATIONS.detect{|win_array| win_array.all?{|num| self.board.cells[num] == "O"}}
+      "O"
+    end
+  end
+
```
aviflombaum: Here's a cleaner implementation of `winner`. You want to avoid duplicating the logic of your `won?` method and instead use it.

```ruby
  def winner
    if winning_combo = won?
      board.cells[winning_combo.first]
    end
  end
```

Respond to this comment explaining how that works.

lib/player.rb
```diff
@@ -0,0 +1,17 @@
+
+class Player
+  attr_reader :token
+
+  def initialize(token)
+    @token = token
+  end
+end
+
+class Human < Player
```
aviflombaum: Why not define the `Human` player in it's own file in `lib/players/human.rb`?

aviflombaum:
@onyoo @frankNowinski great job on this! I left some comments, one of which I'd love you to respond to. Want to schedule time to discuss the code?

lib/game.rb
onyoo: Thanks for the comment, it is much cleaner. If we change the suggested code to:
```
def winner


our suggestion works like this:
1
2 if a winning combination matches the combination detected (and returned) by the won? method, it will

Comment (onyoo) [inline lib/game.rb]: Thanks for the comment, it is much cleaner. If we change the suggested code to include "==" :
```
def winner
  board.cells[winning_combo.first] if winning_combo == won?
end
```
your suggestion would work like this:
If a winning combination matches the combination detected (and returned) by the #won? method, #winner will return the value of the first cell in the winning combination on the current instance of board.

It is way cleaner and more abstract.

onyoo: Line 2 assigns the  'winning_combo' variable to the detected winning combination from the '#won?' method (and starts a conditional block). Line 3 takes the instance of the board, and gets the first element of the 'winning_combo' from the board. That is, either "X" or "O".

It eliminates the repetitious code.

PR #5, opened 2015-11-27
Project submitted

frankNowinski:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
...
+    game.play
+    TicTacToeCLI.new if play_again? == 'y'
+  end
+
+  def play_again?
+    puts "\nEnter 'y' if you would like to play another game of Tic Tac Toe."
+    answer = gets.strip.downcase
+  end
+end
+
```
aviflombaum: That's great! I'd just move this to it's own class in `lib/tic_tac_toe_cli.rb` which will be loaded by environment so this file can just have the line below.

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |combo|
+      if (board.cells[combo[0]] == board.cells[combo[1]] && 
+        board.cells[combo[1]] == board.cells[combo[2]]) && 
+        board.cells[combo[0]] != " "
+
+        @winner_token = board.cells[combo[0]]
+        return true
```
aviflombaum: Use `break` here to stop the loop.

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+    false
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    won?
+    @winner_token
```
aviflombaum: `won? && @winner_token` is a little nicer

PR #8, opened 2015-11-29
Done

KhalilC:


bin/tictactoe
```diff
@@ -1,3 +1,48 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+require 'pry'
```
aviflombaum: This is already required by the Gemfile and not needed here. If anything it also would go in `config/environment.rb` which is responsible for loading libraries and dependencies.

bin/tictactoe
```diff
@@ -1,3 +1,48 @@
...
+    game = Game.new
+  end
+  game.play
+  begin
+    puts "Play again?(y/n)"
+    answer = gets.chomp.downcase
+  end until %w{y n}.include?(answer)
+  break if answer == 'n'
+end
+puts "Thanks for playing!  Goodbye!"
```

`TicTacToeCLI.new`

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
aviflombaum: Why not use `self.reset!

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def full?
+    cells.all? { |position| position != " " }
+  end
+
+  def turn_count
+    count = 0
+    cells.each { |position| count += 1 if position != " "}
+    count
```
aviflombaum: You can just use the block format of count here.

`cells.count{|x| x != " "}`

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def taken?(position)
+    cells[position.to_i - 1] != " "
+  end
+
+  def valid_move?(move)
+    options = (1..9).map {|number| number.to_s}
```
aviflombaum: `("1".."9").to_a`

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    options.include?(move) && !taken?(move)
```
aviflombaum: Better, I'd just say `move.to_i.between?(1,9) && !taken?(move)`

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def initialize(player_1=Human.new('X'), player_2 = Human.new('O'), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    return player_1 if board.turn_count % 2 == 0
+    return player_2 if board.turn_count % 2 != 0
```
aviflombaum: `board.turn_count.even? ? "X" : "O"`

lib/players/computer.rb
```diff
@@ -0,0 +1,36 @@
...
+    available_moves = (1..9).select { |number| !board.taken?(number) } 
+    return win_or_stop_win(board).to_s if win_or_stop_win(board)
+    return best_moves(board, available_moves).to_s
+  end
+
+  def win_or_stop_win(board)
+    marker = self.token
+    opponent = ['X', 'O']
+    opponent.delete(marker)
+    Game::WIN_COMBINATIONS.each do |combo|
```
aviflombaum: I'd try to refactor these decisions into individual methods.

KhalilC: One thing I didn't notice with including winning and blocking opponent winning in the same method, is that it causes the computer to make the wrong decision in a certain scenario.  If the computer can win on the next move, but the opponent can also win on the next move, it will choose to block the opponent from winning if the opponent's winning combo comes first in the array before the computer's winning combo.  Putting them in their own methods fixes this, because then it will only check for a winner first, and then proceed to blocking if no win is available.

PR #11, opened 2015-12-04
Done

Louiefigz:


bin/tictactoe
```diff
@@ -1,3 +1,94 @@
...
+# player_2= Player::Computer.new("O")
+
+# game = Game.new(player_1, player_2)
+
+# board = Board.new
+
+# board.display
+
+
+game.play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end 
+
+  def position(input)
+    # binding.pry
+      if input.to_i > 0 && input.to_i < 10
```
randallreedjr: You can use [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F) to simplify your logic. (between is inclusive)
`if input.between?(1,9)`

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    end 
+  end
+
+  def full?
+    if cells.include?(" ")
+      false
+    else 
+      true
+    end 
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding an if statement. You can use the bang operator `!` to negate or revers the results. 
```
def full?
  !cells.include?(" ")
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    position(input).include?("O") || position(input).include?("X") 
+  end 
+
+  def valid_move?(input)
+    if input.to_i > 0
+      !taken?(input) 
+    else
+      false 
+    end 
+  end
```
randallreedjr: Again, using the return values of boolean expressions, this can be simplified to
```
def valid_move?(input)
  input.to_i > 0 && !taken?(input)
end
```
But, what happens if `input` is 10...?

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+    @player_2 = player_2 
+    @board = board
+  end
+
+  def current_player
+    board.turn_count.even? ? player_1 : player_2
+  end 
+
+  def over?
+    board.full?
```
randallreedjr: Can a game end before the board fills up?

lib/players/computer.rb
```diff
@@ -0,0 +1,53 @@
...
+
+
+  def move(board)
+    return "5" if !board.taken?(5)
+    best_option(board)
+  end
+
+  def best_option(board)
+    win_move(board) || block_move(board) || strategic_move(board)
+  end
```
randallreedjr: Nice! This is very expressive and readable.

lib/players/human.rb
```diff
@@ -0,0 +1,15 @@
+class Player::Human < Player 
+
+  attr_accessor :board
+
+  def move(board)
+    puts "where on the board would you like to go? Choose a number from 1-9"
+    input = gets.strip
+  end
```
```
def move(token, board)
  puts "where on the board would you like to go? Choose a number from 1-9"
  gets.strip
end
```

spec/05_computer_player_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'spec_helper'
+wrequire 'spec_helper'
```
randallreedjr: This looks like a typo.

randallreedjr:
@Louiefigz Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #12, opened 2015-12-09
Project submitted

ed-karabinus:


bin/tictactoe
```diff
@@ -1,3 +1,50 @@
...
+  100.times do
+    if i % 2 == 0
+      losses += 1 if Game.new(Player::Computer.new('X'), Player::Random.new('O')).play == 'O'
+    else
+      losses += 1 if Game.new(Player::Random.new('O'), Player::Computer.new('X')).play == 'O'
+    end
+    i += 1
+  end
+  puts 'The computer played a random opponent 100 times and lost ' + losses.to_s + ' times. Goodbye!'
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,71 @@
...
+    position(input) != ' '
+  end
+
+  def turn_count
+    turn_count = 0
+    ('1'..'9').each do |index|
+      turn_count += 1 if taken?(index)
+    end
+    turn_count
+  end
```

```ruby
def turn_count
  (1..9).count{|index| taken?(index)}
end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+    self.board.token_1 = self.player_1.token
+    self.board.token_2 = self.player_2.token
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
+  end
```

`self.board.turn_count.even? ? @player_1 : @player_2`

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
+    if self.won?
+      self.board.cells[self.won?[0]]
+    end
+  end
```

```ruby
if won = won?
  self.board.cells[won[0]]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Player::Human < Player
+  def move(board)
+    puts "#{self.token}, please enter your move as a number 1-9."
+    input = gets.strip
+    input
+  end
```
```
def move(token, board)
  puts "#{self.token}, please enter your move as a number 1-9."
  gets.strip
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,77 @@
...
+    if self.current_player?(board)
+      max_score_index = scores.each_with_index.max[1]
+      self.intelligent_move = moves[max_score_index]
+      return scores[max_score_index]
+    else
+      min_score_index = scores.each_with_index.min[1]
+      self.intelligent_move = moves[min_score_index]
+      return scores[min_score_index]
+    end
+  end
```
randallreedjr: Nice!

randallreedjr:
@ed-karabinus Great job, especially with minimax! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #13, opened 2015-12-09
Project submitted

mmmeyers:


randallreedjr:
@mmmeyers Is your project complete? I don't see a playable bin file.

mmmeyers:
Hi Randall,

It was done, but I don't know why it wasn't all pushed to GitHub when I ran
learn submit. Just pushed it again a few minutes ago - let me know if you
don't see it. Thanks!

Marsi

randallreedjr:
@mmmeyers This looks better, thanks for re-pushing!

bin/tictactoe
```diff
@@ -1,3 +1,45 @@
...
+    input = gets.chomp.downcase
+
+    if input == "y"
+      play_game
+    elsif input == "n"
+      puts "Thanks for playing! Goodbye."
+    end
+  end
+
+ play_game
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def full?
+    cells.all? {|space| space == "X" || space == "O"}
+  end
+
+  def turn_count
+    # cells.count gives you total number of spaces in array (9)
+    cells.count {|space_taken| #gives you total number of spaces taken up with X or O
+      space_taken == "X" || space_taken == "O"}
+  end
```
randallreedjr: Great choice of iterators for both `full?` and `turn_count`

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def taken?(location)
+    location = (location.to_i) - 1
+    cells[location] == nil || cells[location] == " " ? false : true
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. You can reverse (or negate) the boolean with the `!` (bang) operator.

`!(cells[location] == nil || cells[location] == " ")`

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+    @player_2 = player_2
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
+      @player_1
+    else
+      @player_2
+    end
+  end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def move(board)
+    "1" if !board.taken?(1)
+    strategies(board).to_s
+  end
+
+  def strategies(board)
+    winning_move(board, token) || block_move(board, token) || random_move(board)
+  end
```
randallreedjr: Great method, very readable

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+
+class Player::Human < Player
+
+  def move(board)
+    puts "Where would you like to go (choose a number from 1-9)?"
+    input = gets.chomp
+  end
```
```
def move(token, board)
  puts "Where would you like to go (choose a number from 1-9)?"
  gets.chomp
end
```

randallreedjr:
@mmmeyers Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #15, opened 2015-12-15
Project submitted

tuckerbohman5:


bin/tictactoe
```diff
@@ -1,3 +1,23 @@
...
+#     Game.new(Player::Human.new('X'), Player::Computer.new('O')).play
+
+#   when 2
+#     Game.new(Player::Human.new('X'), Player::Human.new('O')).play
+
+#   when 0
+#     Game.new(Player::Computer.new('X'), Player::Computer.new('O')).play
+# end
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,80 @@
+require "pry"
+
+class Board
+
+  attr_accessor :cells, :player, :token_1, :token_2
+
+  def initialize
+    @cells = Array.new(9, " ")
```
randallreedjr: You already have this code in the `reset!` method. No need to duplicate it; just call `reset!` from here.

lib/board.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def turn_count
+    move_count = []
+    self.cells.each do |cell|
+      if cell != " "
+        move_count << cell
+      end
+    end
+    move_count.count
+  end
```

```ruby
def turn_count
  cells.count {|cell| cell != " "}
end
```

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+  end
+#---------------------------------------------------#
+
+  def current_player
+    if board.turn_count.even?
+      player_1
+    else
+      player_2
+    end
+  end
```
randallreedjr: You could simplify the `current_player` method by using the ternary operator.
```
def current_player
  board.turn_count.even? ? player_1 : player_2
end
```

lib/game_status.rb
```diff
@@ -0,0 +1,50 @@
...
+    if @winner == nil
+      return false
+    end
+  end
+
+  def draw?
+        if !won?  && board.full? == true
+            return true
+        end
+    end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the `if` statement and explicit `return`. Also, avoid literal boolean equality comparison. 

```
def draw?
  !won? && board.full?
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,15 @@
+class Player::Human < Player
+
+ def move(board)
+  puts "Please pick your move by entering a location 1-9"
+  @players_pick = gets.chomp
+  @players_pick
+
+ end
```
randallreedjr: You can return the result of an assignment without needing an additional line.
```
def move(board)
  puts "Please pick your move by entering a location 1-9"
  @players_pick = gets.chomp
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,87 @@
...
+      max_score_index = scores.each_with_index.max[1]
+      self.best_move = moves[max_score_index]
+      return scores[max_score_index]
+    else
+      # This is the min calculation
+      min_score_index = scores.each_with_index.min[1]
+      self.best_move = moves[min_score_index]
+      return scores[min_score_index]
+    end
+  end
```
randallreedjr: Awesome!

randallreedjr:
@tuckerbohman5 Great job, especially using minmax! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #17, opened 2015-12-19
Project submitted

dannyd4315:


bin/tictactoe
```diff
@@ -1,3 +1,18 @@
...
+
+  again = gets.strip.downcase
+
+  until again == "n"
+    Start.new
+    puts "Do you want to play again? Enter Y or N"
+    again = gets.strip.downcase
+  end
+  
+  puts "Have a nice day!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end 
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    won? ? board.position(won?[0] + 1) : nil 
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+require 'pry'
+
+class Human < Player
+
+  def move(board)
+    puts "Please enter 1-9:"
+    position = gets.strip
+  end
```
```
def move(token, board)
  puts "Please enter 1-9:"
  gets.strip
end
```

randallreedjr:
@dannyd4315 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #18, opened 2015-12-22
Project submitted

Lukeghenco:


bin/tictactoe
```diff
@@ -1,3 +1,65 @@
...
+
+puts " "
+puts " "
+puts "||||||||  ||||||   ||||||    ||||||||    |||        ||||||    ||||||||   |||||    |||||||||"
+puts "   ||       ||    |||           ||      || ||      |||           ||     ||   ||   ||       "
+puts "   ||       ||   |||            ||     ||   ||    |||            ||    ||     ||  ||       "
+puts "   ||       ||   ||             ||    |||||||||   ||             ||    ||     ||  |||||||||"
+puts "   ||       ||   |||            ||   ||       ||  |||            ||    ||     ||  ||       "
+puts "   ||       ||    |||           ||  ||         ||  |||           ||     ||   ||   ||       "
+puts "   ||     ||||||   ||||||       || ||           ||  ||||||       ||      |||||    |||||||||"
```
randallreedjr: Nice touch!

bin/tictactoe
```diff
@@ -1,3 +1,65 @@
...
+    call
+  elsif input == "N" 
+    exit
+  else 
+    puts 'That s not a valid choice!!'
+    again?
+  end
+end
+
+call
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    !(@cells[position.to_i - 1].nil? || @cells[position.to_i - 1] == " ")
+  end
+
+  def valid_move?(position)
+    if (taken?(position.to_i) == false) && (position.to_i.between?(1,9) == true)
+    true
+    else (taken?(position.to_i) == true) || (position.to_i.between?(1,9) == false)
+    false
+    end
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the `if` statement. 
```
def valid_move?(position)
  (taken?(position.to_i) == false) && (position.to_i.between?(1,9) == true)
end
```
Also, eqaulity comparison to a boolean literal value is never necessary; just use the implicit return value of the expression. Finally, use the bang `!` operator to negate or reverse a logical comparison.
```
def valid_move?(position)
  !taken?(position.to_i) && position.to_i.between?(1,9)
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def full?
+    @cells.all? { |i| i == "X" || i == "O" }
+  end
+
+  def turn_count
+    @cells.count{ |i| i == "X" || i == "O"}
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
+    if self.won?
+      self.board.cells[self.won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,39 @@
+class Player::Computer < Player
+
+  def move(board)
+    return "5" if !board.taken?("5")
+    strategic_move(board)
+  end 
+
+  def strategic_move(board)
+    win(board) || block_move(board) || best_move(board)
+  end
```
randallreedjr: This method does a great job of utilizing other methods in a readable way.

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+  def move(board)
+    puts "#{self.token}, please choose a number between 1-9 to make your move."
+    player_move = gets.strip
+    player_move
+  end
```
```
def move(token, board)
  puts "#{self.token}, please choose a number between 1-9 to make your move."
  gets.strip
end
```

randallreedjr:
@Lukeghenco Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

Lukeghenco:
Oh, cool. I’ll go back and work on it and see how the implementations work. Thanks

PR #21, opened 2015-12-30
Done

Alvinluipo:


bin/tictactoe
```diff
@@ -1,3 +1,38 @@
...
+    game.play
+  elsif input=="ZERO" || "0"
+    board=Board.new
+    game=Game.new(player_1=Player::Computer.new("X"), player_2=Player::Computer.new("O"))
+    game.play
+  else
+    puts "You entered an invalid number, please pick again"
+    start_game
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    @cells[position.to_i-1]
+  end
+
+  def full?
+    @cells.all?{|x| x=="X" || x=="O"}
+  end
+
+  def turn_count
+    @cells.count{|x| x=="X" || x=="O"}
+  end
```
randallreedjr: Great choice of iterators for `full?` and `turn_count`

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
```

randallreedjr:
@Alvinluipo Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #22, opened 2015-12-30
Complete with good but not perfect AI. "Wargame" ends in draws much more often than not.

febbraiod:
I also wonder if I should house my "WIN_COMBOS" constant in a different file and have the two classes I use it in inheriting it.

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -31,7 +31,7 @@ require_relative '../config/environment'
 
     
 
-puts "Greetings Professor? Would you like to play a game?"
+puts "Greetings Professor. Would you like to play a game?"
 
 
 
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,96 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+  
+  puts "                (         )                     (  "                                                 
+  puts "    (        )\\ )   ( /(      (      (       )\\ )  "                                             
+  puts "    )\\ )    (()/(   )\\())   ( )\\     )\\     (()/(  "                                              
+  puts "  (()/(     /(_)) ((_)\    )((_) ((((_)(    /(_))  "                                             
+  puts "   /(_))_  (_))     ((_)  ((_)_   )\\ _ )\\  (_))    "                                            
+  puts "  (_)) __| | |     / _ \\   | _ )  (_)_\\(_) | |     "                                            
+  puts "    | (_ | | |__  | (_) |  | _ \\   / _ \\   | |__   "                                            
+  puts "     \\___| |____|  \\___/   |___/  /_/ \\_\\  |____|  "                                            
+  puts ""
+  puts "               )           *         )       )                   (                      (    "      
+  puts "   *   )   ( /(         (  `     ( /(    ( /(             (     )\\ )           (       )\\ )  " 
+  puts "  ` )  /(   )\\())  (     )\\))(    )\\())   )\\())     (      )\\   (()/(   (       )\\     (()/( "   
+  puts "   ( )(_)) ((_)\\   )\\   ((_)()\\  ((_)\\   ((_)\\      )\\   (((_)   /(_))  )\\   ((((_)(    /(_))" 
+  puts "  (_(_())   _((_) ((_)  (_()((_)   ((_)   _((_)  _ ((_)  )\\___  (_))   ((_)   )\\ _ )\\  (_))  "  
+  puts "  |_   _|  | || | | __| |  \\/  |  / _ \\  | \\| | | | | | ((/ __| | |    | __|  (_)_\\(_) | _ \\ "   
+  puts "    | |    | __ | | _|  | |\\/| | | (_) | | .` | | |_| |  | (__  | |__  | _|    / _ \\   |   / "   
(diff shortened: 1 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,96 @@
...
+  puts "    |_|    |_||_| |___| |_|  |_|  \\___/  |_|\\_|  \\___/    \\___| |____| |___|  /_/ \\_\\  |_|_\\ "  
+  puts "            (                                                                 )              "  
+  puts "    *   )   )\\ )     (            *   )     (         (            *   )   ( /(              "   
+  puts "  ` )  /(  (()/(     )\\         ` )  /(     )\\        )\\         ` )  /(   )\\())   (         "  
+  puts "  ( )(_))  /(_))  (((_)   ___   ( )(_)) ((((_)(    (((_)   ___   ( )(_)) ((_)\\    )\\         "  
+  puts "  (_(_())  (_))    )\\___  |___| (_(_())   )\\ _ )\\   )\\___  |___| (_(_())    ((_)  ((_)       "   
+  puts "  |_   _|  |_ _|  ((/ __|       |_   _|   (_)_\\(_) ((/ __|       |_   _|   / _ \\  | __|      "   
+  puts "    | |     | |    | (__          | |      / _ \\    | (__          | |    | (_) | | _|       "   
+  puts "    |_|    |___|    \\___|         |_|     /_/ \\_\\    \\___|         |_|     \\___/  |___|      "  
+
```
randallreedjr: Nice touch!

bin/tictactoe
```diff
@@ -1,3 +1,96 @@
...
+      end
+  elsif response == "n" || response.include?("no")
+    puts "That is too bad. Goodbye Professor."
+    break
+  else
+    puts "That is not a valid choice."
+  end
+
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+                      [2, 4, 6]]
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+                      [0, 3, 6],
+                      [1, 4, 7],
+                      [2, 5, 8],
+                      [0, 4, 8],
+                      [2, 4, 6]]
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
randallreedjr: Looks like you already have this code elsewhere in your class. Here you could just call the `reset!` method.

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+    cells[position.to_i - 1] != " "
+  end
+
+  def valid_move?(user_input)
+    if !(1..9).include?(user_input.to_i)
+      return false
+    end
+
+    !taken?(user_input)
+  end
```
randallreedjr: Use logical and operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
(1..9).include?(user_input.to_i) && !taken(user_input)
```

lib/game.rb
```diff
@@ -0,0 +1,123 @@
...
+        return win
+      end
+    end
+
+    false
+  end
+
+  def draw?
+    over? && !won? ? true : false
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. 

`over? && !won?`

lib/game.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def over?
+
+    if board.turn_count == 9
+      return true
+    else
+      return false
+    end
```
randallreedjr: Use the implicit return value of a boolean expression instead of adding the `if` statement.
```
def over?
  board.turn_count == 9
end
```

lib/game.rb
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  def winner
+    if won? 
+      winning_line = won?
+      return board.cells[winning_line[0]]
+    end
+
+    nil
+  end
```

```ruby
if winning_line = won?
  board.cells[winning_line[0]]
end
```

Additionally, the `if` statement will return `nil` if the condition is not met, so a separate `nil` return value is not necessary.

lib/players/human.rb
```diff
@@ -0,0 +1,14 @@
+require_relative '../player.rb'
+
+class Human < Player
+
+  def move(token, board)
+    puts "Where would you like to move?"
+    move = gets.chomp
+    move
+  end
```
```
def move(token, board)
  gets.chomp
end
```

randallreedjr:
@febbraiod Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

febbraiod:
Hey thanks!  If I were to make these adjustments, should I resubmit?

randallreedjr:
Only if you would like to. If you do choose to resubmit, one of our team of
reviewers would be happy to provide updated feedback.

febbraiod:
excellent. thanks again!

PR #23, opened 2016-01-01
Working!

matthewpaulmcgowan:
working!

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+      
+Game.new.start  
--- a/lib/board.rb
+++ b/lib/board.rb
@@ -53,11 +53,10 @@ class Board
         return false #is the catch for invalid input
     end
     
-    !taken?(input) #reverse the response of taken to determine if it is a valid move
+    !taken?(input) #determines if the input is not taken
    end
    
    def update(input, player) #updates the @cells to the token of the player
       @cells[input.to_i-1]=player.token
    end
-     
 end
--- a/lib/board.rb
+++ b/lib/board.rb
@@ -1,62 +1,189 @@
 class Board
-    attr_accessor :cells
-
-   
-   
-    def reset! #resets the @cells which empties the board
-        @cells=[" "," "," "," "," "," "," "," "," "]
-    end
+  WIN_COMBINATIONS=[[0,1,2],[3,4,5],[6,7,8],[0,4,8],[2,4,6],[1,4,7],[0,3,6],[2,5,8]] #defines all of the possible combinations to win
+  attr_accessor :cells
+  def reset! #resets the @cells which empties the board
+    @cells=[" "," "," "," "," "," "," "," "," "]
+  end
     
-   def cells #is the board 
-       @cells
-   end
+  def cells #is the board 
+    @cells
+  end
    
-   def initialize # resets and creates a new board for a new game
-      self.reset!
-   end
+  def initialize # resets and creates a new board for a new game
+    self.reset!
+  end
    
-   def display #displays the board currently in cells
+  def display #displays the board currently in cells
     print " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} \n"
     print "-----------\n"
     print " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} \n"
     print "-----------\n"
     print " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
-   end
+  end
    
-   def position (input) #returns the contents of the selected cell
-       return @cells[input.to_i-1]
-   end
+  def position(input) #returns the contents of the selected cell
+    return @cells[input.to_i-1]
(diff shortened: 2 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+      
+Game.new.start
```
randallreedjr: Great bin file, very short exactly as expected. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Separating the interface from the implementation of the game allows for multiple interfaces (e.g. a web app).

lib/board.rb
```diff
@@ -0,0 +1,189 @@
...
+  attr_accessor :cells
+  def reset! #resets the @cells which empties the board
+    @cells=[" "," "," "," "," "," "," "," "," "]
+  end
+    
+  def cells #is the board 
+    @cells
+  end
+   
+  def initialize # resets and creates a new board for a new game
```
randallreedjr: Generally, the `initialize` method is placed at the top of the class.

lib/board.rb
```diff
@@ -0,0 +1,189 @@
+class Board
+  WIN_COMBINATIONS=[[0,1,2],[3,4,5],[6,7,8],[0,4,8],[2,4,6],[1,4,7],[0,3,6],[2,5,8]] #defines all of the possible combinations to win
+  attr_accessor :cells
+  def reset! #resets the @cells which empties the board
+    @cells=[" "," "," "," "," "," "," "," "," "]
+  end
+    
+  def cells #is the board
```
randallreedjr: With `attr_accessor :cells`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,189 @@
+class Board
+  WIN_COMBINATIONS=[[0,1,2],[3,4,5],[6,7,8],[0,4,8],[2,4,6],[1,4,7],[0,3,6],[2,5,8]] #defines all of the possible combinations to win
+  attr_accessor :cells
+  def reset! #resets the @cells which empties the board
+    @cells=[" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,189 @@
...
+  def position(input) #returns the contents of the selected cell
+    return @cells[input.to_i-1]
+  end
+   
+  def full? #determines if the board is full
+    !@cells.include?(" ") 
+  end
+   
+  def turn_count #returns the amount of turns in the game
+    return @cells.count {|cell| cell!=" "}
```
randallreedjr: Excellent choice of iterator.

lib/board.rb
```diff
@@ -0,0 +1,189 @@
...
+  end
+   
+  def taken?(input) #determines of the input cell (changed) is already taken or not
+    input=input.to_i-1
+    if @cells[input]!=" "
+      return true
+    else
+      return false
+    end
+  end
```
randallreedjr: Prefer implicit return values of boolean statements.
```
def taken?(input)
  input=input.to_i-1
  @cells[input]!=" "
end
```

lib/board.rb
```diff
@@ -0,0 +1,189 @@
...
+   
+  def valid_move?(input) #determines if a given number is a valid number and if the position is taken
+    input.split.all? {|local_input| local_input.to_i<10 && local_input.to_i>0} #turn the input into an array so that i can compare with a number, was getting nomethod error "<" previously
+    if taken?(input) #determines if the input is not taken
+      return false #if the input is taken, then it is not a valid move
+    else
+      return true #if the input is not taken, then it could be a valid move
+    end
+  end
```
randallreedjr: Reverse (or negate) the boolean with the `!` (bang) operator.

`!taken?(input)`

lib/game.rb
```diff
@@ -0,0 +1,142 @@
+require 'pry'
+class Game
+  attr_reader :board, :player_1, :player_2, :total_ties
+  WIN_COMBINATIONS=[[0,1,2],[3,4,5],[6,7,8],[0,4,8],[2,4,6],[1,4,7],[0,3,6],[2,5,8]] #defines all of the possible combinations to win
+   
+  def board=(board) 
+    @board=[]
```
randallreedjr: This method isn't using the parameter `board` being passed in?

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+  def player_1=(player_1) 
+    @player_1=player_1
+  end
+   
+  def player_2=(player_2)
+    @player_2=player_2
+  end
+  
+  def current_player #determines the current player using turn count from the Board class
+    if board.turn_count%2==0
```

randallreedjr:
@matthewpaulmcgowan Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #25, opened 2016-01-06
Done

jsgoldb:


bin/tictactoe
```diff
@@ -1,3 +1,12 @@
...
 require_relative '../config/environment'
+
+
+play_again = 'y'
+while play_again == 'y'
+    Game.new.start
+    puts "If you would like to play again, enter y."
+    puts "If not, enter any other key."
+    play_again = gets.chomp
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def update(pos, player)
+    cells[pos.to_i - 1] = player.token
+  end
+
+  def full?
+   cells.detect {|cell| cell == " " || cell == nil} ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def turn_count
+    if cells.select {|cell| cell == "O" || cell == "X"}
+      return cells.select {|cell| cell == "O" || cell == "X"}.count
+    else
+      return 0
+    end
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def taken?(pos)
+    position(pos) != " " && position(pos) != nil ? true : false
+  end
```

`position(pos) != " " && position(pos) != nil`

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def valid_move?(pos)
+    if (1..9).include?(pos.to_i) && taken?(pos) == false
+      return true
+    else 
+      false
+    end
+  end
```
randallreedjr: Prefer implicit return values of boolean statements instead of adding `if` statement. Use bang operator `!` to reverse (or negate) results instead of comparing equality to `false`.
```
def valid_move?(pos)
  (1..9).include?(pos.to_i) && !taken(pos)
end
```

lib/game-check.rb
```diff
@@ -0,0 +1,40 @@
...
+  def won?
+    @winner = WIN_COMBINATIONS.detect do |combo|
+      (board.cells[combo[0]] == 'X' && board.cells[combo[1]] == 'X' && board.cells[combo[2]] == 'X') ||
+      (board.cells[combo[0]] == 'O' && board.cells[combo[1]] == 'O' && board.cells[combo[2]] == 'O')
+    end
+    @winner ? true : false
+  end
+
+  def winner
+    won? ? board.cells[@winner[0]] : nil
```
randallreedjr: Prefer inline `if` statement to ternary with nil value
```
def winner
  board.cells[@winner[0]] if won?
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,80 @@
...
+  def reset_space(space)
+    board.cells[space] = " "
+  end
+
+  def next_player
+    if current_player == player_1
+      return player_2
+    else
+      return player_1
+    end
```
randallreedjr: This would be a good spot to use the ternary operator
`current_player == player_1 ? player_2 : player_1`

randallreedjr:
@jsgoldb Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

jsgoldb:
Thanks for the suggestions Randall, extremely helpful and instructive.

PR #26, opened 2016-01-06
Done

plai217:
AI not the best

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,31 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts 'Welcome to TicTacToe'
+
+players = ''
+first = ''
+until players == '0' || players == '1' || players == '2'
+  puts 'Please enter number of 0,1 or 2 player(s)'
+  players = gets.strip
+end
+
+if players == '1'
+  puts 'Would you like to go first? (Y/N)'
+  first = gets.strip.upcase
+  until first == 'Y' || first == 'N'
+    puts 'Please enter Y or N'
+    first = gets.strip.upcase
+  end
+end
+
+if players == '0'
+  Game.new(Player::Computer.new('X'), Player::Computer.new('O')).play
+elsif players == '2'
+  Game.new(Player::Human.new('X'), Player::Human.new('O')).play
+elsif players == '1' && first == 'Y'
+  Game.new(Player::Human.new('X'), Player::Computer.new('O')).play
+else
+  Game.new(Player::Computer.new('X'), Player::Human.new('O')).play
+end
--- a/lib/board.rb
+++ b/lib/board.rb
@@ -1,0 +1,45 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = Array.new(9, ' ')
+  end
+
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts '-----------'
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts '-----------'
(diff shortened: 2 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,31 @@
...
+
+if players == '0'
+  Game.new(Player::Computer.new('X'), Player::Computer.new('O')).play
+elsif players == '2'
+  Game.new(Player::Human.new('X'), Player::Human.new('O')).play
+elsif players == '1' && first == 'Y'
+  Game.new(Player::Human.new('X'), Player::Computer.new('O')).play
+else
+  Game.new(Player::Computer.new('X'), Player::Human.new('O')).play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts '-----------'
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    cells[position.to_i - 1]
+  end
+
+  def full?
+    true if cells.include?(' ') == false
```
randallreedjr: Prefer implicit return values of boolean statements.

`cell.include?(' ') == false`

Or, going further, avoid equality comparison with boolean values. Here you can use the bang operator `!` to negate the results.

`!cells.include?(' ')`

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def turn_count
+    cells.count { |x| x if x != ' ' }
```
randallreedjr: The `count` iterator increments whenever the condition returns true, and it does not update the original collection. This could be simplified to
`cells.count { |x| x != ' '}`

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def taken?(num)
+    cells[num.to_i - 1] != ' ' ? true : false
```

`cells[num.to_i - 1] != ' '`

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def over?
+    if board.full?
+      true
+    elsif won?
+      true
+    else
+      false
+    end
```
randallreedjr: Use or operator `||` to combine logical conditions
```
def over?
  board.full? || won?
end
```

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+    WIN_COMBINATIONS.any? { |combo| board.cells[combo[0]] == board.cells[combo[1]] && board.cells[combo[1]] == board.cells[combo[2]] && board.cells[combo[0]] != ' ' } 
+  end
+
+  def draw?
+    true if won? == false && board.full?
+  end
+
+  def winner
+    board.cells[WIN_COMBINATIONS.find { |combo| board.cells[combo[0]] == board.cells[combo[1]] && board.cells[combo[1]] == board.cells[combo[2]] && board.cells[combo[0]] != ' ' }.first] if won?
+  end
```
randallreedjr: Consider a way to encapsulate some of the logic in common between the `winner` and `won?` methods.

lib/players/computer.rb
```diff
@@ -0,0 +1,83 @@
...
+    move.to_s 
+  end
+
+  def win(position, board)
+    win = Game::WIN_COMBINATIONS.any? do |combo|
+      combo.all? do |x|
+        (board.cells[x] == token || x == position - 1) && board.valid_move?(position)
+      end
+    end
+    win ? position : nil
```
randallreedjr: Inline `if` is generally preferable to a ternary operator that returns nil.

`position if win`

randallreedjr:
@plai217 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #29, opened 2016-01-07
implement wargames

adamgfraser:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+puts "Welcome to Tic Tac Toe!"
+Game.start
```
randallreedjr: Exactly how the bin/tictactoe should look. However, we would prefer to encapsulate CLI logic into a separate class. For example, you could have a file: `lib/tic_tac_toe_cli.rb` with:

```ruby
class TicTacToeCLI
  def start
   # Your game logic
  end
end
```

Separating the execution interface from the implementation of the game allows you to have multiple interfaces - e.g. a web app.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
randallreedjr: This code already exists in `reset!`. Here you could just call the method.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    cells.all?{|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    cells.count{|cell| cell == "X" || cell == "O"}
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,129 @@
...
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    else
+      won?
+    end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def opponent
+    tokens = ["X", "O"]
+    tokens.delete(self.token)
+    Player.new(tokens.sample)
+  end
+
+  def game(board)
+    game = Game.new
+    game.board.cells = board.cells
```
randallreedjr: Here you could try passing `board` into `Game.new`, since it is part of the `initialize` method.

randallreedjr:
@adamgfraser Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #30, opened 2016-01-08
Corey

c1505:


bin/tictactoe
```diff
@@ -1,3 +1,77 @@
...
+
+
+
+if game_type == "1"
+	human_vs_computer
+elsif game_type == "0"
+	computer_vs_computer
+else
+	human_vs_human
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+require 'pry'
+
+class Board
+	attr_accessor :cells
+	def initialize
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+	end
+
+	def reset!
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,82 @@
+# require_relative "./player.rb"
+# require_relative "./players/human.rb"
+# require_relative "./players/computer.rb"
+# require_relative "./game.rb"
+require 'pry'
+
+class Board
+	attr_accessor :cells
+	def initialize
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
randallreedjr: This code is already in `reset!`; you could just call the method here.

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+
+	def position(user_input_string)
+		cells[user_input_string.chomp.to_i - 1] # user is entering 1-9.  posiitons are 0-8
+	end
+
+	def update(position_arg, player_object)
+		if valid_move?(position_arg)  #added to try and fix human#move spec
+			cells[position_arg.chomp.to_i - 1] = player_object.token
+		else
+			nil
```
randallreedjr: With Ruby's implicit return values, an `if` statement whose condition fails returns `nil`. So this method could be simplifed to:
```
def update(position_arg, player_object)
  if valid_move?(position_arg)
    cells[position_arg.chomp.to_i - 1] = player_object.token
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+		end
+	end
+
+	def full?
+		if cells.any? {|cell| cell.strip.empty?}
+			false
+		else
+			true
+		end
```
randallreedjr: Prefer implicit return values of boolean statements. You can reverse (or negate) the boolean with the `!` (bang) operator.

`!cells.any? { |cell| cell.strip.empty?}`

However, Ruby generally provides a more apt iterator. `Board#full?` is asking if all elements are not empty strings. You can use logic that is symmetrical to those semantics with the [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) iterator.

`cells.all? {|cell| !cell.strip.empty? }`

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+	def turn_count #ahh crap. i think this doesn't actually change anything.  that is why i thought i had an error.
+		#well back to the drawing board
+		count = 0
+		cells.each do |cell|
+			unless cell.strip.empty?
+				count += 1
+			end
+		end
+		count
+	end
```

```ruby
def turn_count
  cells.count { |cell| !cell.strip.empty? }
end
```

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+
+	def valid_move?(input)
+		if taken?(input)
+			false
+		elsif (1..9).include?(input.to_i)
+			true
+		else
+			false
+		end
```
randallreedjr: Here you could combine your checks into a single boolean expression:
`!taken?(input) && (1..9).include?(input.to_i)`
And remember implicit return values

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+      [6,4,2] ]
+
+    def initialize(player_1= Human.new("X"), player_2= Human.new("O"), board= Board.new) #refactor to better take in optional arguments
+    	self.player_1 = player_1
+    	self.player_2 = player_2
+    	self.board = board
+    end
+
+    def current_player
+    	x_count = board.cells.select {|place| place == "X"}.count
```
randallreedjr: `count` could replace `select` as the iterator here

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+    		combo << self.board.cells[combination[1]]
+    		combo << self.board.cells[combination[2]]
+
+    		if combo.all? {|letter| letter == 'X'}
+    			won = true
+    			break
+    		elsif combo.all? {|letter| letter =='O'}
+    			won = true
+    			break    			
+    		end
```
randallreedjr: You should rarely need to `break` from within a loop. If you need a return value from within a loop, you need to use a higher level iterator or some sort of local variable. 

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+    	end
+    	won
+    end
+
+
+    def draw?
+    	if won? == false && board.full? == true
```
randallreedjr: Equality comparisons with boolean literals are not necessary; `x == true` and `x` are equivalent. So this method could be simplifed to
```
def draw?
  if !won? && board.full?
    true
  else
    false
  end
end
```
Then, because of implicit return values, we could further simplify to
```
def draw?
  !won? && board.full?
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,85 @@
...
+			"1"
+		elsif board.cells[2] == " "
+			"3"
+		elsif board.cells[6] == " "
+			"7"
+		elsif board.cells[8] == " "
+			"9"
+		else	
+			rand(1..9).to_s
+		end
```
randallreedjr: You could avoid the duplicate method calls with something like
```
winning_move(board) || blocking_move(board) || other_move(board)
```
and moving the rest of the `elsif` logic into the new `other_move` method. If `winning_move` or `blocking_move` return nil, the next method in the chain will be evaluated; but if they return a truthy value, `move` will exit and return that value.

randallreedjr:
@c1505 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #31, opened 2016-01-12
create CLI Tic-Tac-Toe with unbeatable AI option

eric-an:


bin/tictactoe
```diff
@@ -1,3 +1,9 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic-Tac-Toe!" 
+
+game = Game.new
+
+game.start
```
randallreedjr: This executable is laudably terse, but it would be preferable to separate the interface into it's own class (e.g. GameCLI). This separates the execution logic from the implementation logic, so that you could add additional interfaces.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    reset!
+  end
+
+  def reset!
+    @cells.clear
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
randallreedjr: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the value for the elements. It's more concise and will prevent errors from typing `" "`.
Also, calling `clear` before performing the array assignment should not be necessary.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
```
randallreedjr: Just calling `reset!` should be sufficient here.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+    spot_filled = []
+    cells.each do |cell| 
+      if cell.include?("X") || cell.include?("O")
+        spot_filled << true
+      else 
+        spot_filled << false
+      end
+    end
+    ( spot_filled.include?(false) ? false : true ) 
+  end
```

```
def full?
  cells.all? { |cell| cell == "X" || cell == "O" }`
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def turn_count
+    turn_total = 0
+    cells.each do |cell| 
+      turn_total += 1 if cell.include?("X") || cell.include?("O")
+    end
+    turn_total
+  end
```
randallreedjr: For turn_count, the [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) iterator on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
  cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def taken?(number)
+    spot = number.to_i
+    ( cells[spot-1] == "X" || cells[spot-1] == "O" ? true : false )
```

`cells[spot-1] == "X" || cells[spot-1] == "O"`

lib/game.rb
```diff
@@ -0,0 +1,156 @@
...
+  ]
+
+ def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    ( board.turn_count % 2 == 0 ? @player_1 : @player_2 )
```

lib/game.rb
```diff
@@ -0,0 +1,156 @@
...
+      win_index_3 = win_combination[2]
+
+      position_1 = board.cells[win_index_1]
+      position_2 = board.cells[win_index_2]
+      position_3 = board.cells[win_index_3]
+      
+      return win_combination if (position_1 == "X" && position_2 == "X" && position_3 == "X") || (position_1 == "O" && position_2 == "O" && position_3 == "O")
+    end
+    return false
+  end
```

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

lib/game.rb
```diff
@@ -0,0 +1,156 @@
...
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,43 @@
+class Computer < Player
+  attr_accessor :board
+
+  def move(board)
+    ( !board.taken?("5") ? "5" : possible_move(board) + 1)
+  end
+
+  def possible_move(board)
+    win_or_block(board) || corner_move(board) || edge_move(board)
```
randallreedjr: Pretty good heuristic, very clean logic

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Human < Player
+
+  def move(board)
+    puts "Please choose a move by choosing a number."
+    input = gets.strip
+    input
+  end
```
randallreedjr: Because of Ruby's implicit return, `input` is unnecessary in this method. It could be simplified to
```
def move(board)
  puts "Please choose a move by choosing a number."
  input = gets.string
end
```
But since `input` isn't being used elsewhere in this method, it could be further simplified to 
```
def move(board)
  puts "Please choose a move by choosing a number."
  gets.strip
end
```

randallreedjr:
@eric-an Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #32, opened 2016-01-12
Project submitted

authorbeard:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TTTController.new
```
randallreedjr: Perfect ruby executable

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+
+class Board
+  attr_accessor :cells, :turn_count
+
+  def initialize
+    @cells = Array.new(9, " ")
```
randallreedjr: Great way to initialize the array, but isn't this code already in `reset!`? You could just call the method here.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def position(move)
+    @cells[move.to_i - 1]
+  end
+
+  def full?
+    cells.none?{|c| c == " "}
```
randallreedjr: Excellent choice of iterator, though [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) might be slightly more semantic.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def ai_update(move, player)
+    @cells[move] = player.token
+    @cells
+  end
+
+  def turn_count
+    cells.select{|c| c != " "}.size 
+  end
```
randallreedjr: Use the highest level and most expressive iterator possible to solve a problem. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect, so the entire body could read:

```ruby
def turn_count
  cells.count {|c| c != " "}
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+    @turn_count = 0
+
+  end
+  
+
+  def current_player
+    (board.turn_count + 1) % 2 == 0 ? @player_2 : @player_1
```

`(board.turn_count + 1).even? ? @player_2 : @player_1`

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def over?
+    won? || draw?
+  end
+
+  def won?
+    @winning_array = WIN_COMBINATIONS.detect{|a| 
+      a.all?{|p| board.cells[p] == "X" } || a.all?{|p| board.cells[p] == "O" }
+      } 
+    @winning_array ? true : false
```
randallreedjr: Good job storing the winning move. This makes `winner` more efficient. However, you could rely on ruby's implicit return value instead of using a ternary operator (if `detect` fails, it will return nil). If you want to ensure a `true` or `false` value, you could also use the double bang operator, e.g.  `!!@winning_array`.

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
+    won? ? board.cells[@winning_array[0]] : nil
```
randallreedjr: Here you can use an inline if statement to avoid a ternary operator with an explicit nil.
```
board.cells[@winning_array[0]] if won?
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def turn
+    puts "\e[H\e[2J"
+    board.display
+    if current_player.is_a?(Player::Human) 
+      @move = current_player.move(current_player.token) 
+      if board.valid_move?(@move) 
+        board.update(@move, current_player)
+      else
+        puts "NOPE. Try again, dumbass."
```
randallreedjr: Nice lol

randallreedjr:
@authorbeard Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

authorbeard:
Randall, 

Thank you for taking the time to go through this. I think I misunderstood some of the instructions; thought I'd get an assessment only after all the projects were done (then I got to the readme of the last one, the gem, and it sounded like that was the only one that anybody would be looking at). So I didn't refactor or anything. My AI was just a brutal weekend of brute-forcing and once I saw everything was working, I moved on, even knowing this was ugly and clunky as all hell. 

I've had all sorts of problems with inline if statements, hence the clunky ternaries ("at least it's only one line, and I'll go back to finish it later," I thought).  And I had a couple of "Ah, dumbass! of course!" moments reading through this. Like, "Of course there's a count method ... why didn't you even look?"

So thank you again. It's already helped.

Michael

PR #33, opened 2016-01-12
Project submitted

cbruen:


bin/tictactoe
```diff
@@ -1,3 +1,51 @@
...
+elsif num_players == "0"
+	computer1 = Computer.new("X")
+	computer2 = Computer.new("O")
+	game = Game.new(computer1, computer2)
+	game.play
+else
+	puts "Invalid number of players"
+end
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+	attr_accessor :cells
+	
+	def initialize
+		@cells = Array.new(9," ")
+	end
+	
+	def reset!
+		@cells = @cells.clear
```
randallreedjr: You don't need to reassign the array after [`#clear`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-clear); `@cells.clear` would have the same effect as `@cells = @cells.clear`.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+	attr_accessor :cells
+	
+	def initialize
+		@cells = Array.new(9," ")
```
randallreedjr: Great way to initialize the array, but isn't this code already in `#reset!`? You could just call the method here.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+	
+	def full?
+		!@cells.include?(" ")
+	end
+	
+	def turn_count
+		output = 0
+		@cells.each { |x| output+=1 if x!= " "}
+		output
+	end
```
randallreedjr: Use the highest level and most expressive iterator possible to solve a problem. [`#count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
  @cells.count { |x| x != " " }
end
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
+require_relative 'board.rb'
+require_relative 'players/human.rb'
+require_relative 'players/computer.rb'
+#require_relative 'player.rb'
```
randallreedjr: Require statements should not be necessary in class files; the environment.rb should take care of these.

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+		@player_2 = player_2
+		@board = board
+	end	
+	
+	def current_player
+		self.board.cells.count("O") == self.board.cells.count("X") ? self.player_1 : self.player_2
+	end
+	
+	def won?
+		!WIN_COMBINATIONS.detect { |x| self.board.cells[x[0]] == self.board.cells[x[1]] && self.board.cells[x[0]] == self.board.cells[x[2]] && self.board.cells[x[0]] != " "}.nil?
```
randallreedjr: `#detect` is a great iterator to use when you care about which element satisfies the provided criteria. When you only care about whether *any* element meets the criteria, [`#any?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-any-3F) is a good alternative as it returns `true` or `false`.

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+	end
+	
+	def winner
+		if self.won? && !WIN_COMBINATIONS.detect { |x| self.board.cells[x[0]] == self.board.cells[x[1]] && self.board.cells[x[0]] == self.board.cells[x[2]] && self.board.cells[x[0]] == "X"}.nil?
+			return "X"
+		elsif self.won?
+			return "O"
+		else 
+			return nil
+		end
```
randallreedjr: As written, this method iterates through `WIN_COMBINATIONS` three times - for `!WIN_COMBINATIONS.detect` in this method, and the two calls to `self.won?`. I bet you could find a way to simplify this.

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../player.rb'
+require 'pry'
+
+class Human < Player
+	attr_accessor :name
+	def move(arr)
+		puts "What move would you like to make?"
+		move = gets.chomp
+		move
+	end
```
randallreedjr: Because of Ruby's implicit return, the varible `move` is unnecessary in this method. It could be simplified to
```
def move(arr)
  puts "What move would you like to make?"
  move = gets.chomp
end
```
But since `move` isn't being used elsewhere in this method, it could be further simplified to 
```
def move(arr)
  puts "What move would you like to make?"
  gets.chomp
end
```

randallreedjr:
@cbruen Great job! I left a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #35, opened 2016-01-13
tictac toe with AI

kylegraydev:


lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```
aviflombaum: I'd just call `self.reset!` here instead of reimplementing the logic below.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    cells[position.to_i-1]
+  end
+
+  def full?
+    full = cells.include?(" ")
+    !full
```
aviflombaum: Nice use of `include?`. However, you could just negate in place with the `!` operator.

`!cells.include?(" ")` and not need the locale variable.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def turn_count
+    counter = 0
+    cells.each do |p|
+      if p != " "
+        counter += 1
+      end
+    end
+    counter
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def cells=(set_cells)
+    @board.cells = set_cells
+  end
+
+  def cells
```
aviflombaum: I think this method should just be called `board` and still read out of `@cells` and you wouldn't have needed to change your test suite in 100 places.

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    (@board.cells == []) ? @board.cells : @board
+  end
+
+  def won?
+    $winner = nil
```
aviflombaum: No global variables!

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+        break
+      elsif @board.cells[win_combo[0]] == 'O' && @board.cells[win_combo[1]] == 'O' && @board.cells[win_combo[2]] == 'O'
+        $winner = player_2.token
+        returnable = win_combo
+        break
+      else
+        returnable = false
+      end
+    end
+    returnable
```
aviflombaum: You should never `return` from within a loop. If you need a return value from within a loop you need to use a higher level iterator or some sort of local variable. 

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    true if @board.full? == true && won? == false
+  end
+  #
+  def over?
+    true if draw? || won?
+  end
+  #
+  def winner
+    over?
+    $winner
```
aviflombaum: No global variables!!!

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+    returnable
+  end
+  #
+  def draw?
+    true if @board.full? == true && won? == false
+  end
+  #
+  def over?
+    true if draw? || won?
```
aviflombaum: `draw? || won?` would suffice

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+        break
+      else
+        returnable = false
+      end
+    end
+    returnable
+  end
+  #
+  def draw?
+    true if @board.full? == true && won? == false
```
aviflombaum: `@board.full?  && !won?` would suffice

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+  def winner
+    over?
+    $winner
+  end
+  #
+  #
+  #
+  def current_player
+    player = @board.turn_count
+    player.even? ? player_1 : player_2
```
aviflombaum: Nice use of `even?` instead of `%`

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  end
+
+  def turn
+    spot = current_player.move(@board)
+    !@board.valid_move?(spot) ? turn : @board.update(spot, current_player)
```
aviflombaum: Nice

lib/players/computer.rb
```diff
@@ -0,0 +1,65 @@
...
+      temp_array = []
+      temp_array.push(board_state.cells[combo[0]])
+      temp_array.push(board_state.cells[combo[1]])
+      temp_array.push(board_state.cells[combo[2]])
+      if temp_array[0] == temp_array[1] && temp_array[2] == ' '
+        @best_move = combo[2]
+        break
+      elsif temp_array[1] == temp_array[2] && temp_array[0] == ' '
+        @best_move = combo[0]
+        break
```
aviflombaum: Try not to `break` out of loops.

aviflombaum:
@kylegraydev nice job! left you some comments, going to close, feel free to reopen with questions.

PR #37, opened 2016-01-13
Done

ryuichi7:


bin/tictactoe
```diff
@@ -1,3 +1,79 @@
...
+    puts "Goodbye"
+  else
+    invalid_input
+    game_end
+  end
+end
+
+
+puts "Welcome to tic tac toe!"
+game_play
```
randallreedjr: Prefer to encapsulate CLI logic into a class. For example the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

config/environment.rb
```diff
@@ -1,4 +1,5 @@
+require 'pry'
 require 'bundler'
 Bundler.require
 
-require_all 'lib'
+require_all 'lib'
```
randallreedjr: It is a [Ruby style convention](https://github.com/styleguide/ruby) to end every file with a blank newline.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```
randallreedjr: Great way to initialize the array, but isn't this code already in `reset!`? You could just call the method here.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    cells.all? { |token| token == "X" || token == "O" }
```
randallreedjr: Great choice of iterator for this method.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def taken?(input)
+    position(input) == "X" || position(input) == "O"
+  end
+
+  def turn_count
+    cells.grep(/[XO]/).size
```
randallreedjr: Nice use of a regular expression. FYI you can also use [`#count`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-count).

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+  def draw?
+    board.full? && !won?
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
+    if winning_combo = won?
```
randallreedjr: Excellent use of if statement assignment

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+    
+  
+  def move(input = nil)
+    input = gets.strip
+    input
```
randallreedjr: Because of Ruby's implicit return, `input` is unnecessary in this method. It could be simplified to
```
def move(input = nil)
  input = gets.string
end
```
But since `input` isn't being used elsewhere in this method, it could be further simplified to 
```
def move(input = nil)
  gets.strip
end
```

randallreedjr:
@ryuichi7 Great submission! I left a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any.

PR #38, opened 2016-01-13
Project submitted

azheng249:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
...
+puts "Do you want to play again? Y/N "
+answer = gets.strip.upcase
+
+until answer == "NO" || answer == "N"
+	new_game
+	puts "Do you want to play again? Y/N "
+	answer = gets.strip.upcase
+end
+
+puts "I hope you enjoyed playing Tic Tac Toe."
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+class Board
+	attr_accessor :cells
```
aviflombaum: Indent this class correctly. Two spaces, no tabs.

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+	def position(input)
+		@cells[input.to_i - 1]
+	end
+
+	def update(input, player)
+		@cells[input.to_i - 1] = player.token
+	end
+
+	def full?
+		turn_count == 9 ? true : false
```
aviflombaum: `turn_count == 9` is enough. clever too.

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+	end
+
+	def turn_count
+		@cells.count{|cell| cell == "X" || cell == "O" }
```
aviflombaum: Nice use of count.

lib/game.rb
```diff
@@ -0,0 +1,73 @@
+require_relative 'board.rb'
+require_relative 'player.rb'
+require_relative 'players/human.rb'
+require_relative 'players/computer.rb'
+
```
aviflombaum: All requires should be in environment.

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+	]
+
+	def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+		@board = board
+		@player_1 = player_1
+		@player_2 = player_2
+	end
+
+	def current_player
+		@board.turn_count%2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+	end
+
+	def draw?
+		!won? && board.full?
+	end
+
+	def winner
+		if won?
+			return board.cells[won?[0]]
+		end
```
aviflombaum: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,112 @@
...
+
+	# Evaluates if the current player wins, loses or draws and sends back a score based on the result.
+	def score(board, depth)
+		if winner(board) == @token
+			return 10 - depth
+		elsif winner(board) == @opponent
+			return depth - 10
+		else
+			return 0
+		end
```
aviflombaum: implicit returns would work here so no need for `return`.

Also I think I'd use a case statement:

```ruby
case winner(board) ==
when @token
  10 - depth
when @opponent
  depth - 10
else
  0
end
```

aviflombaum:
Nice job @azheng249. I left some comments, closing, feel free to re-open with questions.

PR #39, opened 2016-01-14
Project submitted

MitulMistry:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
...
+      game_loop
+    else
+      puts "Goodbye."
+      break
+    end
+  end
+
+end
+
+game_loop
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    9.times { @cells << " " }
+  end
+
+  def reset!
+    @cells.clear
+    9.times { @cells << " " }
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.all? { |cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    @cells.select { |cell| cell == "X" || cell == "O"}.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    ]
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
```
aviflombaum: Nice use of `even?` instead of `%`

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+      @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.taken?(combo[0] + 1) #@board.taken?(combo[0] + 1) to make sure it == "X" or "O" and not empty spaces, +1 for offsetting to user input
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    !won? || draw? ? nil : @board.cells[won?.first]
```
aviflombaum: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

aviflombaum:
@MitulMistry nice job! left you some comments and closed. Feel free to re-open or ask questions.

PR #40, opened 2016-01-14
Project submitted

jamcelr:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+again = "Y"
+while again == "Y"  
+  Game.new.start
+  puts "Would you like to play again? (Y/N)"
+  again = gets.strip.upcase
```
randallreedjr: Prefer to encapsulate CLI logic into a class. For example the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")                                                                                            
+  end
+
+  def cells
+    @cells
```
randallreedjr: With `attr_accessor :cells`, this method is not necessary. The `attr_accessor` adds two (hidden) methods:

```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+    puts " "
+  end
+
+  def position(pos)
+    @cells[pos.to_i-1]
+  end
+
+  def taken?(pos)
+    position(pos) == "X" || position(pos) == "O" ? true : false
```

`position(pos) == "X" || position(pos) == "O"`

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def full?
+    @cells.all? {|e| e == "X" || e == "O"}
+  end
+
+  def turn_count
+    @cells.count {|e| e == "X" || e == "O"}
+  end
```
randallreedjr: Your choice of iterators in both `full?` and `turn_count` is perfect.

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def valid_move?(pos)
+    !taken?(pos.to_i) && pos.to_i.between?(1,9) ? true : false
```
randallreedjr: Again, this would suffice:
`!taken?(pos.to_i) && pos.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize (player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count.even? ? player_1 : player_2
```
randallreedjr: Perfect use of `even?` method and ternary operator.

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  def over?
+    won? || draw? || board.full? ? true : false
```
randallreedjr: The last term, `board.full? ? true : false`, is already covered by the `draw?` method. This could be reduced to:
```
def over?
  won? || draw?
end
```

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+      win_index_3 = win_combo[2]
+
+      position_1 = board.cells[win_index_1]
+      position_2 = board.cells[win_index_2]
+      position_3 = board.cells[win_index_3]
+      if position_1 == "X" && position_2 == "X" && position_3 == "X" || position_1 == "O" && position_2 == "O" && position_3 == "O"
+        return win_combo
+      end
+    end
+    return false
```

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,40 @@
+class Player::Computer < Player
+  attr_accessor
+
+  def move(board)
+    return "5" if !board.taken?("5")
+    win_move(board)
+  end
+
+  def win_move(board)
+    win(board) || block_move(board) || best_move(board)
```
randallreedjr: Love the way that reads as a precedence of choices.

lib/players/computer.rb
```diff
@@ -0,0 +1,40 @@
...
+      (board.cells[i[0]] == token && board.cells[i[1]] == token && !board.taken?(i[2]+1)) ||
+      (board.cells[i[1]] == token && board.cells[i[2]] == token && !board.taken?(i[0]+1)) ||
+      (board.cells[i[2]] == token && board.cells[i[0]] == token && !board.taken?(i[1]+1))
+    end
+  end
+ 
+  def win(board)
+    if !stop_win(board, self.token).nil?
+      stop_win(board, self.token).find {|pos| board.cells[pos] == " "}.to_i+1
+    end
```
randallreedjr: We can also apply if statement assignment here.

randallreedjr:
@jamcelr Great work! I left some comments, feel free to respond if you want to discuss any of them.

@aviflombaum Ready to close.

PR #41, opened 2016-01-16
Project submitted

zacscodingclub:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToe.new.call
```
randallreedjr: Exactly how the bin/tictactoe should look. Great encapsulation of CLI in separate class.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    "-----------"
+  end
+
+  def position(index)
+    @cells[index.to_i - 1]
+  end
+
+  def full?
+    @cells.any? { |cell| cell == " "} ? false : true
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. Reverse (or negate) the boolean with the `!` (bang) operator.

`!@cells.any? { |cell| cell == " "}`

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def taken?(index)
+    @cells[index.to_i - 1 ] == " " ? false : true
+  end
```
randallreedjr: This could similarly be written as
```ruby
@cells[index.to_i - 1 ] != " "
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    index.to_i.between?(1,9) && !taken?(index)
+  end
+
+  def empty_cell_count
+    10 - turn_count
+  end
+
+  def turn_count
+    @cells.select { |cell| cell != " " }.length
+  end
```
randallreedjr: Use the highest level and most expressive iterator possible to solve a problem. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect:

```ruby
def turn_count
  @cells.count { |cell| cell != " " }
end
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+  ]
+
+  def initialize(*args)
+    @player_1 = args[0] || Player::Human.new("X")
+    @player_2 = args[1] || Player::Human.new("O")
+    @board = args[2] || Board.new
+  end
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
```
randallreedjr: Perfect use of `even?` method and ternary operator.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def draw?
+    ( @board.full? && !won? ) ? true : false
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. 

```ruby
def draw?
  @board.full? && !won?
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    cell_start
+  end
+
+  def reset!
+    cell_start
+  end
+
+  def cell_start
+    @cells = Array.new(9, " ")
+  end
```
randallreedjr: The delegation to the `cell_start` method here is unnecessary. You can just put this login in the `reset!` method, and call that method from `initialize`.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    [2,5,8],
+    [0,4,8],
+    [6,4,2]
+  ]
+
+  def initialize(*args)
+    @player_1 = args[0] || Player::Human.new("X")
+    @player_2 = args[1] || Player::Human.new("O")
+    @board = args[2] || Board.new
+  end
```
randallreedjr: Nice use of splat arguments and default values for the instance variables.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def over?
+    won?  || draw?
+  end
+
+  def won?
+    WIN_COMBINATIONS.any? do |combination|
+      return combination if winner_x(combination) || winner_y(combination)
+    end
```
randallreedjr: Instead of using `any?` iterator with a return statement, use [`detect`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect). The element that satisfies the condition will automatically be returned.

```ruby
def won?
  WIN_COMBINATIONS.detect do |combination|
    winner_x(combination) || winner_y(combination)
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def winner
+    sequence = won?
+    sequence.is_a?(Array) ? @board.cells[sequence[0]] : nil
+  end
```
randallreedjr: Consider using end-of-line if statement instead of ternary. `nil` will be returned if statement is not true.

`@board.cells[sequence[0]] if sequence.is_a?(Array)`

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def winner_x(combination)
+    combination.all? { |x| @board.cells[x] == "X" }
+  end
+
+  def winner_y(combination)
+    combination.all? { |o| @board.cells[o] == "O" }
+  end
```
randallreedjr: `winner_x` and `winner_y` are great methods for delegating this logic. But shouldn't `winner_y` be `winner_o`?

lib/players/computer.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def win(board)
+    if !never_win(board, self.token).nil?
+      never_win(board, self.token).find { |location| board.cells[location] ==  " " }.to_i+1
+    end
+  end
+
+  def winning_move(board)
+    block_move(board) || win(board) || best_move(board)
+  end
```
randallreedjr: Love the way that reads as a precedence of choices. However, shouldn't it be:

Try to win
Try to block
Pick a best move

?

lib/tictactoe.rb
```diff
@@ -0,0 +1,88 @@
...
+        again?
+      when 1
+        one_game(goes_first_token)
+        again?
+      when 2
+        two_game(goes_first_token)
+        again?
+      else
+        puts "I couldn't understand that, please try again."
+    end
```
randallreedjr: Case statement is an excellent choice here.

lib/game.rb
aviflombaum: While splat is possible, I think given this game the explicit arguments might even be better. Less chance or error.

randallreedjr:
@zacscodingclub Great submission! I left some comments, feel free to respond if you want to discuss any of them. Closing for now.

lib/board.rb
zacscodingclub: I think it's strange that I used so many ternary operators explicitly returning a boolean value. I am trying to track down where I started this pattern and I believe it's some code I took from a previous TTT project and copy + pasted over here. Once I got one test passing, I just used the explicit version again and again throughout my code because it worked...  This tells me that I need to give some focus on the refactor stage of TDD and not be so excited about tests being green.

I prefer your use of the implicit boolean and agree that the #all? best expresses what I'm trying to complete with this #full? method.

lib/game.rb
zacscodingclub: Good catch here and I remember when I originally wrote this method, the code read: `combination.all? { |y| @board.cells[y] == "Y" }` and it took me awhile to figure out why that didn't work.  I remember fixing the internals of the method upon realizing `"Y" != "O"`, but left the method name the same for no logical reason.

lib/game.rb
zacscodingclub: @aviflombaum Looking for clarification here.  What's the use case for splat and why wouldn't it be as good as explicit arguments?

I included it in this project for two reasons: 
1. I just read about it in POODR and wanted to try it out on my own code because it's a good way to learn things.
2. It extracts the default value assignment outside of the method call, which I thought looked ugly.

lib/players/computer.rb
zacscodingclub: Yes, yes it should.

zacscodingclub:
@randallreedjr First of all, I just wanted to say I appreciate your review of my code!  I do believe this is the first time anyone else has looked over a complete project and given me specific feedback.  I was reading the code + your notes, and I was thinking "Did I write this code? This doesn't look like my code..."  I think it's just a weird thing of being in a little bit of flow while writing certain parts of this project.  I can be chugging along passing tests without a single regard for the quality of the code that I'm writing.  

I think I understood all the points you made and I replied to a couple above.  So I went ahead and refactored the project to reflect the patterns and code you pointed out above and have submitted them under this commit:

https://github.com/zacscodingclub/ttt-with-ai-project-v-000/commit/97cffa33b69c81af213fc2256e5ab63b6650ea13

Thanks for your time!

PR #42, opened 2016-01-16
Project submitted

Rawradin:


bin/tictactoe
```diff
@@ -1,3 +1,78 @@
...
+    elsif temp == "O"
+      wins_2 += 2
+    end 
+  end
+  puts "         WARGAME OVER!"
+  puts "         _____________"
+  puts "Computer 1           Computer 2"
+  puts "WINS:#{wins_1}       WINS:#{wins_2}"
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

The code in your bin would generally just go in call. You'd also make these other methods like `#wargame` just instance methods in your class and your call method would just be the procedural code.

lib/board.rb
```diff
@@ -0,0 +1,64 @@
+class Board
+  attr_accessor :cells, :token
+  def initialize(amount=9,cell=" ")
+    @cells = []
+    @token = "X"
+    amount.times do |i|
+      @cells << cell
+    end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def reset!
+    @cells.clear
+    9.times do |i|
+      @cells << " "
+    end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    return @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.include?(" ") ? (return false) : (return true)
```
aviflombaum: Use the most expressive and efficient iterator for a problem. `Board#full?` is asking if all elements are not empty strings. You can use logic that is symmetrical to those semantics with the [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) iterator.

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def turn_count
+    count = 0
+    @cells.each do |cell|
+      count += 1 unless cell == " "
+    end
+    # @token = "O" if count % 2 == 0
+    # bindind.pry
+    return count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def taken?(input)
+    if (position(input.to_i) == "X" || position(input.to_i) == "O")
+      return true
+    else 
+      return false
+    end
```
aviflombaum: Prefer implicit return values of boolean statements. This would suffice: `position(input.to_i) == "X" || position(input.to_i) == "O"`

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def valid_move?(input)
+    if (input.to_i > 0 && input.to_i <= 9) && !taken?(input.to_i)
+      return true
+    else 
+      return false
+    end
```
aviflombaum: Prefer implicit return values of boolean statements. This would suffice: `input.to_i > 0 && input.to_i <= 9) && !taken?(input.to_i)`

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def current_player
+    turns = @board.turn_count
+    # binding.pry
+    if turns % 2 == 0
+      @current_player = @player_1
+    else 
+      @current_player = @player_2
+    end
+    return @current_player
```

Notice the implicit return values to.

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def over?
+    if draw? || @board.full? || won? 
+      return true
+    else 
+      return false
+    end
```
aviflombaum: `draw? || @board.full? || won? ` would suffice.

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+      position_1 = @board.cells[combo[0]]
+      position_2 = @board.cells[combo[1]]
+      position_3 = @board.cells[combo[2]]
+
+      if(position_1 == "X" && position_2 == "X" && position_3 == "X" || position_1 == "O" && position_2 == "O" && position_3 == "O")
+        puts "Congratulations #{position_1}!"
+        return combo
+      end
+    end
+    return false
```
aviflombaum: You should never `return` from within a loop. If you need a return value from within a loop you need to use a higher level iterator or some sort of local variable. 

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def draw?
+    if !@board.full? || won?
+      return false
+    else
+      puts "Cats Game!"
+      return true
+    end
```
aviflombaum: `!(@board.full? || won?)` would suffice

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def winner
+    if self.won?
+      return @board.cells[won?[0]]
+    else
+      return nil
+    end
```
aviflombaum: `@board.cells[won?[0]] if self.won?` would suffice. But better yet, to avoid calling `won?` twice, I'd write.

```ruby
if won = self.won?
  @board.cells[won[0]]
end
```

This would assign the return value of `won?` to a local variable `won`. If that local variable is assigned and true, the statement executes and we can use that variable so we don't call `won?` twice.

lib/players/computer.rb
```diff
@@ -0,0 +1,93 @@
...
+          end
+          i += 1
+        end
+      end
+    end
+    # puts "IM CHOOSING #{input}"
+    # puts "CHOICES FROM #{odds.to_s}"
+    # puts "BOARD IS #{board.to_s}"
+    return input
+  end
```
aviflombaum: This method is really long. I'd think of some ways to break this up into more discrete logic. Things like:

`block_move` or `try_to_win` or `take_middle`

aviflombaum:
@Rawradin nice job! Let me know if you have questions, feel free to re-open.

PR #43, opened 2016-01-16
Project submitted

zivolution921:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
...
+play_again = ""
+until play_again.downcase = "n" || play_again.downcase == "no"
+  puts "Would you like to play again?"
+  play_again = gets.strip
+  play_again.downcase == "y" || play_again.downcase == "yes"
+  game = Game.new
+  until game.won?
+    game.play
+  end
+end
```
aviflombaum: Prefer to encapsulate CLI logic into a class.

For example the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
  		 
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
   # ex:
  puts "Welcome to Tic Tac Toe!"
    game = Game.new
    until game.won?
      game.play
    end

    play_again = ""
    until play_again.downcase = "n" || play_again.downcase == "no"
      puts "Would you like to play again?"
      play_again = gets.strip
      play_again.downcase == "y" ||lay_again.downcase == "yes"
      game = Game.new
      until game.won?
        game.play
      end
    end
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def update(pos, player)
+    cells[pos.to_i - 1] = player.token
+  end
+
+  def full?
+   cells.detect {|cell| cell == " " || cell == nil} ? false : true
```
aviflombaum: 1. Prefer implicit return values of boolean statements.

2. A ternary operator that returns `true` or `false` is never needed.

3. Use `!` negation to reverse the "truthyness" of the boolean. `!cells.detect {|cell| cell == " " || cell == nil}`

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def turn_count
+    if cells.select {|cell| cell == "O" || cell == "X"}
+      return cells.select {|cell| cell == "O" || cell == "X"}.count
+    else
+      return 0
+    end
```
aviflombaum: I think the use the block version of [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def taken?(pos)
+    position(pos) != " " && position(pos) != nil ? true : false
```
aviflombaum: Use the implicit boolean return value of the expression instead of adding the ternary.

This would suffice.

```ruby
position(pos) != " " && position(pos) != nil
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def valid_move?(pos)
+    if (1..9).include?(pos.to_i) && taken?(pos) == false
+      return true
+    else 
+      false
+    end
```
aviflombaum: Use the implicit boolean return value of the expression instead of an if statement.

This would suffice.

`(1..9).include?(pos.to_i) && !taken?(pos)`

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def winner
+    if winning_combo = won?
```
aviflombaum: Loved the way you assigned the value of `winning_combo` to a local variable within the if statement condition so that if the variable was assigned correctly you could use that value within the body of the if statement.

lib/players/computer.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../player.rb'
```
aviflombaum: Classes should not be requiring files, that should be handled by the environment.

lib/players/computer.rb
```diff
@@ -0,0 +1,42 @@
...
+class Player::Computer < Player
+  attr_accessor :player, :choice, :board
+
+  def move(board)
+    return "5" if !board.taken?("5")
+    winning_move(board)
+  end 
+
+  def winning_move(board)
+    block_move(board) || win(board) || best_move(board)
```
aviflombaum: Love the way that reads as a precedence of choices.

aviflombaum:
@zivolution921 Great job! I left you some comments. Closing for now but feel free to respond and re-open if you'd like to discuss any of them.

PR #44, opened 2016-01-19
Project submitted

Tsundu:


bin/tictactoe
```diff
@@ -1,3 +1,41 @@
...
+  #if num_players == "1"
+    Game.new(Player::Human.new('X'),Player::Human.new('O'),Board.new).play
+  ## loop game instead of 9 times
+  ## what happens when their spot is already taken?
+end 
+
+puts "Do you want to play again? ( y / n )"
+again = gets.chomp 
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  attr_accessor :cells 
+
+ def initialize 
+   @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
+   
+ end
+   
+  def reset!
+    @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
+class Board 
+  attr_accessor :cells 
+
+ def initialize 
+   @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
```
randallreedjr: You already have this code in your class; avoid repetition by calling the `reset!` method here.

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+    @cells.count{|x|x=='X'||x=='O'}
+  end 
+
+  def taken?(num)
+     if @cells[num.to_i-1] == "X" || @cells[num.to_i-1] == "O"
+      true 
+    else 
+      false 
+    end 
+  end
```
randallreedjr: Prefer implicit return values of boolean statements instead of adding the if statement to return `true` or `false`.

```
def taken?(num)
  @cells[num.to_i-1] == "X" || @cells[num.to_i-1] == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  end
+
+  def current_player
+    if board.turn_count % 2 == 0 
+      @player_1 
+    else 
+      @player_2 
+    end 
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def over?
+    true if board.full? 
+  end
```
randallreedjr: Simplifies to
```
def over?
  board.full?
end
```
But...can a game end before the board fills up?

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+
+class Player::Human < Player 
+ 
+def move(arr)
+  
+input = gets.strip 
+
+end
```
```
def move(token, board)
  gets.strip
end
```

randallreedjr:
@Tsundu Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #46, opened 2016-01-20
Done

ShanaLMoore:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeCLI.new
```
randallreedjr: Perfect Tic Tac Toe bin file!

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def turn_count
+    count = 0
+    cells.each {|cell| 
+      if (cell == "X") || (cell == "O") 
+        count +=1 
+      end
+    }
+    count
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def taken? (position)
+    @cells[position.to_i-1] == "X" || @cells[position.to_i-1] == "O"
+  end
+
+  def valid_move?(position)
+    (position.to_i > 0 && position.to_i < 10) != taken?(position)
+  end
```

`position.between?(1,9) != taken?(position)`

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  WIN_COMBINATIONS = [
+  [0,1,2], #top row
+  [3,4,5], #middle row
+  [6,7,8], #bottom row
+  [0,3,6], #left column
+  [1,4,7], #middle column
+  [2,5,8], #right column
+  [0,4,8], #diagonal right
+  [6,4,2]] #diagonal left
```
randallreedjr: Very helpful comments

randallreedjr:
@ShanaLMoore Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #48, opened 2016-01-23
ttt-with-ai

yangc5:


bin/tictactoe
```diff
@@ -1,3 +1,42 @@
...
+      puts "please enter 0, 1 or 2."
+    end
+
+    puts "Would you like to play another game? Y/N"
+    answer = gets.chomp
+    if answer != "Y"
+      continue_play = false
+    end
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def turn_count
+    count = 0
+    cells.each do |cell|
+      if cell != " "
+        count += 1
+      end
+    end
+    count
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def taken?(position)
+    if cells[position.to_i-1] == "X" || cells[position.to_i-1] == "O"
+      return true
+    else
+      return false
+    end
+  end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the if statement. 
```
def taken?(position)
  cells[position.to_i-1] == "X" || cells[position.to_i-1] == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def valid_move?(input)
+    if ("1".."9").to_a.include?(input) && !taken?(input)
+      return true
+    else
+      return false
+    end
+  end
```
randallreedjr: (same as above)

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+
+  def current_player
+    current_player = @player_1
+    if board.turn_count%2!=0
+      current_player = @player_2
+    end
+    return current_player
+  end
```

```
if board.turn_count.odd? 
  current_player = @player_2
end
```
You could further simplify with the ternary operator:
```
current_player = (board.turn_count.even? ? @player_1 : @player_2)
return current_player
```
Finally, you could use the implicit return value of the expression to remove the local variable `current_player`
```
def current_player
  board.turn_count.even? ? @player_1 : @player_2
end
```

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+          end
+      end
+      return false
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won[0]]
end
```

spec/01_board_spec.rb
```diff
@@ -46,6 +47,8 @@
       expect(output).to include("-----------")
       expect(output).to include(" X | O | O ")
 
+      # binding.pry
```
randallreedjr: Good idea to try to remove commented code before submitting pull requests.

randallreedjr:
@yangc5 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #51, opened 2016-01-24
Done

aenser:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,79 @@
...
+    @cells.all? do |char|
+      char == "X" || char == "O"
+    end
+  end
+
+  def turn_count
+    @cells.count do |num|
+      num == "X" || num == "O"
+    end
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def taken?(input)
+    input = input.to_i 
+    input = input - 1
```
randallreedjr: You can perform both of these operations in one step: `input = input.to_i - 1`

lib/board.rb
```diff
@@ -0,0 +1,79 @@
...
+    if @cells[input] == "X" || @cells[input] == "O"
+      true
+    else
+      false
+    end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the `if` statement. 

```
def taken?(input)
  input = input.to_i - 1
  @cells[input] == "X" || @cells[input] == "O"`
end
```

lib/board.rb
```diff
@@ -0,0 +1,79 @@
...
+
+
+
+
+
+
+
+
+
+
```
randallreedjr: [Look at all this space!](http://giphy.com/gifs/yrFrXTTTcHIY0)

lib/game.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def over?
+    draw? || won? ? true : false
```
randallreedjr: Just like with the `if` statement above, a ternary operator with an explicit `true`/`false` is extraneous.
```
def over?
  draw? || won?
end
```

lib/game.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def draw?
+    board.full? && !won? ? true : false
+  end
+
+  def winner 
+    WIN_COMBINATIONS.detect do |y|
+      
+        if @board.cells[y[0]] == "X" && @board.cells[y[1]] == "X" && @board.cells[y[2]] == "X"
+          return "X"
```
randallreedjr: I think the explicit `return` from inside the `detect` iterator is unnecessary. Good choice of iterator, just let it exit out of the loop normally.

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+  def move(location = "")
+    location = gets.chomp
+    location
+  end
```
```
def move(location = "")
  gets.chomp
end
```

randallreedjr:
@aenser Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #53, opened 2016-01-26
Done

davdkm:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe"
+game = Game.start
+
+until game.over?
+  game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

Furthermore, splitting the interface (`TicTacToeCLI`) from the implementation of the game (`Game`) allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    puts "------------"
+  end
+
+  def position(cell)
+    self.cells[cell.to_i - 1]
+  end
+
+  def cell_taken?(cell)
+    cell == "X" || cell == "O"
+  end
```
randallreedjr: Couldn't you implement this logic directly in `taken?`? Why abstract it out to an additional method?

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    self.cells.all?{|cell| cell_taken?(cell)}
+  end
+
+  def turn_count
+    count = 0
+    self.cells.each do |cell|
+      count += 1 if cell_taken?(cell)
+    end
+    count
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def valid_move?(cell)
+    cell_taken?(self.cells[cell.to_i - 1]) == false && cell.to_i.between?(1, 9)
```
randallreedjr: Instead of comparing equality with false (`== false`), you can reverse (or negate) the boolean with the `!` (bang) operator.

`!cell_taken?(self.cells[cell.to_i - 1])`

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+      self.board.cells[combo[0]] == self.board.cells[combo[1]] && self.board.cells[combo[1]] == self.board.cells[combo[2]] && self.board.cells[combo[0]] != " " ? combo : nil
+    end
+  end
+
+  def draw?
+    !won? && self.board.full?
+  end
+
+  def winner
+    won? ? self.board.cells[won?.first] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,57 @@
...
+      cpu_move(board).to_i + 1
+    end
+  end
+
+  def random
+    rand(1..9)
+  end
+
+  def cpu_move(board)
+    win(board) || block(board) || random
```
randallreedjr: Excellent, readable method

randallreedjr:
@davdkm Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #54, opened 2016-01-27
Done

HarlemSquirrel:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "\t X O X O X O X O X O X O X O X O X O X O X"
+puts "\t\t Welcome to Tic Tac Toe!"
+puts "\t X O X O X O X O X O X O X O X O X O X O X"
+
+game = Game.new
+game.start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

Furthermore, by splitting the interface (`TicTacToeCLI`) from the implementation (`Game`) of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+    position(input) == "X" || position(input) == "O"
+  end
+
+  def valid_move?(input)
+    # ensure that moves are between 1-9 and not taken
+    input.to_i.between?(1,9) && taken?(input) == false
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+  ]
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    return @player_1 if @board.turn_count % 2 == 0
```

`return @player_1 if @board.turn_count.even? `

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+      tokens = [@board.cells[combo[0]], @board.cells[combo[1]], @board.cells[combo[2]]]
+      return "X" if tokens.count("X") == 3
+      return "O" if tokens.count("O") == 3
+    end
+    nil
+  end
+
+  def won?
+    return true if winner != nil
+    false
```
randallreedjr: You seem fond of guard clauses, but this one seems unnecessary. Generally speaking, returning explicit `true` and `false` values with an `if` condition is extraneous. Consider this alternative:
```
def won?
  winner != nil
end
```
which could be further simplified to
```
def won?
  winner
end
```
You could also use `!!winner` if you wanted to guarantee a boolean return value.

randallreedjr:
@HarlemSquirrel Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

bin/tictactoe
HarlemSquirrel: Ah, I see! That makes a lot of sense. Thank you!

lib/board.rb
HarlemSquirrel: Wow, that is cool! A very nice way to do this.

lib/game.rb
HarlemSquirrel: That is definitely cleaner.

lib/game.rb
HarlemSquirrel: I was having some issues at one point and lost faith in my logic for a moment. This is better.

HarlemSquirrel:
Thank you for all the feedback! I learned a bunch!

PR #55, opened 2016-01-28
Done

preetdha:


bin/tictactoe
```diff
@@ -1,3 +1,137 @@
...
+                                        ),_ ,   |
+                                        |  / )  |
+                                        | /   ) |
+                                        ()     ()
+                                        //     ||
+                                       //      ||
+                                      //       ||
+                        -- '' -'-'  ,//        /)   -- '' -'- 
+
+      OSTRICH
```
randallreedjr: But can your ostrich do this?
Ostriches are good skiers

bin/tictactoe
```diff
@@ -1,3 +1,137 @@
...
+        input = gets.strip.downcase
+        if input != "yes" || input != "y"
+           break
+         end
+      end
+      puts "See you next time!"
+    end
+
+## Begin CLI ##
+  play_tic
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+      @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+require_relative '../lib/game.rb'
+require_relative '../lib/players/computer.rb'
+require_relative '../lib/players/human.rb'
+require_relative '../lib/board.rb'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    cells.select {|cell| cell == " "}.empty?
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def turn_count
+    cells.count {|cell| cell != " "}
+  end
+
+  def taken?(position)
+    cells[(position.to_i - 1)] != " " ? true : false
```

```
def taken?(position)
  cells[(position.to_i - 1)] != " "
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def valid_move?(position)
+    position.to_i.between?(1, 9) && taken?(position) == false ? true : false
```
randallreedjr: Avoid doing equality comparisons with literal boolean values. Here, you can instead reverse (or negate) the boolean with the `!` (bang) operator.

`position.to_i.between?(1,9) && !taken?(position)`

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def board
+    @board
+  end
```
randallreedjr: With `attr_accessor :board`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def board
  @board
end
```
and
```
def board=(board)
  @board = board
end
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def over?
+    draw? || won?
+  end
+
+  def won?
+    WIN_COMBINATIONS.detect do |combo|
+     combo.all? {|position| board.cells[position] == "X"} || combo.all? {|position| board.cells[position] == "O"}
+    end
+  end
```
randallreedjr: This is one of the simpler implementations of this method I've seen. Very clear, nice job.

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+  def draw?
+    board.full? && !won?
+  end
+
+  
+  def winner
+   if won?
+     board.cells[won?.first]
+   end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Player::Human < Player
+  attr_accessor :board
+
+  def move(board)
+    puts "Please provide a number from 1-9"
+    input = gets.strip
+  end
```
```
def move(board)
  puts "Please provide a number form 1-9"
  gets.strip
end
```

randallreedjr:
@preetdha Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #59, opened 2016-01-31
Learn review

randallreedjr:


bin/tictactoe
```diff
@@ -1,3 +1,27 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "What's updog?"
```
randallreedjr: Do you smell updog?

bin/tictactoe
```diff
@@ -1,3 +1,27 @@
...
+
+if input == "1"
+  game = Game.new(Human.new("X"), Computer.new("O"))
+  game.play
+end
+
+if input == "2"
+  game = Game.new(Human.new("X"), Human.new("O"))
+  game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

df
```diff
@@ -0,0 +1,21 @@
...
+    [1;34m103[0m:       (board.cells[win_combo[[1;34m1[0m]] == [31m[1;31m"[0m[31mX[1;31m"[0m[31m[0m && board.cells[win_combo[[1;34m2[0m]] == [31m[1;31m"[0m[31mX[1;31m"[0m[31m[0m) &&
+    [1;34m104[0m:       !board.taken?(win_combo[[1;34m0[0m]+[1;34m1[0m)
+    [1;34m105[0m:     ) ||
+    [1;34m106[0m:     (
+    [1;34m107[0m:       (board.cells[win_combo[[1;34m0[0m]] == [31m[1;31m"[0m[31mX[1;31m"[0m[31m[0m && board.cells[win_combo[[1;34m2[0m]] == [31m[1;31m"[0m[31mX[1;31m"[0m[31m[0m) &&
+    [1;34m108[0m:       !board.taken?(win_combo[[1;34m1[0m]+[1;34m1[0m)
+    [1;34m109[0m:     )
+    [1;34m110[0m:   [32mend[0m
+    [1;34m111[0m: [32mend[0m
+
```
randallreedjr: (Don't worry about this file)

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def position(location)
+    cells[location.to_i-1]
+  end
+
+  def full?
+    !cells.include?(" ")
+  end
+
+  def turn_count
+    cells.size - cells.count(" ")
```
randallreedjr: Interesting approach, nice.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def taken?(location)
+    position(location) != " "
+  end
+
+  def valid_move?(location)
+    if location.to_i > 0 && location.to_i < 10
+      !taken?(location)
+    else
+      false
```
randallreedjr: `else false` is unnecessary. If the condition of the `if` statement is not satisfied, the `if` expression will return `nil`, which is also falsey. This would be equivalent:
```
def valid_move?(location)
  if location.to_i > 0 && location.to_i < 10
    !taken?(location)
  end
end
```
Also, you could simplify the location range check using [`between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
`location.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won[0]]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,139 @@
+class Computer < Player
+
+  def move(board)
```
randallreedjr: This method is reallllllly long. Consider breaking it up into some smaller chunks.

lib/players/computer.rb
```diff
@@ -0,0 +1,139 @@
...
+  def corner_or_side(board)
+    if (board.cells[0] == board.cells[8] && board.cells[0] != " ") || (board.cells[2] == board.cells[6] && board.cells[2] != " ")
+      [1,3,5,7].detect{|cell| !board.taken?(cell+1)}
+    else
+      [0,2,6,8].detect{|cell| !board.taken?(cell+1)}
+    end
+  end
+
+  def react(board)
+    win(board) || block(board) || corner_or_side(board)
```
randallreedjr: Excellent, readable method.

randallreedjr:
@morgvanny Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #60, opened 2016-01-31
Done

alexbarron:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def position(pos)
+    return @cells[pos.to_i-1]
+  end
+
+  def full?
+    result = @cells.detect{|pos| pos == " "}
+    result == " " ? false : true
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def turn_count
+    result = @cells.select{|pos| pos == "X" || pos == "O"}
+    return result.count
+  end
```
randallreedjr: Again, consider choosing a more semantic iterator. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def taken?(pos)
+    if @cells[pos.to_i-1] == " "
+      return false
+    else
+      return true
+    end
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the `if`/`else` statements. You can also reverse (or negate) the boolean with the `!` (bang) operator.

```
def taken?(pos)
  !@cells[pos.to_i-1] == " "
end
```
or equivalently `@cells[pos.to_i-1] != " "`

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def valid_move?(pos)
+    pos = pos.to_i
+    if pos >= 1 && pos <= 9 && !taken?(pos)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,138 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    return @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def over?
+    draw? || won? ? true : false
```
randallreedjr: A ternary operator where both results are literal boolean values is redundant; just return the value of the boolean expression.
```
def over?
  draw? || won?
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Enter a position 1-9: "
+    pos = gets.chomp
+    return pos
+  end
```
```
def move(board)
  puts "Enter a position 1-9: "
  gets.chomp
end
```

randallreedjr:
@alexbarron Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

alexbarron:
@randallreedjr Thank you for the review! Very helpful feedback. I've gone ahead and implemented all your feedback and pushed a new commit. I won't submit another pull request for this though.

PR #61, opened 2016-02-01
Done

AdamKing0126:


bin/tictactoe
```diff
@@ -1,3 +1,28 @@
...
+  player_1 = Human.new('X')
+  player_2 = Computer.new('O')
+  game = Game.new(player_1, player_2)
+  game.play
+when '3'
+  game = Game.new
+  game.play
+else
+  puts "Incorrect input."
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts " ----------- "
+    puts "  #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def turn_count
+    9 - @cells.count(" ")
+  end
+
+  def taken?(position)
+    @cells[position.to_i - 1] == " " ? false : true
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. Additionally, you can reverse (or negate) the boolean with the `!` (bang) operator.

`@cells[position.to_i - 1] != " "`

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+  attr_reader :winner
+
+  def initialize(player_1 = Human.new('X'), player_2 = Human.new('O'), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2 
+  end
+
+  def over?
+    (@board.full? || won? || draw?) ? true : false
```
randallreedjr: Won't the board always be full if the game is a draw? I think you could reduce this logically to
`@board.full? || won?`

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+          return true
+        end
+      end
+    end
+    false
+  end
+
+  def draw?
+    return true if @board.full? && (self.won? == false)
+    false
```
randallreedjr: This method always either returns `true` (if a certain condition is met) or `false`. Can you rewrite it to utilize the implicit return value of the boolean expression?

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+  def move(board)
+    board.display
+    puts "What's your move?"
+    input = gets.chomp
+    input
```
```
def move(board)
  board.display
  puts "What's your move?"
  gets.chomp
end
```

spec/06_computer_ai_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it 'returns three critical combinations' do
+      board = Board.new
+      board.cells = ["X", "O", "O", " ", " ", "O", "X", " ", "X"]
+      potentials = [2, 3, 6]
+      computer = Player::Computer.new("X")
+
+      expect(computer.get_critical_combinations("X",potentials, board)).to eq([2, 3, 6])
+    end
+  end
+end
```
randallreedjr: I love that you wrote your own tests for the AI!

randallreedjr:
@AdamKing0126 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

AdamKing0126:
Hi Randall,

Thanks very much for your comments.  I really appreciate it.  I've been
steaming at full speed through SQL and ORM but plan to take a look through
my code and your advised edits this afternoon.

Have a good one!

PR #62, opened 2016-02-01
Project submitted

azemoh:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
...
+  when "3"
+    two_players
+  when "help"
+    help
+  when "exit"
+    break
+  else
+    puts "Invalid! type 'help' for help.\n\n"
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,50 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,50 @@
...
+  def over?
+    draw? || won?
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,138 @@
+class Computer < Player
+  include Concerns::Winable
+  attr_reader :board
+
+  INFINITY = 1.0/0.0
```
randallreedjr: You can also use `Float::INFINITY` and `-Float::INFINITY`.

lib/players/computer.rb
```diff
@@ -0,0 +1,138 @@
...
+    score += evaluate_combo(WIN_COMBINATIONS[4])
+    score += evaluate_combo(WIN_COMBINATIONS[5])
+    score += evaluate_combo(WIN_COMBINATIONS[6])
+    score += evaluate_combo(WIN_COMBINATIONS[7])
+  end
+
+  # Evaluate board with a wining combination
+  # Return +100, +10, +1 for 3, 2, 1 in a row for self
+  # Return -100, -10, -1 for 3, 2, 1 in a row for opponent
+  # Return 0 for empty cell
```
randallreedjr: Very interesting heuristic

randallreedjr:
@azemoh Great job, very impressive! I left you a couple comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

lib/game.rb
azemoh: I think I wanted the expression in one line at the time. I need to learn when to avoid one liners.

lib/players/computer.rb
azemoh: I just realised i should use `WIN_COMBINATIONS.each` here.

lib/players/computer.rb
azemoh: Neat.

azemoh:
Thanks for the review @randallreedjr, your feedback was very helpful. I have made changes appropriately.

lib/game.rb
randallreedjr: You could still keep the one-liner:

`board.cells[won.first] if won = won?`

azemoh: Wow that's nice. Hey @randallreedjr, as you can see I do that alot, and I am wondering at what point does it become a bad thing to do. Trading readability for aesthetics. I want to check myself before it sticks.

randallreedjr: @azemoh Generally, one-line methods are great in ruby. I tend to stick to a couple of guidelines:

1.  Try to avoid any lines that get too long, usually ~80 characters (according to the [Ruby Styleguide](https://github.com/styleguide/ruby)). This can sometimes be tricky with long variable names, and I tend to err on the side of clarity rather than strict adherence to this "rule"

1.  Avoid single line methods that are too complex. This is something that is more subjective and you will get a better feel for the more you write in Ruby. Broadly speaking, one to two operations are fine, and three is borderline. But more basically, if you come back to a single line method that you wrote and you scratch your head for a second, it would probably be better to rewrite it for greater clarity.

Working in development teams that change a lot, I have come to value readability of code very highly.

azemoh: Thanks @randallreedjr I'll try to remember that and you are awesome!

PR #63, opened 2016-02-01
Project submitted

pickledyamsman:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic-Tac-Toe"
+game = Game.start
+
+until game.over?
+  game.play
+end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
+class Board
+
+  attr_accessor :cells
+
+  @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
randallreedjr: Avoid setting instance variables outside of a method. When you instantiate a new Board instance, `cells` will not get set by this code.
```
$ Board.new.cells
# => nil
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def initialize
+    reset
+  end
+
+  def reset
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
randallreedjr: This method is redundant. Prefer to use the bang method `reset!` which clearly communicates that internal state is changed by calling  the method.

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def turn_count
+    counter = 0
+    @cells.each do |cell|
+      if cell == "X" || cell == "O"
+        counter += 1
+      end
+    end
+    counter
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def taken?(move)
+    if @cells[move.to_i-1] == "X" || @cells[move.to_i-1] == "O"
+      true
+    elsif @cells[move.to_i-1] == " " || @cells[move.to_i-1] == ""
+      false
+    end
```
randallreedjr: Is there any other possible value for `@cells[move.to_i-1]` that would belong in an `else` clause? If so, would the return value (`false`) change?
I propose that you could replace your `elsif` with `else` and this method would be logically equivalent.
```
def taken?(move)
  if @cells[move.to_i-1] == "X" || @cells[move.to_i-1] == "O"
    true
  else
    false
  end
end
```
Now, generally you want to avoid a dichotomous method that returns boolean literals based on a single condition. We could utilize the implicit return value of the boolean expression and simplify the updated method to:
```
def taken?(move)
  @cells[move.to_i-1] == "X" || @cells[move.to_i-1] == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def valid_move?(pos)
+    true if pos.to_i.between?(1,9) && taken?(pos) == false
```
randallreedjr: Also, avoid equality comparison with boolean literals as it is always unnecessary. Instead, you can use the bang (`!`) operator to reverse or negate the boolean result.
```
def valid_move?(pos)
  pos.to_i.between(1,9) && !taken?(pos)
end
```

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def draw?
+    @board.full? && !won? ? true : false
+  end
+
+  def winner
+    if winner = won?
+      winner.token
+    else
+      nil
```
randallreedjr: In ruby, an `if` statement whose condition is not satisfied returns `nil`. You could remove this `else` clause and your code would behave the same.

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+  end
+
+  def turn
+    p = current_player.move(@board)
```
randallreedjr: `p` is a method in ruby equivalent to `puts <variable>.inspect`. You should avoid using this for a local variable name as it may be confusing to other programmers looking at your code.

randallreedjr:
@pickledyamsman Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #64, opened 2016-02-02
Learn review

randallreedjr:


bin/tictactoe
```diff
@@ -1,3 +1,13 @@
...
+
+puts "Welcome to Tic Tac Toe!"
+
+game = Game.new
+
+until game.won?
+
+  game.play
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+
+	attr_accessor :cells
+
+	def initialize
+		@cells = [' ',' ',' ',' ',' ',' ',' ',' ',' ']		
+	end
+
+	def reset!
+		@cells.clear
+		@cells = [' ',' ',' ',' ',' ',' ',' ',' ',' ']
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+require 'pry'
+
+class Board
+
+	attr_accessor :cells
+
+	def initialize
+		@cells = [' ',' ',' ',' ',' ',' ',' ',' ',' ']
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+	def position(value)
+		@cells[value.to_i - 1]
+	end
+
+	def full?
+		!@cells.include?(' ')
+	end
+
+	def turn_count
+		cells.count {|cell| cell == "X" || cell == "O"}
```
randallreedjr: Excellent choice of iterator

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+	end
+
+	def taken?(space)
+		@cells[space.to_i - 1] == "X" || @cells[space.to_i - 1] == "O"
+	end
+
+	def valid_move?(space)
+		space.to_i < 10 && space.to_i > 0 && !taken?(space)
```
`space.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+	WIN_COMBINATIONS = [[0,1,2], [3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[2,4,6]]
+
+	def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board=Board.new)
+		@player_1 = player_1
+		@player_2 = player_2
+		@board = board
+	end
+
+	def current_player
+		@board.turn_count % 2 == 0 ? @player_1 : @player_2
```

randallreedjr:
@coltonstaab1 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #65, opened 2016-02-03
Done

kriskanya:


bin/tictactoe
```diff
@@ -1,3 +1,54 @@
...
+    game.player_2 = Player::Human.new(second_player)
+    game.play()
+
+  else
+    puts "invalid number of players"
+    play_ttt()
+  end
+end
+
+play_ttt()
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
+require 'pry'
+
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    position = position.to_i - 1
+    @cells[position]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
```
randallreedjr: Use the implicit boolean return value of the expression instead of adding the ternary. You can reverse (or negate) the boolean with the `!` (bang) operator.

`!@cells.include?(" ")`

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def valid_move?(position)
+    pos = position.to_i - 1
+
+    if taken?(position) == false && pos.between?(0, @cells.length-1)
+      true
+    else
+      false
```
randallreedjr: Here again you can just use the implicit return value of the boolean expression instead of adding the `if` statement.
```
def valid_move?(position)
  pos = position.to_i - 1
  !taken?(position) && pos.between?(0, @cells.length-1)
end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  # this is the current player whose move it currently is
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+    return false
+  end
+
+  def draw?
+    won? == false && board.full?
+  end
+
+  def winner
+    won? ? @board.cells[won?[0]] : nil
```
randallreedjr: Typically, a ternary should not return nil. You can accomplish the same result with an inline if statement.
`@board.cells[won?[0]] if won?`
However, in this case, you may want to use an if statement assignment to avoid calling the `won?` method twice.
```ruby
if won = won?
  @board.cells[won[0]]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,91 @@
...
+    # computer wants to make in the form of a 1-9 string.
+    if available_spaces(board).include?(5)
+      "5"
+    else
+      best_move(board)
+    end
+  end
+
+  def best_move(board)
+    win(board) || block(board) || corner_move(board) || random_move(board)
```
randallreedjr: Nice method, very readable

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    position = position.to_i - 1
+    @cells[position]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
+  end
+
+  def turn_count
+    @cells.count do |cell|
```
randallreedjr: Great choice of iterator

lib/players/computer.rb
```diff
@@ -0,0 +1,91 @@
...
+        if !board.taken?(combo_array_item+1)
+          return combo_array_item+1
+        end
+      end
+    end
+  end
+
+  def random_move(board)
+    random_number = rand(0..8)
+    if !board.taken?(rand(0..8)+1)
```
randallreedjr: This seems dangerous. You are generating two random numbers here, let's call them `a` and `b`; then saying if `b` is not taken on the board, move in position `a`. But what if `a` is taken? I think you'd want to rewrite as follows:
```
def random_move(board)
  random_number = rand(0..8)
  if !board.taken?(random_number + 1)
    random_number
  else
    random_move(board)
  end
end
```

randallreedjr:
@kriskanya Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

kriskanya:
@randallreedjr Thanks!  I will look at these more thoroughly and refactor.

PR #68, opened 2016-02-04
Project submitted

mystycs:


randallreedjr:
Hi @mystycs. There were some extra files included in your submission that make it difficult to review your code. Could you remove the `.bundle` directory and resubmit?
Alternatively, you can add me as a collaborator to your repo and I'll take care of it for you.

* Visit https://github.com/@mystycs/ttt-with-ai-project-v-000/settings/collaboration and add `randallreedjr` in the 'Collaborator' section.

Thanks!

mystycs:
Hey there @randallreedjr  i went ahead and removed the .bundle directory.

Thanks!

bin/tictactoe
```diff
@@ -1,3 +1,8 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
+puts "Please enter 0, 1 or 2 for number of human players."
+input = gets.chomp.to_i
+Game.new.player_count(input)
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def turn_count
+    counter = 0
+      @cells.each do |move|
+      if (move == "X" || move == "O")
+        counter += 1
+      end
+    end
+    return counter
```

```ruby
def turn_count
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Please enter your move from 1 - 9"
+    input = gets.chomp
+  end
```
```
def move(board)
  puts "Please enter your move from 1 - 9"
  gets.chomp
end
```

randallreedjr:
@mystycs Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #69, opened 2016-02-05
Learn review

randallreedjr:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+ttt = TttCli.new
+ttt.play_game
```
randallreedjr: Excellent executable file

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+require 'pry'
+class Board
+
+
+	attr_accessor :cells, :edges, :corners, :center
+
+	def initialize
+		@cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+	end
+
+	def full?
+		@cells.all? {|cell| cell == "X" || cell == "O" }
+	end
+
+	def turn_count
+		turns = @cells.select {|cell| cell == "X" || cell == "O"}
+		turns.size
+	end
```

```ruby
def turn_count
  @cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+
+	def taken?(position)
+		position(position) == "X" || position(position) == "O"
+	
+	end
+
+	def valid_move?(move)
+		move.to_i > 0 && move.to_i < 10 && !taken?(move.to_i)
+	end
```
`move.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+   @player_2 = player_2
+  end
+
+  def ai_vs_ai
+    @player_1 = Player::Computer.new("X")
+    @player_2 = Player::Computer.new("O")
+  end
+
+  def ai_vs_human(choice) ## -->
+    if  choice == "X" || choice == "x"
```
randallreedjr: You could also use [`downcase`](http://ruby-doc.org/core-2.2.4/String.html#method-i-downcase) here.
`if choice.downcase == "x"`

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+      @player_1 = Player::Computer.new("X")
+    end
+  end
+
+  def over?
+   won? || draw?
+  end
+
+  def current_player
+   @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  def winner
+   if winning_combo = won?
```
randallreedjr: Good use of if assignment.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
...
+	attr_accessor :game
+	attr_reader :token
+
+	def initialize(token)
+		@token = token
+	end
+
+	def game(game)
+		@game = game
+	end
```
randallreedjr: This method is a little confusing. By convention, a `game=` method would be used to set the `@game` instance variable. However, that method is already provided by `attr_accessor :game`. I would think you could remove this code.

randallreedjr:
@jj54321 @corya0687 Great job! I left you both a few comments. One other thing I would encourage you to do is double check your formatting. Consistently using 2 spaces, not tabs, for indentation will make the code easier to read.

Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #70, opened 2016-02-05
Done

tejbans:


bin/tictactoe
```diff
@@ -1,3 +1,50 @@
...
+          
+
+    elsif kind =="0"
+            
+            Game.new(Player::Computer.new("X"), Player::Computer.new("O"), Board.new).play
+            puts "Do you want to play again (y/n)?."
+            play_again = gets.strip
+    end
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+require 'pry'
+class Board
+  attr_accessor:cells
+
+   def initialize
+    reset!
+   end
+
+   def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+   end
+
+   def full?
+    self.cells.all?{|token| token == "X" || token == "O"}
+  end
+
+  def turn_count
+    self.cells.count{|t| t=="X" || t=="O"}
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+   def taken?(pos)
+    !(position(pos).nil? || position(pos) == " ")
+   end
+
+   def valid_move?(pos)
+    pos.to_i.between?(1,9) && taken?(pos) == false
```
randallreedjr: Avoid direct equality comparison with literal boolean values. You can reverse (or negate) the boolean with the `!` (bang) operator.

`pos.to_i.between?(1,9) && !taken?(pos)`

lib/players/computer.rb
```diff
@@ -0,0 +1,41 @@
...
+      self.token == "X" ? "O" : "X"
+   end
+ 
+   def move(board)
+     return "5" if !board.taken?(5)
+     best_option(board)
+   end
+ 
+   def best_option(board)
+     win_move(board) || block_move(board) || strategic_move(board)
```
randallreedjr: Nice method, very readable

randallreedjr:
@tejbans Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

tejbans:
Hello Randall,
	Thanks for the comments.
	I tried to refactor the code per comments and resubmit the project, but wasn’t able to?.
	Do I have to re-submit ?

thanks
Tejbans

randallreedjr:
@tejbans You don't have to resubmit, the comments are just for your edification. However, if you would like to, you can try manually opening a new pull request from your github repo.

PR #71, opened 2016-02-06
Done

tjnz:


bin/tictactoe
```diff
@@ -1,3 +1,33 @@
...
+		play
+	when 'n'
+		exit
+	else
+		puts "\nPlease enter 'y' or 'n'"
+		play_again?
+	end
+end
+
+play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,193 @@
...
+			:draws => 0
+		}
+	end
+	
+	def current_player
+		board.turn_count.even? ? player_1 : player_2
+	end
+	
+	def over?
+		won? ? true : draw?
```
randallreedjr: Typically, a ternary operator returning a boolean literal is unnecessary. Here, you could simply say
```
def over?
  won? || draw?
end
```
The or operator `||` is short-circuited, meaning it will return true as soon as one of the expressions returns true.

lib/game.rb
```diff
@@ -0,0 +1,193 @@
...
+			false
+		end
+	end
+	
+	def draw?
+		!board.full? ? false : !won?
+	end
+	
+	def winner
+		won? ? the_winner : nil
```
randallreedjr: When you see the pattern of a ternary operator returning `nil`, you can replace it with an inline if statement, which has the same effect.
`the_winner if won?`

randallreedjr:
@tjnz Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

tjnz:
Great thank you! This was very helpful.

PR #73, opened 2016-02-07
Project submitted

gormanjp:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
...
+		puts "Would you like to play again?"
+		yn = gets.chomp.downcase
+		if yn == "yes"
+			go
+		else
+			puts "Thanks for playing!"
+		end
+	end
+
+	go
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+
+	attr_reader :cells
+
+	def initialize
+		self.reset!
+	end
+
+	def cells=(cells)
+		@cells = cells
```
randallreedjr: Since the behavior you are defining here is the default behavior for the `cells=` method, you could just change your `attr_reader` to `attr_accessor`.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	end
+
+	def reset!
+		@cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	end
+
+	def full?
+		self.cells.all? {|cell| cell != " "}
+	end
+
+	def turn_count
+		count = 0
+		self.cells.each {|cell| count+=1 unless cell == " "}
+		count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	end
+
+	def taken?(cell)
+		self.position(cell) == "X" || self.position(cell) == "O" 
+	end
+
+	def valid_move?(move)
+		move = move.to_i
+		move > 0 && move < 10 && self.taken?(move) == false
```

`move.between?(1,9)`

Also, avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression. Here, you can reverse (or negate) the boolean with the `!` (bang) operator.

`!self.taken?(move)`

lib/game.rb
```diff
@@ -0,0 +1,108 @@
...
+	 WIN_COMBINATIONS = [[0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[2,4,6]]
+	 
+	 def initialize(p1 = Player::Human.new("X"), p2 = Player::Human.new("O"), board = Board.new)
+	 	@board = board
+	 	@player_1 = p1
+	 	@player_2 = p2
+	 end
+
+	 def turn_count
+	 	board.cells.count(player_1.token) + board.cells.count(player_2.token)
```
randallreedjr: Why not reuse `board.turn_count` here?

lib/game.rb
```diff
@@ -0,0 +1,108 @@
...
+	 end
+
+	 def current_player
+	 	if turn_count == 0 || turn_count.even?
```
randallreedjr: Zero is an even number (or at least treated as such in Ruby) and doesn't need to be checked for separately.

lib/game.rb
```diff
@@ -0,0 +1,108 @@
...
+
+
+
+
+
+
+
+
+
+
```
randallreedjr: Look at all this whitespace... So much room for activities!

lib/player.rb
```diff
@@ -0,0 +1,12 @@
+class Player
+
+	attr_reader :token
+
+	def initialize(token)
+		@token = token
+	end
+
+	def token
+		@token
```
randallreedjr: With `attr_reader :token`, this method is not needed. `attr_reader` is a macro that automatically adds the getter method to your class for the selected attribute. In this case,
```
def token
  @token
end
```
has already been included in your class through the `attr_reader`.

lib/players/computer.rb
```diff
@@ -0,0 +1,84 @@
+class Player::Computer < Player
+
+	 COMBINATIONS = [[0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[2,4,6]]
+
+	def move(board)
```
randallreedjr: This method is very long. Consider breaking it up into smaller helper methods (your comments delineate possible helper methods well).

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+class Player::Human < Player
+
+	def move(board)
+		puts "Where would you like to move?"
+		input = gets.chomp
+		input
```
```
def move(board)
  gets.chomp
end
```

randallreedjr:
@gormanjp Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #75, opened 2016-02-08
Project submitted

jenjiyi:


bin/tictactoe
```diff
@@ -1,3 +1,86 @@
...
+   _.-._
+  | | | |_
+  | | | | |
+  | | | | |
+_ |  '-._ |
+\\`\\`-.'-._;
+ \\    '   |
+  \\  .`  /
+   |    |\n
+    BYE!\n\n\n"
```
randallreedjr: Love the ASCII art

bin/tictactoe
```diff
@@ -1,3 +1,86 @@
...
+          game_play
+        end
+      end  
+    end
+
+    play_again?
+
+end
+
+game_play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+ # Finally, a board can return values based on its state such as
+  def full?
+    cells.all?{ |i| i == "X" || i == "O"}
+  end
+
+  def turn_count
+    #turn_count based on how many positions in the cells array are filled.
+    cells.count{ |i| i == "X" || i == "O" }
+  end
```
randallreedjr: Excellent choice of iterators for both `full?` and `turn_count`. Very expressive.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # def taken?(position)
+  #   position = position.to_i-1
+  #   #taken? will return true or false for an individual position. 
+  #   if @cells[position] == "X" || @cells[position] == "O" 
+  #    true 
+  #   elsif @cells[position] == " " || @cells[position] = "" || @cells[position] = "  "
+  #     false
+  #   end
+  # end
```
randallreedjr: It's a good habit to develop to delete commented-out code like this before submitting a pull request.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+    def taken?(input)
+    !(position(input)  == " " || position(input) == "" || position(input) == "  ")
```
randallreedjr: Instead of checking for 0, 1, or 2 spaces, you could try something like `position(input).strip.empty?`

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+     
+  end
+
+  def valid_move?(input) 
+   #valid_move? will ensure that moves are between 1-9 and not taken
+   taken?(input) == false && (input.to_i-1).between?(0,8)
```
randallreedjr: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression. You can reverse (or negate) the boolean with the `!` (bang) operator.

`!taken?(input)`

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+    [0,3,6], # Left column
+    [1,4,7],  # Middle column
+    [2,5,8],  # Right column
+    [0,4,8],  # Left top diagonal
+    [6,4,2]  # Right top diagonal 
+  ]
+
+    #returns current player based on Board #turn_count
+  def current_player
+    if @board.turn_count % 2 == 0
```

`if @board.turn_count.even?`

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+  def current_player
+    if @board.turn_count % 2 == 0 
+      player_1 # return "X"
+    else
+      player_2 # return "O"
+    end
+  end
+
+  def over?
+    won? || @board.full? || draw?
```
randallreedjr: One of these expressions is redundant. If the board is full, either someone `won?` or the game is a `draw?`. Furthermore, the `draw?` method already calls `@board.full?`. You could use either `won? || @board.full?` or `won? || draw?`, depending on whether you value performance or readability more, respectively.

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cats Game!"
+      puts"
+      /\\___/\\
+     ( o   o )
+     (  =^=  )
+     (        )
+     (         )
+     (          )))))))))))\n"
```
randallreedjr: Nice touch!

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+  end
+
+  def winner
+    if winner_combo = won?
```
randallreedjr: Great use of if-assignment.

lib/players/human.rb
```diff
@@ -0,0 +1,27 @@
+class Player::Human < Player
+  def move(board)
+    # takes in a board argument and allows a human player to enter a move via the CLI 
+    puts "What's your move player #{token}?"
+    input = gets.strip
+     # return the value the user enters
+    if input == "quit" || input == "q" || input == "Quit" || input == "Exit" || input == "exit"
```
randallreedjr: I really like that you are checking for so many possible exit commands. One thing that might make this simpler would be to put them in an array.
```
exit_commands = %w(quit q exit)
if exit_commands.include?(input.downcase)
  puts "Ending game.\n"
  ...
```
You could also make `EXIT_COMMANDS` a constant at the top of your class.

randallreedjr:
@jenjiyi Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #76, opened 2016-02-09
Project submitted

drywaterr24:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
...
+  puts "Who would you like to be 'X'? Type 1 for Human or 2 for AI"
+  char = gets.chomp
+  if char == "1"
+    Game.new(Human.new("X"), Computer.new("O")).play
+  elsif char == "2"
+    Game.new(Computer.new("X"), Human.new("O")).play
+  end
+when "3"
+  Game.new.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+  require 'pry'
+  attr_accessor :cells, :board
+
+  def initialize
+    @board = Array.new(9, " ")
+    @cells = @board
```
randallreedjr: Not sure why you're using two different instance variables to store the same values.

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+  require 'pry'
+  attr_accessor :cells, :board
+
+  def initialize
+    @board = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def turn_count
+    count = 0
+
+    self.cells.each do |cell|
+      if cell == "X" || cell == "O"
+        count += 1
+      end
+    end
+    count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def taken?(num)
+    self.cells[num.to_i - 1].eql?("X") || self.cells[num.to_i - 1].eql?("O")
+  end
+
+  def valid_move?(num)
+    !self.taken?(num) && true if num.to_i > 0 && num.to_i <= 9
```
`num.to_i.between?(1,9)`

lib/players/computer.rb
```diff
@@ -0,0 +1,131 @@
...
+    elsif board.cells == [" ", " ", " ", " ", "X", " ", " ", " ", " "]
+      9
+    elsif board.cells == [" ", " ", " ", " ", " ", "X", " ", " ", " "]
+      5
+    elsif board.cells == [" ", " ", " ", " ", " ", " ", "X", " ", " "]
+      5
+    elsif board.cells == [" ", " ", " ", " ", " ", " ", " ", "X", " "]
+      5
+    elsif board.cells == [" ", " ", " ", " ", " ", " ", " ", " ", "X"]
+      5
```
randallreedjr: Couldn't these 9 `elsif` statements be condensed into `elsif board.cells.count("X") == 1`

lib/players/computer.rb
```diff
@@ -0,0 +1,131 @@
...
+    elsif board.cells[1] == board.cells[2] && board.cells[0] == " "
+      1
+    elsif board.cells[1] == board.cells[4] && board.cells[7] == " "
+      8
+    elsif board.cells[2] == board.cells[4] && board.cells[6] == " "
+      7
+    elsif board.cells[2] == board.cells[5] && board.cells[8] == " "
+      9
+    elsif board.cells[2] == board.cells[1] && board.cells[0] == " "
+      1
```
randallreedjr: This long list of enumerations is difficult to read and maintain. Would it be possible to break them up into some smaller methods, such as `opening_move`, `vertical_win`, `horizontal_block`, etc.?

lib/players/untitled
```diff
@@ -0,0 +1,101 @@
+require 'pry'
+class Computer < Player
```
randallreedjr: Should this file have been excluded?

randallreedjr:
@drywaterr24 Great job! I left you a few comments. Closing this PR for now, but feel free to re-open and respond if you would like to discuss any of them.

PR #77, opened 2016-02-10
Project submitted

LKWLaLa:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+Game.start
```

lib/game.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def initialize(player_1 = Human.new("X"),player_2 = Human.new("O"),board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    counter = self.board.turn_count
+    if counter % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,148 @@
...
+    end
+  end
+
+  def over?
+    self.draw? ||  self.won? 
+  end
+
+      
+  def draw?
+   self.board.full? && self.won? == nil
```
randallreedjr: Avoid equality comparison with `nil`. Instead, you can use the [`.nil?`](http://ruby-doc.org/core-2.2.0/NilClass.html#method-i-nil-3F) method or simply `!self.won?`.

lib/game.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def won?
+      WIN_COMBINATIONS.detect do |line|
+        line.all?{|i| self.board.cells[i] == "X" } || line.all?{|i| self.board.cells[i] == "O"}
+      end
+    end
+
+  def winner
+    if won?
+      line = won?
```
randallreedjr: You can accomplish both of these steps using if-assignment.
```
if line = won?
  self.board.cells[line[0]]
end
```
If `won?` returns nil, the expression `line = won?` will also return nil and the body of the if statement will not be executed.

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(board)
+    puts "Your turn #{self.token}! Choose a cell number (1-9):"
+    input = gets.strip
```
```
def move(board)
  puts "Your turn #{self.token}! Choose a cell number (1-9):"
  gets.strip
end
```

randallreedjr:
@LKWLaLa Great job! I left you a few comments, feel free to respond. Closing this PR.

LKWLaLa:
@randallreedjr  Thank you so much for the feedback!  Will go over these notes.  Much appreciated!

PR #84, opened 2016-02-12
Project submitted

NStephenson:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+
+  def draw?
+    board.full? && !won? 
+  end
+
+  def winner
+    if won? && board.cells[won?[0]] == player_1.token
+      player_1.token
+    elsif won? && board.cells[won?[0]] == player_2.token
+      player_2.token
```
randallreedjr: In both cases, you are just returning whatever `board.cells[won?[0]]` equals. Isn't this equivalent?
```
if won?
  board.cells[won?[0]]
end
```

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+    puts "Hello, welcome to Tic Tac Toe!"
+    puts "Please select an amount of human players 0, 1 or 2."
+    loop do
+      input = gets.chomp
+      case input.downcase
+      when '0'
+        create_two_computers
+        play
+        restart_game
+        break
```
randallreedjr: I really like how readable this case statement is.

randallreedjr:
@NStephenson Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #86, opened 2016-02-13
Done

macgeek30:


randallreedjr:
Hi @macgeek30. Unfortunately, there were some extra files included in your submission that make it impossible to review your code. Could you please remove the `.bundle` directory and resubmit?

randallreedjr:
Hi @macgeek30, anything I can do to help?

macgeek30:
done

bin/tictactoe
```diff
@@ -1,3 +1,34 @@
...
+    call
+  elsif input == "N"
+    exit
+  else
+    puts "A Valid Choice !!"
+    again?
+  end
+end
+
+call
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  def taken?(position)
+    !(@cells[position.to_i - 1].nil? || @cells[position.to_i - 1] == " ")
+  end
+
+  def valid_move?(position)
+    if (taken?(position.to_i) == false) && (position.to_i.between?(1,9) == true)
+    true
+    else (taken?(position.to_i) == true) || (position.to_i.between?(1,9) == false)
+    false
+    end
```
randallreedjr: Look out for patterns like
```
if #some condition
  return true
else
  return false
end
```
It is generally preferable to rely on the implicit return value of the boolean expression you are evaluating rather than returning `true` or `false`. Similarly, you should usually avoid equality comparisons with those boolean literals. In this case, you could rewrite your method as
```
def valid_move?(position)
  !taken?(position.to_i) && position.to_i.between?(1,9)
end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+  ]
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      if (position1 == "X" && position2 == "X" && position3 == "X") || (position1 == "O" && position2 == "O" && position3 == "O")
+        return combo
+      end
+    end
+    return false
+  end
+
+  def over?
+    if self.board.full? || self.won? || self.draw?
```
randallreedjr: Are all three of these conditions necessary? Will `self.board.full?` ever be true without `self.won?` or `self.draw?` also being true.

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
+    if self.won?
+      self.board.cells[self.won?[0]]
```

```ruby
if won = self.won?
  self.board.cells[won[0]]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,10 @@
+require 'pry'
+
+class Computer < Player
+  def move(board)
+    sleep 0.3
```
randallreedjr: Player: "Computer, are you asleep?"
Computer: "No no, just thinking with my eyes closed."

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+  def move(board)
+    puts "#{self.token}, please choose a number between 1-9 to make your move."
+    player_move = gets.strip
+    player_move
```
```
def move(board)
  puts "#{self.token}, please choose a number between 1-9 to make your move."
  gets.strip
end
```

randallreedjr:
@macgeek30 Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #87, opened 2016-02-14
Done

michaelsimon:


bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+    end
+  elsif num_players == 2
+    game = Game.new(Player::Human.new("X"), Player::Human.new("O"))
+  end
+  game.play
+  num_players = -1
+  first=""
+  print "(P)lay Again or Exit?: "
+  status = gets.chomp!
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")
+    @cells
```
randallreedjr: The previous line, `@cells = Array.new(9, " ")` will already return `@cells` (the result of the assignment operation); you don't need to call `@cells` again separately.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  def turn_count
+    @cells.count {|turn| turn != " "}
+  end
+
+  def taken?(pos)
+    position(pos) != " "
+  end
+
+  def valid_move?(pos)
+    # binding.pry
```
randallreedjr: Good practice to remove commented lines like this before submitting pull requests.

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+  ]
+
+  def initialize(player_1 = Player::Human.new("X") , player_2 = Player::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ?  @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      if (val_position_0 == @player_1.token && val_position_1 == @player_1.token && val_position_2 == @player_1.token) or (val_position_0 == @player_2.token && val_position_1 == @player_2.token && val_position_2 == @player_2.token)
+        return win_combo
+      end
+    end
+    return false
+  end
+
+  def winner
+    @board.cells[won?[0]] if over? && won?
```
randallreedjr: `if won?` should be sufficient, instead of `if over? && won?`. If `won?` is true, `over?` will also be true, and `true && x` is the same as `x`.

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Please Make Your Move Player #{self.token}: "
+    position = gets
```
```
def move(board)
  puts "Please Make Your Move Player #{self.token}: "
  gets
end
```

randallreedjr:
@michaelsimon Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #88, opened 2016-02-15
Project submitted

brianmenard:


randallreedjr:
Hi @brianmenard. Unfortunately, there were some extra files included in your submission that make it impossible to review your code. Could you please remove the `.bundle` directory and resubmit?

brianmenard:
Hi @randallreedjr !  Should be good now, thanks!

bin/tictactoe
```diff
@@ -1,3 +1,52 @@
...
+  else
+    player1 = Human.new("O")
+    player2 = Human.new("X")
+  end
+end
+
+board = Board.new
+game = Game.new(player1, player2, board)
+
+game.play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [
+        " ", " ", " ",
+        " ", " ", " ",
+        " ", " ", " ",
+      ]
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  def valid_move?(input)
+    self.cells[(input.to_i)-1] == " " && input.to_i.between?(1,9)
+  end
+
+  def turn_count
+    count = 0
+    self.cells.each do |cell|
+      count += 1 if cell == "X" || cell == "O"
+    end
+    return count
```

```ruby
def turn_count
  self.cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def taken?(input)
+    self.position(input) == "X" || self.position(input) == "O" ? true : false
```
```
def taken?(input)
  self.position(input) == "X" || self.position(input) == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+      [0,4,8],
+      [6,4,2]
+    ]
+
+  def initialize(player_1 = nil, player_2 = nil, board = nil) #assumes players and board are already initialized
+    if player_1 == nil
+      @player_1 = Human.new("X")
+    else
+      @player_1 = player_1
+    end
```
randallreedjr: You can simplify these assignments using the or operator (||).
```
@player_1 = player_1 || Human.new("X")
```
If the parameter `player_1` is set, it will be assigned to the instance variable `@player_1`. However, if the parameter is not passed in and defaults to `nil`, then the second half of the expression will be evaluated, setting the value of `@player_1` to `Human.new("X")`

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "../player.rb"
+class Human < Player
+
+  def move(board)
+    puts "Enter a move (1-9):"
+    input = gets.strip
+    input
```
```
def move(board)
  puts "Enter a move (1-9):"
  gets.strip
end
```

randallreedjr:
@brianmenard Great job! I left you a few comments, feel free to respon. Closing this PR for now.

PR #89, opened 2016-02-15
Project submitted

lawrend:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+StartGame.new
```
randallreedjr: Perfect!

lib/board.rb
```diff
@@ -0,0 +1,86 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells=[" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells=[" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells=[" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def turn_count
+    counter = 0
+    cells.each do |spot|
+      if spot == "X" || spot == "O"
+      counter += 1
+      end
+    end
+    counter
```

```ruby
def turn_count
  cells.count {|spot| spot == "X" || spot == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def all_X
+    ex = []
+    cells.each_with_index do |p, i| 
+      if p == "X"
+        ex << i
+      end
+    end
+    ex
```
randallreedjr: Typically, the pattern of
```
instantiate new array (ex)
iterate through another array (cells) and update the new array (ex)
return new array (ex)
```
indicates that a higher level iterator can be used. This one's a little tricky, but I would recommend [`collect`](http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-collect) to first transform the array to contain the indices of the elements equaling 'X'.
```
cells.collect.with_index {|p,i| i if p == "X"}
```
Any elements that don't equal "X" will now be `nil`, something like `["X", "O", "X", " ", "X"," ","O",X","O"]` => `[0,nil,2,nil,4,nil,nil,7,nil]`. Then, to remove the nil elements, just use [`compact`](http://ruby-doc.org/core-2.3.0/Array.html#method-i-compact).
```
def all_x
  cells.collect.with_index {|p,i| i if p == "X"}.compact
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+class Human < Player
+  attr_accessor :human_move, :board
+
+  def move(board)
+    puts "Where do you want to move, #{self.token}? 1-9:"
+    @human_move = gets
+    human_move
```
randallreedjr: This last line isn't necessary; `human_move` will already be returned by the assignment operation in the previous line.

lib/start.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+  end
+
+  def token_setter(token)
+    if /x/i === token 
+      @player_token = "X"
+    elsif /o/i === token
+      @player_token = "O"
+    elsif token == "0"
+      puts "Uh, that's a zero.  I mean the letter.  Do you want the 'X' or the 'O'?"
```
randallreedjr: Haha

randallreedjr:
@lawrend Great job! I left you a few comments, feel free to respond. Closing this PR.

lawrend:
Thank you SO MUCH for the feedback!!!  

It's rare that I look at code a second time once I get that green text of the test passing, but I know that refactoring is a huge part of really developing skill.  

So I truly appreciate a chance to look back at the more straight-forward code and see how it could be better.  I don't think I'd ever even bothered to think about the .compact method on arrays because my head was already full whenever I saw it in the documentation :) 

And I just want someone else to know...my AI, yes it is hard-coded and it doesn't use minmax, BUT it will give you 100 ties out of 100 when you play computer vs. computer, and it is unbeatable when you play against it as a human.  

I spent WAY too long thinking about tic-tac-toe to not tell at least one other person that fact :) 

Thanks again!

Doug

randallreedjr:
Same here, and your AI looks strikingly similar to my first submission.

PR #90, opened 2016-02-16
Done

jaeku:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
...
+      game_loop
+    else
+      puts "Goodbye."
+      break
+    end
+  end
+
+end
+
+game_loop
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    9.times { @cells << " " }
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def reset!
+    @cells.clear
+    9.times { @cells << " " }
```
randallreedjr: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements. I think it may have slightly better performance than using the shovel operator.

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.all? { |cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    @cells.select { |cell| cell == "X" || cell == "O"}.count
```

```ruby
def turn_count
  @cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+      @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.taken?(combo[0] + 1) #@board.taken?(combo[0] + 1) to make sure it == "X" or "O" and not empty spaces, +1 for offsetting to user input
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    !won? || draw? ? nil : @board.cells[won?.first]
```
randallreedjr: Avoid returning `nil` when using the ternary operator. That is usually an indication that you can simplify your expression. In this case, make your last return value the default. It could look something like this
```
if !(!won? || draw?)
  @board.cells[won?.first]
end
```
Remember that, if the `if` statement criteria is not satisfied, this expression will return nil.

Next, we can simplify a little further by distributing the negation. This might be a little confusing, but `!(a || b)` <=> `!a && !b` (this property is one of [DeMorgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)). So, we can rewrite the updated method as
```
if won? && !draw?
  @board.cells[won?.first]
end
```
But reading that aloud, you may realize that `!draw?` is redundant; anytime `won?` returns true, `!draw?` will also return true.

That was a lot and I hope I didn't overwhelm you. Feel free to reach out to discuss more.

lib/players/computer.rb
```diff
@@ -0,0 +1,82 @@
+class Player::Computer < Player
+  def move(board)
+    #1. Make winning move
+    location = make_winning_move(board)
+
+    #2. Block winning move
+    location = block_winning_move(board) if location.nil?
```
randallreedjr: I think you could use `||=` here to the same effect.
```
location ||= block_winning_move(board)
```
The assignment will only occur if `location` is nil.

randallreedjr:
@jaeku Great job! I left you a few comments, including one very long one; feel free to respond. Closing this PR.

PR #92, opened 2016-02-17
Create ttt with AI

hscoder:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
...
+    elsif input.downcase != "exit"
+      puts "Invalid input. Try again."
+      game_loop
+    else
+      puts "Goodbye."
+      break
+    end
+  end
+
+end
```
aviflombaum: Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

aviflombaum: Additionally even in your example, when is the method you're defining even being evoked?

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset! # the #reset! method will run when object gets instantiated and clear the []
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")
```
aviflombaum: Nice use of `Array.new`

bin/tictactoe
hscoder: Avi, thanks for helping me keep what's private private with encapsulation. I'll implement the changes. Thx.

lib/board.rb
hscoder: Thx. :grin:

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+    !won? && full?
+  end
+
+  def full?
+    board.cells.all? { |token| token == "X" || token == "O" } # returns true if elements in cells [] have an "X" or "O"
+  end
+
+  def winner
+    if won?
+      win_combo = won?
```
randallreedjr: You can accomplish both of these steps in a single line of code.
```
if win_combo = won?
```
If `won?` returns false, the result of the assignment operation will also be false, so the interior of the if statement will not be executed. This pattern is sometimes referred to as 'if assignment.'

randallreedjr:
Great submission @hscoder! I left you one comment in addition to the feedback Avi had already provided. Closing this PR.

PR #93, opened 2016-02-19
Project submitted

jjlippman:


randallreedjr:
Hi @jjlippman. Is this project still in progress? Do you have any quesitons?

jjlippman:
body{font-family: Geneva,Arial,Helvetica,sans-serif;font-size:9pt;background-color: #ffffff;color: black;}Hi Randall,I submitted it a while ago. As of now, no questions.Are we going to have a pairing session over it?Thanks,Jim-----Original Message-----

randallreedjr:
@jjlippman I was going to look through the code and leave a few comments. Can you make sure your latest version is pushed.

jjlippman:
body{font-family: Geneva,Arial,Helvetica,sans-serif;font-size:9pt;background-color: #ffffff;color: black;}ok i just ran learn submit again-----Original Message-----

jjlippman:
body{font-family: Geneva,Arial,Helvetica,sans-serif;font-size:9pt;background-color: #ffffff;color: black;}Randall,actually i just made i slight change... going to push it again.thanks,Jim-----Original Message-----

bin/tictactoe
```diff
@@ -1,3 +1,63 @@
...
+  end
+end
+
+
+make_game(greeting)
+answer=play_again?
+
+while answer=="Y"
+  answer=play_again?
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def turn_count
+    count=0
+    cells.each do |c|
+      if c=="X" || c=="O"
+        count+=1
+      end
+    end
+    count
```

```ruby
def turn_count
  cells.count {|c| c == "X" || c == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def taken?(position)
+    p=position.to_i-1
+    if cells[p]=="X" || cells[p]=="O"
+        true
+    else
+        false
+    end
```
```
def taken?(position)
  p = position.to_i - 1
  cells[p] == "X" || cells[p] == "O"
end
```

P.S. Whitespace improves code readability.

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+  
+  def valid_move?(move)
+    if move==" "
+      return true
```
randallreedjr: Why is an empty string a valid move?

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    move=move.to_i
+     
+    move-=1
+
+    if (move>=0 && move <=9)
```
`position.between?(1,9)`

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    if move==" "
+      return true
+    end
+
+    move=move.to_i
+     
+    move-=1
+
+    if (move>=0 && move <=9) 
+      if taken?(move+1)
```
randallreedjr: I would combine this check with the integer range check above. Instead of returning `false` from an if statement, you can use the bang (`!`) operator to reverse or negate the truthyness of an expression.

```
def valid_move?(move)
  ...
  move.between(0, 9) && !taken?(move + 1)
end
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+    else
+     false
+   end
+  end  
+
+  def update(p, player)
+   move=p.to_i-1
+   cells[move] = player.token
+   cells[move]
```
randallreedjr: You don't need to put the expression `cells[move]` here again to use it as the return value; the assignment operation in the previous line will already return `cells[move]`.

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+      player_1 = Human.new("X") if !player_1
+      player_2 = Human.new("O") if !player_2
+      
+      @board = board if board
+      @player_1 = player_1 if player_1
+      @player_2 = player_2 if player_2
+  end
+
+  def current_player
+    if (@board.turn_count % 2) == 0
```

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1=nil, player_2=nil, board=nil)
+      board = Board.new if !board
+      player_1 = Human.new("X") if !player_1
+      player_2 = Human.new("O") if !player_2
+      
+      @board = board if board
+      @player_1 = player_1 if player_1
+      @player_2 = player_2 if player_2
```
randallreedjr: Wouldn't board, player_1, and player_2 always be set by this point?

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+    end
+  end
+  
+  def over?
+
+  end
+
+  def won?
+    won=false
+    WIN_COMBINATIONS.each do |arr|
```
randallreedjr: You could avoid the need for the local variable `won` by changing your iterator. Here you could use [`any?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-any-3F) to return true/false based on your criteria.

```
def won?
  WIN_COMBINATIONS.any? do |arr|
    @board.cells[arr[0]] == "X" && @board.cells[arr[1]] == "X" && @board.cells[arr[2]] == "X" ||
    @board.cells[arr[0]] == "O" && @board.cells[arr[1]] == "O" && @board.cells[arr[2]] == "O"
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+# game.turn 
+# game.board.display
+# game.turn 
+# game.board.display
+# game.turn 
+# game.board.display
+# game.turn 
+# game.board.display
+# game.turn 
+# game.board.display
```
randallreedjr: Good practice to clean up these comments before submitting a pull request.

randallreedjr:
@jjlippman Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

PR #94, opened 2016-02-21
Done

mcglaser:


bin/tictactoe
```diff
@@ -1,3 +1,71 @@
...
+          play_again?
+      end
+  end
+
+
+puts " "
+puts "     *****Welcome to Tic Tac Toe!*****"
+puts "     The GREATEST and most ENJOYALBE game of all-time!"
+puts " "
+choose_players
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,71 @@
...
+        puts " "
+        puts "     Invalid entry. Please only enter Y or N."
+          play_again?
+      end
+  end
+
+
+puts " "
+puts "     *****Welcome to Tic Tac Toe!*****"
+puts "     The GREATEST and most ENJOYALBE game of all-time!"
```
randallreedjr: Too true

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    puts "      #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+  end
+
+
+  def position(number)
+    self.cells[number.to_i - 1]
+  end
+
+  def full?
+    return true if self.cells.none? { |position| position == " " }
```
randallreedjr: Avoid returning boolean literals `true` and `false`. Instead, rely on the implicit return value of the boolean expression.
```
def full?
  self.cells.none? { |position| position == " " }
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def turn_count
+  self.cells.select {|position| position != " "}.count
```

```ruby
def turn_count
  self.cells.count {|position| position != " "}
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  def taken?(number)
+    if self.position(number) == "X" || self.position(number) == "O"
+      return true
+      else false
+    end
+  end
+
+  def valid_move?(number)
+    number = number.to_i
+    if number < 1 || number > 9
```
`number.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+
+
+  def initialize(player_1 = Human.new('X'), player_2 = Human.new('O'), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+    puts "     OK, let's play some Tic Tac Toe!"
+    puts "     Player 1 is X and Player 2 is O. Player 1 goes first."
+    puts "     Choose a number 1 - 9 (left to right, top to bottom)"
+    puts " "
```
randallreedjr: This output really belongs in the interface (CLI) class.

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  def turn_count
+    board.count{|space| space == "X" || space == "O"}
+  end
+
+  def current_player
+   if board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+      letter == "X" || letter == "O"
+    end
+  end
+
+  def draw?
+    full? && !won?
+  end
+
+  def over?
+    won? || draw? || full?
```
randallreedjr: Are all three of these terms necessary? Can the board ever be `full?` if no player has `won?` and the game is not a `draw?`?

randallreedjr:
Great job @mcglaser! I left you a few comments, feel free to respond on any you'd like to discuss. Closing this PR.

PR #95, opened 2016-02-22
Done

fiveinfinity:


README.md
```diff
@@ -19,7 +19,7 @@ You'll be implementing Tic Tac Toe using multiple objects that relate and collab
 
 *Instructions for how to work on a Group Project with Learn*
 
-### Some Hints on Working Together 
+### Some Hints on Working Together
```
randallreedjr: Wow, how did you even find this?

bin/tictactoe
```diff
@@ -1,3 +1,26 @@
...
+ if game.over?
+   puts "Would you like to play again?"
+   answer = gets.strip.downcase
+   if answer == "yes"
+     game.board.reset!
+     game.play
+   elsif answer == "no"
+     exit
+   end
+ end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+
+  #resets board
+  def reset!
+    @cells.clear
+    9.times do
+      @cells.push(" ")
+    end
+    @cells
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  def position(position)
+    @cells[position.to_i-1]
+  end
+
+  #returns boolean if board is full or not
+  def full?
+    if @cells.include?(" ")
+      false
+    else
+      true
```
randallreedjr: Avoid returning boolean literals `true` and `false`. Instead, rely on the implicit return value of the boolean expression.
```
def full?
  @cells.include?(" ")
end
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+  end
+
+  #returns boolean if move is a valid move e.g. if the array index is " ", and not an X or O.
+  def valid_move?(position)
+    if position.to_i.between?(1,9)
+    return true if @cells[position.to_i-1] == " "
```
randallreedjr: You can combine both of these validations into a single line.
```
def valid_move?(position)
  if position.to_i.between?(1,9) && @cells[position.to_i-1] == " "
    return true
  else
    return false
  end
end
```
But remembering the previous comment about implicit return values, we can simplify further:
```
def valid_move?(position)
  position.to_i.between?(1,9) && @cells[position.to_i-1] == " "
end
```

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def over?
+      #draw
+      if self.draw?
+        true
+        #won
+      elsif self.won?
+        true
+        #in-progress
+      elsif !self.draw?
```
randallreedjr: This third condition is confusing. Haven't you already checked `self.draw?`?

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+    elsif !@board.full? && !self.won?
+      false
+    end
+  end
+
+  def winner
+    if self.won?
+      self.won?
+    else
+      nil
```
randallreedjr: In ruby, when the condition on an if statement fails, the whole expression returns nil (try it out in the console!). So `else nil` is redundant.
```
def winner
  if self.won?
    self.won?
  end
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+class Player::Human < Player
+
+  #gets player move, inherits from Player. 
+  def move(board)
+    puts "What's your next move?"
+    player_move = gets.chomp
+    player_move
```
```
def move(board)
  puts "What's your next move?"
  gets.chomp
end
```

randallreedjr:
Great job @fiveinfinity! I left you a few comments, feel free to respond. Closing this PR.

PR #96, opened 2016-02-22
Working TTT with AI

Smulligan85:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
+game.start
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(position)
+    position = position.to_i - 1
+    @cells[position]
+  end
+
+  def full?
+    @cells.all? { |position| position == "X" || position == "O" } ? true : false
```

`@cells.all? { |position| position == "X" || position == "O" }`

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def turn_count
+    counter = 0
+    @cells.each do |i|
+      if i == "X" || i == "O"
+        counter += 1
+      end
+    end
+    counter
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+      true
+    end
+  end
+
+  def valid_move?(position)
+    position = position.to_i
+    if taken?(position) || !(position.between?(1, 9))
+      false
+    else
+      true
```
randallreedjr: Also avoid returning boolean literals `true` and `false` from if/else statements. The implicit return value is also sufficient here. You can use the `!` (bang) operator to reverse or negate the boolean value.
```
def valid_move?(position)
  position = position.to_i
  !(taken?(position) || !(position.between?(1,9)))
end
```
But all of those negatives make this method a little confusing. We can simplify it by distributing the negation. This might be a little confusing, but `!(a || b)` <=> `!a && !b` (this property is one of [DeMorgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)). So, we can rewrite the updated method as
```
def valid_move?(position)
  position = position.to_i
  !taken?(position) && position.between?(1,9)
end
```

lib/game.rb
```diff
@@ -0,0 +1,122 @@
...
+def over?
+  (won? || draw? || @board.full?) ? true : false
+end
+
+def winner
+  if !won?
+    nil
+  elsif won?
+    win = won?
+    char = @board.cells[win[0]]
```

```ruby
elsif won = won?
  char = @board.cells[won[0]]
```

lib/game.rb
```diff
@@ -0,0 +1,122 @@
...
+      "X"
+    elsif char == "O"
+      "O"
+    end
+  end
+end
+
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Player::Human < Player
+
+  def move(move)
+    move = gets.chomp.upcase
+    move
```
```
def move(board)
  gets.chomp.upcase
end
```

randallreedjr:
@Smulligan85 Great job! I left you a few comments, feel free to respond. Closing this PR

Smulligan85:
Thanks Randall, these are really helpful. I'll update my code this week :+1:

PR #97, opened 2016-02-22
Project submitted

shinyrachel:


bin/tictactoe
```diff
@@ -1,3 +1,11 @@
...
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
+puts "   |   |   "
+puts "-----------"
+puts "   |   |   "
+puts "-----------"
+puts "   |   |   "
+Controller.new.start
```
randallreedjr: Great bin executable!

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  CORNERS = [0, 2, 6, 8]
+  OUTER_MIDDLES = [1, 3, 5, 7]
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+    number = number.to_i - 1 if number.class == String
+    cells[number]
+  end
+
+  def full?
+    cells.none? { |cell| cell == " "}
+  end
+
+  def turn_count
+    cells.count { |cell| cell != " " }
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Player::Human < Player
+  def move(board)
+    puts "Player #{self.token}, please enter your move position:"
+    input = gets.strip
```
```
def move(board)
  puts "Player #{self.token}, please enter your move position:"
  gets.strip
end
```

randallreedjr:
@shinyrachel Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #98, opened 2016-02-23
Project submitted

TheLucasMoore:


bin/tictactoe
```diff
@@ -1,3 +1,13 @@
...
 require_relative '../config/environment'
+
+# ## TO DO ## #
+# 3. Who will go first and be X? Hmmm?
+# 4. With this input, start a new game (Game.new)
+# 5. Play through the game
+
+# 7. wargames option too?
+
+tictactoe
```
randallreedjr: Great bin executable

lib/board.rb
```diff
@@ -0,0 +1,71 @@
...
+
+	attr_accessor :cells, :token, :player
+
+	def initialize
+		reset!
+	end
+
+	def reset!
+		#How the board looks at the start of a game
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,71 @@
...
+		#keeps track of the number of moves that have been played
+		counter = 0
+		cells.each do |count|
+			if count == "X"
+				counter +=1
+			elsif count == "O"
+				counter +=1
+			end
+		end
+		counter
```

```ruby
def turn_count
  cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,71 @@
...
+
+	def taken?(position)
+		#returns true/false for a specific positon on the board
+		if cells[position.to_i-1] == "X"
+			return true
+		elsif cells[position.to_i-1] == "O"
+			return true
+		elsif cells[position.to_i-1] == " "
+			return false
+		end
```
randallreedjr: Avoid explicitly returning boolean literals, relying instead on the implicit return value of the boolean expression.
```
def taken?(position)
  cells[position.to_i-1] == "X" || cells[position.to_i-1] == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  def won?
+    WIN_COMBINATIONS.each do |combination|
+      next unless @board.get_cells(combination).uniq.length == 1
+      token = @board.cells[combination[0]]
+      if (token == 'X') || (token == 'O')
+        @winner = token
+        return true
+      end
+    end
+    false
```

```ruby
the_even_number = nil
[1,2,3].each do |i|
  if i.even?
    the_even_number = i
  end
end
the_even_number
```

```ruby
  i.even?
end
```

randallreedjr:
@TheLucasMoore Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #99, opened 2016-02-23
Project submitted

marcotsui2003:


bin/tictactoe
```diff
@@ -1,3 +1,53 @@
...
+		else
+			puts "Thanks for playing TicTacToe. Have a nice day!"
+			exit
+		end
+	end
+
+end
+
+
+Cli.new.run_game
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe~
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require_relative '../config/environment'
+
+puts "Welcome to Tic-Tac-Toe."
```
randallreedjr: Is this a duplicate?

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+	attr_accessor :cells
+
+	def initialize
+		@cells = [" "]*9
+	end
+
+	def reset!
+		@cells = [" "]*9
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+	attr_accessor :cells
+
+	def initialize
+		@cells = [" "]*9
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+		input = input.to_i
+		self.cells[input-1]
+	end
+
+	def turn_count
+		self.cells.count {|cell| (cell == "X" || cell == "O") }
+	end
+
+	def full?
+		self.turn_count == 9 ? true : false
```

```
def full?
  self.turn_count == 9
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+	end
+
+	def taken?(input)
+		input = input.to_i
+		self.cells[input-1] == "X" || self.cells[input-1] == "O" ? true:false
```
randallreedjr: I would probably do the `input = input.to_i - 1` transformation all in one step.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      [6,4,2]]
+
+	def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+		@player_1 = player_1
+		@player_2 = player_2
+		@board = board
+	end
+
+	def current_player
+		self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+		o_array = self.board.cells.map.with_index {|cell,i| i if cell == "O"}
+		WIN_COMBINATIONS.each do |combination|
+			return "X" if combination.all? {|num| x_array.include?(num)}
+			return "O" if combination.all? {|num| o_array.include?(num)}
+		end
+		nil
+	end
+
+	def won?
+		if self.winner == "X" || self.winner == "O"
```
randallreedjr: Here again, the if/else structure is superfluous, because of the implicit return value of the boolean condition.

```
def won?
  self.winner == "X" || self.winner == "O"
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,60 @@
...
+      [2,5,8],
+      [0,4,8],
+      [6,4,2]]
+
+	def move (board)
+		
+		
+		valid_move = (1..9).select {|num| board.valid_move?(num)}
+
+			#binding.pry
```
randallreedjr: It's good to get in the habit of removing these comments before submitting a pull request.

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Human < Player
+
+	 def move(board)
+	 	input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

randallreedjr:
@marcotsui2003 Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #100, opened 2016-02-23
Done

jmheyd:


bin/tictactoe
```diff
@@ -1,3 +1,51 @@
...
+  
+    puts "Do you want to play another game? (Y/N)?"
+    play_again = gets.strip.upcase
+    if play_again == "Y"
+      start
+    else
+      puts "Thanks for playing!"
+    end
+  end
+  start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+ class Board
+
+	attr_accessor :cells
+
+	#can reset the state of the cells in the board
+  #sets the cells of the board to a 9 element array of " "
+	def reset!
```
randallreedjr: Typically `initialize` is the first method defined in a class.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+	end
+
+	#returns the amount of turns based on cell value
+  def turn_count  
+		cells.count {|c| c != " "}
+	end
+
+	def taken?(input)
+		#(position(input) == "X" || position(input) == "O") ? true : false
+			cells[input.to_i - 1] == "X" || cells[input.to_i - 1] == "O" ? true : false
```

`cells[input.to_i - 1] == "X" || cells[input.to_i - 1] == "O"`

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+    self.player_2 = player_2
+    self.board = board
+  	#accepts two players and a board
+  	#defaults to two human players, X and O, with an empty board
+  end
+
+  def current_player
+  	#returns the correct player for the third move
+    #so if board has 4 spots filled, it is the 5th turn
+    if board.turn_count % 2 == 0 then
```

```
if board.turn_count.even? 
  current_player = self.player_1
else
  current_player = self.player_2
end
```

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner 
+    if won?
+      winner_array = Game::WIN_COMBINATIONS.select {|win_combination|
+        (board.cells[win_combination[0]] == board.cells[win_combination[1]]) &&
+        (board.cells[win_combination[1]] == board.cells[win_combination[2]]) &&
+        board.taken?(win_combination[0] + 1)}
```
randallreedjr: I would probably go ahead and use do/end instead of {} to set off the block. I think it would improve readability.

lib/players/computer.rb
```diff
@@ -0,0 +1,137 @@
+class Player::Computer < Player
+
+	attr_accessor :board
+
+
+	def move(*board)
+		#asks the computer for input and return it
+		binding.pry
```
randallreedjr: !!!

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+	def move(*board)
+		#asks the user for input and return it
+		puts "What position would you like to take?"
+		input = gets.strip
```
```
def move(board)
  puts "What position would you like to take?"
  gets.strip
end

randallreedjr:
@jmheyd Great job! I left you a few comments, feel free to respond. Closing this PR.

randallreedjr:
Also, congrats on being our 100th pull request! :100: :100: :100: :+1: :+1: :balloon: :tada:

PR #101, opened 2016-02-24
Done

seanliu93:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new(Computer.new("X"), Human.new("O"))
+game.play
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
+class Board
+  def initialize
+    @cells = Array.new(9, " "   )
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def full?
+    # returns if board is full
+    @cells.each {|cell|
+      if cell == " "
+        return false
+      end
+    }
+    true
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def turn_count
+    count = 0
+    @cells.each {|cell|
+      if cell != " "
+        count += 1
+      end
+    }
+    count
```
randallreedjr: Here, we have another opportunity to use a higher-level, more expressive iterator. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
  cells.count {|cell| cell != " "}
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def taken?(pos)
+    if position(pos) == "X" || position(pos) == "O"
+      true
+    else
+      false
+    end
```
randallreedjr: Avoid returning boolean literals, relying instead on the implicit return value of the boolean expression.

```
def taken?(pos)
  position(pos) == "X" || position(pos) == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def valid_move?(input)
+    if input.to_i >= 1 && input.to_i <= 9
```

`input.to_i.between?(1,9)`

With that change, you can also use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
```

lib/game.rb
```diff
@@ -0,0 +1,112 @@
...
+    else
+      false
+    end
+  end
+
+  def full?
+    @board.cells.all? do |spot|
+      spot != " "
+    end
+  end
```
randallreedjr: You shouldn't need to redefine this method...is it possible to use `board.full?`?

randallreedjr:
@seanliu93 Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #102, opened 2016-02-25
Done

jinjo39:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
...
+    players = gets.strip
+  end
+if players == "0" then game = Game.new(player_1 = Player::Computer.new("X"), player_2 = Player::Computer.new("O"), board = Board.new)
+elsif players == "1" then game = Game.new(player_1 = Player::Human.new("X"), player_2 = Player::Computer.new("O"), board = Board.new)
+else 
+
+  game = Game.new
+end
+
+game.play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+        cells.all?{|token| token == "X" || token == "O"}
+    end
+    
+    def turn_count
+        cells.count{|token| token == "X" || token == "O"}
+    end
+    
+    def taken?(number)
+        cell_number = (number.to_i)-1
+        cells[cell_number] == 'X' || cells[cell_number] == 'O' ? true : false
```

`cells[cell_number] ==  'X' || cells[cell_number] == 'O'`

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+        
+    end
+    
+    def valid_move?(number)
+        (1..9).any? {|item| item == number.to_i} && !taken?(number)
```
randallreedjr: You can simplify the integer range by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F). This keeps you from having to iterate over the entire range.
`number.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+     false
+    end
+    
+    def draw?
+        board.full? && !won? 
+    end
+    
+    def winner
+      board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won[0]]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,60 @@
+class Player::Computer < Player
+    
+    def move(board)
+      
+      def opponent_token
```
randallreedjr: I would avoid nesting methods here.

randallreedjr:
@jinjo39 Great job! I left you a few comments, feel free to respond. Closing this PR.

PR #103, opened 2016-02-27
Done

swdavies7:


bin/tictactoe
```diff
@@ -1,3 +1,59 @@
...
+      else
+        draws +=1
+      end
+    end
+    puts "X won #{x_wins} times, O won #{o_wins} times, and the players tied #{draws} times."
+    puts "Thanks for playing! Would you like to play again? Enter yes or no"
+    response = gets.chomp
+  end
+end
+puts "I hope you enjoyed playing Steven's Tic Tac Toe. Play again soon!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    cells[position.to_i-1]
+  end
+
+  def full?
+    cells.include?(" ") ? false : true
```
randallreedjr: Avoid using the ternary operator to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. You can use the bang (`!`) operator to reverse or negate the truthyness of the expression, or sometimes find a more fitting method.

`!cells.include?(" ")` or `cells.none? { |cell| cell == " " }`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    cells.count("X") + cells.count("O")
+  end
+
+  def taken?(position)
+    cells[position.to_i-1] == " " ? false : true
+  end
+
+  def valid_move?(position)
+    if position.match(/[1-9]/)
+    taken?(position) ? false : true
```
randallreedjr: Here I would use the 'and' operator (`&&`) to combine these two logical expressions and utilize their implicit return value.
```
position.match(/[1-9]/) && !taken(position)
```
Also your use of regex here is nice. Another, possibly more semantic, option is [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
```
position.between?(1,9)
```

lib/players/computer.rb
```diff
@@ -0,0 +1,144 @@
+class Computer < Player
+  def move(board)
+    if board.cells == [" "," "," "," "," "," "," "," "," "] || board.cells == [" ","X"," "," "," "," "," "," "," "] || board.cells == [" "," "," ","X"," "," "," "," "," "] || board.cells == [" "," "," "," "," ","X"," "," "," "] || board.cells == [" "," "," "," "," "," "," ","X"," "]
+      "5"
```
randallreedjr: Either no Xes, or one is a non-corner...any easier ways to check for this?

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  def move(board)
+    puts "Where would you like to move? Enter board position 1 - 9"
+    input = gets.chomp
```
```
def move(board)
  puts "Where would you like to move? Enter board position 1 - 9"
  gets.chomp
end
```

randallreedjr:
@swdavies7 Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

PR #104, opened 2016-02-27
Project submitted

bhollan:


bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+      g.exit
+    end    
+  elsif input == "0"
+    g = Game.new(Computer.new("X", true), Computer.new("O", true))
+    g.play
+    g.exit
+  else
+    puts "Try choosing from the options next time."
+    Game.new.exit
+  end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+      g = Game.new(Human.new("X"), Computer.new("O", true))
+      g.play
+      g.exit
+    end    
+  elsif input == "0"
+    g = Game.new(Computer.new("X", true), Computer.new("O", true))
+    g.play
+    g.exit
+  else
+    puts "Try choosing from the options next time."
```
randallreedjr: haha

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(num)
+    "#{cells[num.to_i-1]}"
+  end
+
+  def full?
+    self.cells.all? {|cell| cell == "X" || cell == "O"} ? true : false
```
```
self.cells.all? { |cell| cell == "X" || cell == "O" }
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def turn_count
+    counter = 0
+      self.cells.each do |cell| if cell == "X" || cell == "O"
+        counter += 1
+      end
+    end
+    counter
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,136 @@
...
+                        [6,4,2]]
+  def initialize(player_1=nil, player_2=nil, board=nil)
+    @board = board || Board.new()
+    @player_1 = player_1 || Human.new("X")
+    @player_2 = player_2 || Human.new("O")
+    @winner = nil
+  end
+
+  def current_player
+    return @board.turn_count%2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def over?
+    if won?
+      return true
+    elsif draw?
+      return true
```
randallreedjr: This kinda looks like `won? || draw?` to me...

lib/game.rb
```diff
@@ -0,0 +1,136 @@
...
+    else
+      return false
+    end
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |triplet|
+      if triplet.all?{|s| @board.cells[s]=="O"}
```
randallreedjr: Nice combination of iterators here to keep the code succinct.

lib/game.rb
```diff
@@ -0,0 +1,136 @@
...
+        puts "Computer's move: "
+      end
+    else
+      turn
+    end
+  end
+
+  def play
+    #don't print out a board if the computer plays first vs human
+    if @board.turn_count==0 && current_player.is_a?(Computer)
```
randallreedjr: Good handling of an edge case.

lib/players/computer.rb
```diff
@@ -0,0 +1,76 @@
...
+        #this will return a 1-element array
+        hotspot = triplet.select{|s| board.cells[s]==" "}
+        return hotspot.first + 1
+      end
+    end
+    return normal
+  end
+
+  def impossible
+    return rand(1..9).to_s
```
randallreedjr: Perhaps not so impossible after all...

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Human < Player
+
+  def move(token, board=nil)
+    input = gets.strip
+    return input
```
```
def move(token, board=nil)
  gets.strip
end
```

randallreedjr:
@bhollan Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

bin/tictactoe
bhollan: This was all @the-widget .  I loved it too.

Several times during testing: "right, yes, thank you. >:P"

lib/players/computer.rb
bhollan: We were both planning to do the minimax algorithm, but we also were/are both highly interested in our speed through the curriculum.  It was a strategic team choice.  :)

I don't know about @the-widget , but I plan to do minimax on this later and do a blog post about it.  Maybe even write a 'connect 4' version too.

lib/players/human.rb
bhollan: I'm more used to it now, but this was my early 'ruby jitters'.  I'm more bent towards the python "explicit is better than implicit" way of specifying things.

bin/tictactoe
bhollan: Is that just 'common/best practices' or does that actually affect performance/security type stuff?

lib/board.rb
bhollan: Duh.  Whoops.

lib/game.rb
bhollan: That's my engineering background smudging my new skills.  :/

lib/game.rb
bhollan: This was a copy/paste from the procedural TTT.  That one was a testing nightmare for me and this was how that ended up.  I like the look of `won? || draw?` for sure though.

lib/game.rb
bhollan: This one took me half an hour to find, and about 30 seconds to fix.

lib/players/computer.rb
randallreedjr: @bhollan definitely understandable. I would be very interested in reading that blog post. Minimax still confuses me...

bin/tictactoe
randallreedjr: I don't think there are performance or security impacts, just separation of concerns/encapsulation from an Object Oriented design perspective.

PR #105, opened 2016-02-28
Done

Saturn226:
code passes, need to make AI

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,4 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+TicTacToeCLI.new.call
--- a/lib/board.rb
+++ b/lib/board.rb
@@ -1,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  @cells = []
+  def initialize
+    #@cells = Array.new(9," ")
+    reset!
+  end
+
+  def reset!
+    @cells = Array.new(9," ")
+  end
+
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts '-----------'
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts '-----------'
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(position)
+    @cells[position.to_i - 1]
+   
+  end
+
+  def full?
+    !(@cells.any? {|cell| cell == " "}) 
+    #return false if any cells are empty
+  end
+
+
+  def turn_count
+    cells.count { |cell| cell == "X" || cell == "O" }
+  end
+
+  def taken?(position)
+    self.position(position) == "X" || self.position(position) == "O"
+  end
+
+  def valid_move?(position)
+      !taken?(position) && position.to_i.between?(1,9)
+  end
+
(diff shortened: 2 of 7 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+TicTacToeCLI.new.call
```
randallreedjr: Beeeeyooootiful! Exactly what an executable should look like.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  @cells = []
```
randallreedjr: I'm not sure this does what you think it does...

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(position)
+    @cells[position.to_i - 1]
+   
+  end
+
+  def full?
+    !(@cells.any? {|cell| cell == " "})
```
randallreedjr: Use the most expressive and efficient iterator for a problem. `any?` is a pretty good choice, but I think we can do better. `Board#full?` is asking if all elements are not empty strings. You can use logic that is symmetrical to those semantics with the [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) iterator.

`@cells.all? {|cell| cell != " "}`

Sometimes, having a method that returns `true` for a certain situation is more readable than using `!` with a method that returns `false`.

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+  def won?
+      WIN_COMBINATIONS.detect do |combination|
+       board.taken?(combination[0]+1) && board.cells[combination[0]] == board.cells[combination[1]] && board.cells[combination[2]] == board.cells[combination[1]]
+      end
+  end
+
+
+
+  def winner
+    won? ? board.cells[won?.first] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+
+
+      # expect(Game::WIN_COMBINATIONS).to include_array([0,1,2])
+      # expect(Game::WIN_COMBINATIONS).to include_array([3,4,5])
+      # expect(Game::WIN_COMBINATIONS).to include_array([6,7,8])
+      # expect(Game::WIN_COMBINATIONS).to include_array([0,3,6])
+      # expect(Game::WIN_COMBINATIONS).to include_array([1,4,7])
+      # expect(Game::WIN_COMBINATIONS).to include_array([2,5,8])
+      # expect(Game::WIN_COMBINATIONS).to include_array([0,4,8])
+      # expect(Game::WIN_COMBINATIONS).to include_array([6,4,2])
```
randallreedjr: Good habit to try to delete comments like these before submitting a pull request.

randallreedjr:
@Saturn226 Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now, but you can reopen if you update the AI and want me to take a look at it.

lib/board.rb
Saturn226: you're right, I realize that I create the cells array in initialize. So line 4 is unneccessary.

PR #109, opened 2016-03-02
Completed project

sameera91:
Hi @randallreedjr. Here is my resubmitted project.

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,39 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+play_again = " "
+
+while play_again != "exit"
+
+puts "Hello! Enter the number of players. 0, 1, or 2."
+num_players = gets.strip
+
+if num_players == "0"
+  puts "You will watch the computer play!"
+  player_1 = Computer.new("X")
+  player_2 = Computer.new("O")
+
+elsif num_players == "1"
+  puts "You will play against the computer! Who do you want to go first and be X? Enter Human or Computer."
+  first_player = gets.strip
+  if first_player == "Human"
+    player_1 = Human.new("X")
+    player_2 = Computer.new("O")
+  elsif first_player == "Computer"
+    player_1 = Computer.new("X")
+    player_2 = Human.new("O")
+  end
+
+elsif num_players == "2"
+  puts "Play against another human!"
+  player_1 = Human.new("X")
+  player_2 = Human.new("O")
+end
+
+game = Game.new(player_1, player_2)
+game.play
+puts "Would you like to play again? If yes, hit enter. If no, type 'exit'."
+play_again = gets.strip
+
+end
--- a/lib/board.rb
+++ b/lib/board.rb
@@ -1,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    self.reset!
+  end
+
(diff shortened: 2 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
...
+  player_1 = Human.new("X")
+  player_2 = Human.new("O")
+end
+
+game = Game.new(player_1, player_2)
+game.play
+puts "Would you like to play again? If yes, hit enter. If no, type 'exit'."
+play_again = gets.strip
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def full?
+    self.cells.all? {|cell| cell != " "}
+  end
+
+  def turn_count
+    count = 0
+    self.cells.each {|cell| count += 1 if cell != " "}
+    count
```

```ruby
def turn_count
  self.cells.count {|cell| cell != " "}
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def taken?(position)
+    self.cells[position.to_i - 1] == " " ? false : true
```
randallreedjr: Avoid using the ternary operator to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. Additionally, you can reverse (or negate) the boolean with the `!` (bang) operator.

`self.cells[position.to_i - 1] != " "`

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def valid_move?(position)
+    if position.to_i >= 1 && position.to_i <= 9
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+ ]
+
+  def initialize(player_1=Human.new("X"), player_2=Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+      puts "Invalid move! Try again." if player.instance_of? Human
+      position = player.move(self.board)
+    end
+
+      self.board.update(position, player)
+      self.board.display
+  end
+
+  def play
+    while !over?
```
randallreedjr: Here you could use `until over?` if you want.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Human < Player
+
+  def move(board)
+    puts "Enter a number from 1-9."
+    input = gets.strip
+    input
```
```
def move(board)
  puts "Enter a number from 1-9."
  gets.strip
end
```

randallreedjr:
Great job @sameera91! I left you a few comments, feel free to respond. Closing this PR for now.

PR #110, opened 2016-03-02
created new repo for this ttt project. trying to submit this again

bacitracin:
no bundle folder this time

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,58 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+def play_ttt
+  puts "Let's play Tic-Tac-Toe! How many players?"
+  number_players = gets.chomp.to_s
+  puts "Which user should go first and be 'X'? Player 1 or Player 2? Please enter '1' or '2'"
+  starting_player = gets.chomp.to_s
+  if number_players == "2" && starting_player == "1"
+    player_1=Player::Human.new("X")
+    player_2=Player::Human.new("O")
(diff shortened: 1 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,58 @@
...
+      play_ttt
+    elsif play_again == "N"
+      puts "Okay, see you next time."
+    else
+      puts "Sorry I didn't catch that. Please try again."
+    end
+  end
+end
+
+play_ttt
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
```
randallreedjr: Typically `initialize` is the first method defined in a class.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Please choose a move, 1 - 9."
+    puts "The cell numbers go top to bottom, left to right."
+    move = gets.chomp
```
```
def move(board)
  puts "Please choose a move, 1 - 9."
  puts "The cell numbers go top to bottom, left to right."
  gets.chomp
end
```

randallreedjr:
@bacitracin Great job! I left you a few comments, feel free to respond. Closing this PR for now.

PR #111, opened 2016-03-02
Project submitted

bedwardsphoto:


randallreedjr:
Hi @bedwardsphoto. Are you still working on this project? It looks like this submission is missing the bin/tictactoe file.

bedwardsphoto:
I'm not sure what happened I'll have to look and see. I finished it but I was working between two computers on that project because I wasn't home when I was doing it and I may have messed something up. I'll go back and fix it once I finish the CLI gem. 

Thanks for pointing that out. 

B

Brian Edwards Photography
(917)363-8753
brian@brianedwardsphotography.com

www.brianedwardsphotography.com

bin/tictactoe
```diff
@@ -1,3 +1,56 @@
...
+else
+  puts "Please enter a '2', '1', or '0'."
+  player_types
+end
+play_again
+end
+
+
+
+   def which_player
```
randallreedjr: Try to be consistent with your indentation; it will help keep your code clean and readable.

bin/tictactoe
```diff
@@ -1,3 +1,56 @@
...
+def play_again
+  puts " "
+  puts "Would you like to play again? (Y/N)"
+  input=gets.strip
+  if input=="Y" || input=="y"
+    welcome
+  end
+end
+
+welcome
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
+require 'pry'
+class Board
+attr_accessor :cells
+@cells=[]
```
randallreedjr: Initializing `@cells` outside of an instance method context makes this a 'class instance variable' instead of a typical instance variable, which is probably not what you intended here. Class instance variables are only accessible through class methods. Instead, initialize your variable in the `initialize` or `reset` method.

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+cel_num= input.to_i - 1
+cells[cel_num]
+end
+
+def full?
+  if cells.include?(" ")
+    false
+  else
+    true
+  end
```
randallreedjr: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. Reverse (or negate) the boolean with the `!` (bang) operator.
```
def full?
  !cells.include?(" ")
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+def taken?(input)
+  if position(input)=="X" || position(input)=="O"
+    true
+  else
+    false
+  end
+end
+
+def valid_move?(input)
+  if input.to_i.between?(1,9) && taken?(input)==false
```
randallreedjr: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.
```
!taken?(input)
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+]
+def initialize(player_1= Player::Human.new("X"), player_2= Player::Human.new("O"), board= Board.new)
+@player_1=player_1
+@player_2=player_2
+@board=board
+end
+
+
+def current_player
+ if board.turn_count % 2==0
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+  if board.turn_count==9 && !won?
+    true
+  else
+    false
+  end
+end
+
+def winner
+  if won?
+board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won[0]]
end
```

randallreedjr:
@bedwardsphoto Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

bedwardsphoto:
Hi Randall,

Thanks for the feedback! I will make those changes, but I’m unclear about the process. Should I make these changes and submit a new pull request, or just make the changes?

Best,
Brian

Brian Edwards Photography
brian@brianedwardsphotography.com
(917) 363-8753
Fax: (212) 918-9333

http://www.brianedwardsphotography.com

PR #112, opened 2016-03-02
Project submitted

LuKrebs:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
...
+play_again = ""
+until play_again.downcase == "n" || play_again.downcase == "no"
+  puts "Would you like to play again?"
+  play_again = gets.strip
+  play_again.downcase == "y" || play_again.downcase == "yes"
+  game = Game.new
+  until game.won?
+    game.play
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def turn_count
+    @cells.count{|i| i == "X" || i == "O"}
+  end
+
+  def taken?(input)
+    if @cells[input.to_i-1] == "X" || @cells[input.to_i-1] == "O"
+      true
+    else @cells[input.to_i-1] == "" || @cells[input.to_i-1] == " "
+      false
```
```
def taken?(input)
  @cells[input.to_i-1] == "X" || @cells[input.to_i-1] == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  def valid_move?(input)
+    if taken?(input.to_i) == false && input.to_i.between?(1,9) == true
```
randallreedjr: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression. You can use the bang (`!`) operator to reverse or negate the truthyness of the expression.

`taken?(input.to_i) == false` is equivalent to `!taken?(input.to_i)`

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    ]
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+
+
+  def won?
+     WIN_COMBINATIONS.detect do |combo|
+      @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.cells[combo[0]] != " "
+    end
+  end
+
+  def draw?
+    !won? && @board.cells.all?{|i| i == "X" || i == "O"}
```
randallreedjr: This looks suspiciously familiar...is there another method that you have already defined which you could call here?

lib/players/computer.rb
```diff
@@ -0,0 +1,59 @@
...
+  def move(board)
+    if !board.taken?('5')
+      '5'
+    else
+      best_move(board) + 1
+    end
+  end
+
+  def best_move(board)
+    win(board) || block(board) || corner(board) || random
```
randallreedjr: Very nice method, readable and clear.

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Player::Human < Player
+
+  def move(board)
+    input = gets.strip
```
```
def move(board)
  gets.strip
end
```

randallreedjr:
@LuKrebs Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

PR #113, opened 2016-03-03
Project submitted

jakeekaj:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = Array.new(9," ")
+  end
+
+  def position
+    x = gets.chomp.to_i
```
randallreedjr: I don't think `gets

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} \n-----------\n #{@cells[3]} | #{@cells[4]} | #{@cells[5]} \n-----------\n #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+  #to display before the start of the game, making it user-friendly
+  def display_intro
+    puts " 1 | 2 | 3 \n-----------\n 4 | 5 | 6 \n-----------\n 7 | 8 | 9 "
+  end
+
+  def full?
+    !@cells.any?{|x| x == " " || x == "" }
```
randallreedjr: Use the most expressive and efficient iterator for a problem. any? is a pretty good choice, but I think we can do better. Board#full? is asking if all elements are not empty strings. You can use logic that is symmetrical to those semantics with the #all? iterator.

@cells.all? {|x| x != " " && x != ""}

Sometimes, having a method that returns true for a certain situation is more readable than using ! with a method that returns false.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def turn_count
+    x = @cells.count("X")
+    o = @cells.count("O")
+    x + o
```
randallreedjr: This could be done a bit more succinctly using the code block version of count[`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count):

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,172 @@
...
+  ]
+
+  def initialize(player_1 = Human.new("X"),player_2 = Human.new("O"),board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+  
+  def player_1
+    @player_1
```
randallreedjr: With `attr_accessor :player_1`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def player_1
  @player_1
end
```
and
```
def player_1=(player_1)
  @player_1 = player_1
end
```

lib/game.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  def player_2
+    @player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,172 @@
...
+  def over?
+    self.won?||self.draw?
+  end
+
+  def won?
+    if self.winner == "X" || self.winner == "O" 
+      return true
+    else
+      return false
+    end
```
```
def won?
  self.winner == "X" || self.winner == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  def draw?
+    !self.won? && !@board.cells.any?{|x| x == " " || x == "" }
```
randallreedjr: Part of this method looks suspiciously familiar...could you possibly use a previously defined method?

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  def move
+    x = gets.chomp
+    return x
```
```
def move
  gets.chomp
end
```

randallreedjr:
@jakeekaj Great job! I left a few comments, feel free to respond with any questions or comments. Closing this PR for now.

bin/tictactoe
jakeekaj: Thanks Randall! Great point. I'll make sure to keep this in mind, moving forward.

lib/board.rb
jakeekaj: I understand. I did not think this through. I simply thought of the easiest way that I could implement it. :S

lib/board.rb
jakeekaj: This looks cleaner! Thanks!

lib/game.rb
jakeekaj: I can't believe I missed that one. This one is a rookie mistake. I should have reviewed my code even more.

lib/game.rb
jakeekaj: Thanks Randall. The` % 2 == 0 ` got stuck with me from my beginner labs and exercises. Lol.

lib/game.rb
jakeekaj: This needs to stick with me. I sometimes forget about this.

lib/game.rb
jakeekaj: the `full?` method. What a noob mistake.

jakeekaj:
Thank you so much for this review! This helped me a lot. Now, I'm starting to regret not taking the Premium version of Learn so that I can get code reviews like this more often.

randallreedjr:
@jakeekaj Glad you found the comments helpful!

PR #115, opened 2016-03-04
Done

codeesh:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require_relative '../lib/game.rb'
+Game.new_game
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+class Board 
+  attr_accessor :cells 
+  
+  
+  def initialize
+    @cells = Array.new(9," ")
```
randallreedjr: I would use this implementation in the `reset!` method, then call that method from `initialize`.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  def taken?(position)
+    if @cells[position.to_i-1] == "X" || @cells[position.to_i-1] == "O"
+      true
+  else
+    false
+  end
+  end
+
+  def valid_move?(move)
+    (move.to_i).between?(1, 9) && taken?(move) == false
```
randallreedjr: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression. You can reverse (or negate) the boolean with the `!` (bang) operator.
```
&& !taken?(move)
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+    end
+    x
+  end
+
+  def taken?(position)
+    if @cells[position.to_i-1] == "X" || @cells[position.to_i-1] == "O"
+      true
+  else
+    false
```
```
def taken?(position)
  @cells[position.to_i-1] == "X" || @cells[position.to_i-1] == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def turn_count
+    x=0
+    @cells.each do |piece|
+      if piece == "X" || piece == "O"
+        x+=1
+      end
+    end
+    x
```

```ruby
def turn_count
  @cells.count {|piece| piece == "X" || piece == "O"}
end
```

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+  ]
+    def initialize(player_1=Human.new("X"),player_2=Human.new("O"),board=Board.new)
+	    @player_1 = player_1
+	    @player_2 = player_2
+	    @board = board
+    end
+
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  def over?
+    if @board.full? || won? || draw?
+    	true
+  end
+end
+
+
+ def won?
+  @winner = nil
```
randallreedjr: Good idea to store this result in an instance variable.

lib/game.rb
```diff
@@ -0,0 +1,134 @@
...
+          return true
+        else
+          false
+        end
+  end
+
+  def winner
+    if won?
+      @winner
+    end
```
randallreedjr: Could also be written as `@winner if won?` if you would like a single-line approach.

lib/player.rb
```diff
@@ -0,0 +1,10 @@
+
+class Player
+  attr_reader:token
+  def initialize(token)
+    @token =token
+  end
+  def token
```
randallreedjr: With `attr_reader :token`, this method is not needed. `attr_reader` is a macro that automatically adds both the getter method to your class for the selected attribute. In this case, both
```
def cells
  @token
end
```
has been included in your class through the `attr_reader`.

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+
+class Human < Player
+  attr_accessor :move
+  
+  def move(board)
+   input = gets
+   input
```
```
def move(board)
  gets
end
```

randallreedjr:
@codeesh Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #116, opened 2016-03-06
Done

smithWEBtek:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeCLI.new
```
randallreedjr: Perfect bin file

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def taken?(input)
+    result = position(input)
+    if result == "X" || result == "O"; true
+      else false
+    end
+  end
+
+  def valid_move?(input)
+    (input.to_i > 0 && input.to_i < 10) && (taken?(input.to_i) == false)
```
```
input.between?(1,9)
```

Also, you should avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression. You can reverse (or negate) the boolean with the `!` (bang) operator.
```
&& !taken?(move)
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def turn_count
+    cells.count{|cell| cell == "X" || cell == "O"}
+  end
+
+  def taken?(input)
+    result = position(input)
+    if result == "X" || result == "O"; true
+      else false
```
```
def taken?(input)
  result = position(input)
  result == "X" || result == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    self.cells[(input.to_i)-1]
+  end
+  
+  def full?
+    if cells.detect {|cell| cell == " " }; false
```
randallreedjr: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. You can reverse (or negate) the boolean with the `!` (bang) operator.
```
def full?
  !cells.detect {|cell| cell == " " }
end
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+WIN_COMBINATIONS = [[0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[6,4,2],[0,4,8]]
+
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    if board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+      player_1
+    else
+      player_2
+    end
+  end
+           
+  def over?
+    board.full?
```
randallreedjr: Can a game end before the board fills up?

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  def draw?
+    if won? 
+      return false
+    elsif board.full? == true && won? != true
+      return true    
+    elsif board.full? != true
+      return nil
+    end
```
randallreedjr: I think you could simplify this logic down to:
```
def draw?
  board.full? && !won?
end
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+  end 
+
+  def winner
+    if won?
+      board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won[0]]
end
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def turn
+    choice = current_player.move(board)
+    if board.valid_move?(choice)
+      board.update(choice, current_player)
+      board.display
+    else
+      turn
+      puts ""
+      puts ""
```
randallreedjr: Why the two `puts`?

lib/player.rb
```diff
@@ -0,0 +1,12 @@
+class Player
+  attr_reader :token
+
+  def initialize(token)
+    @token = token
+  end
+
+  def token
+    @token.freeze
```
randallreedjr: If you want to override the `token` method, you don't need the `attr_reader`. It's not doing anything here.

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+class Human < Player
+  attr_accessor :input
+  
+  def move(input)
+    puts "Please enter a valid position between 1-9:"
+    puts ""
+    puts ""
+    input = gets.to_s.strip
+    input
```
```
def move(input)
  puts "Please enter a valid position between 1-9:"
  gets.strip
end
```

notes.rb
```diff
@@ -0,0 +1,327 @@
+
```
randallreedjr: Probably want to remove files like this before submitting a pull request.

randallreedjr:
@smithWEBtek Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #117, opened 2016-03-06
Project submitted

glee38:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+#can chain methods to a new instance of the class
+Cli.new.greeting
```
randallreedjr: Great bin file!

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    @cells[position.to_i - 1] = player.token
+  end
+
+  def full?
+    #@cells.count {|token| token == "X" || token == "O"} == 9 ? true : false
+    #refactored to:
+    !@cells.any? {|position| position == " "}
+    
+    #can also do:
+    #@cells.all?{|i| i == "X" || i == "O"}
```
randallreedjr: Yep. I would argue that `all?` is slightly more semantic, but either approach is really fine.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def taken?(position)
+    position(position) == "X" || position(position) == "O"
+    
+    #the following works also:
+    #(@cells[position.to_i - 1] == "X" || @cells[position.to_i - 1] == "O")
+  end
+
+  def valid_move?(position)
+    position.to_i.between?(1,9) && !taken?(position) ? true : false
```

```
position.to_i.between?(1,9) && !taken?(position)
```

lib/cli.rb
```diff
@@ -0,0 +1,61 @@
...
+   
+      puts "Please choose from the following options:"
+      puts "Type '0' for zero player mode"
+      puts "Type '1' for one player mode"
+      puts "Type '2' for two player mode"
+
+      player_option = gets.strip
+        case player_option
+          when "0"
+            puts "Ah, so you just want me to do all the work, huh?"
```
randallreedjr: Snarky

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+    #WIN_COMBINATIONS.any? do |win|
+      #if (board.cells[win[0]] == "X" || board.cells[win[0]] == "O") && board.cells[win[0]] == board.cells[win[1]] && board.cells[win[1]] == board.cells[win[2]]
+        #true
+      #else
+        #false
+      #end
+    #end
+  #end
+
+#refactored to the following:
```
randallreedjr: Very nice refactor

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+        #true
+        #else
+        #false
+      #end
+      #end
+    #end
+  #end
+
+  def winner
+    @board.cells[won?.first] if won? != nil
```

```ruby
def winner
  @board.cells[won.first] if won = won?
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,144 @@
...
+    "5" #if !board.taken?(5) #&& board.turn_count == 1
+    #had to move if !board.taken?(5) to the if statement bc otherwise, it would go into an infinite loop
+  end
+
+  #def opponent_token
+    #self.token == "X" ? "O" : "X"
+  #end
+
+  def random_move(board)
+    moves = ["1", "2", "3", "4", "5", "6", "7", "8", "9"] #should make this into an array of int when refactoring
```
randallreedjr: Good idea

lib/players/computer.rb
```diff
@@ -0,0 +1,144 @@
...
+    # manipulate #potential_win?(board) to return a true/false boolean
+    # if the boolean returns false, reset @count = 0
+    # the boolean only returns true if @count == 2
+      # @counter will increment up based on whether or not an opponent's token occupies the given space in combo. ex: ["X", "X", " "] #=> @count = 2
+
+  @count = 0 
+
+  @potential_loss_array = [] #should return array of cell values. ex: ["X", "X", " "]
+
+  @combo_actual_position = [] #should return an array of the position combo within WIN_COMBINATIONS that renders the boolean 'true'. ex: [0, 4, 8]
```
randallreedjr: Do these all need to be instance variables? Could local variables work?

randallreedjr:
@glee38 Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #118, opened 2016-03-06
Project submitted

abadfish:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeCLI.new.call
```
randallreedjr: Great bin file!

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def valid_move?(input)
+    !self.taken?(input) && input.to_i.between?(1, 9)
+  end
+
+  def update(input, player)
+    cells[input.to_i-1] = player.token
+  end
+end
```
randallreedjr: Everything here looks great!

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+      combo.all? { |position| board.cells[position] == player_2.token }
+    end
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
def winner
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,93 @@
...
+      take_middle(board)
+    when win(board)
+      win(board)
+    when block(board)
+      block(board)
+    when corner(board)
+      corner(board)
+    else
+      random(board)
+    end
```
randallreedjr: One downside to the `case` statement approach here is that you will always execute one of these methods twice (once for the `when` criteria and once for the return value).
Because of the way you've structured your methods, you could also do something like this:
```
move_index = take_middle(board) || win(board) || block(board) || corner(board) || random(board)
```
If you like this approach, I would say that line is looking a little long. Maybe it makes sense to group some of those methods together into another method?

lib/players/computer.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    "#{move_index+1}"
+
+  end
+
+  def take_middle(board)
+    4 if board.turn_count < 4 && !board.taken?(5)
```
randallreedjr: Switching back and forth between 0-index and 1-index based notation could get confusing. How could you make this consistent?

lib/tic_tac_toe_cli.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    if game.over?
+      puts "Would you like to play again? (Yn)"
+      option = gets.strip
+
+      case option
+      when "y".downcase
+        self.call
+      when "n".downcase
+        puts "So Long and Thanks for All the Fish!"
```
randallreedjr: Don't panic!

randallreedjr:
@abadfish Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #119, opened 2016-03-07
Project submitted

ShiLi007:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
...
+  player_2 = Player::Computer.new("O")
+else
+  player_1 = Player::Human.new("X")
+  player_2 = Player::Human.new("O")
+end
+puts "Player 1 is X, Player 2 is O."
+board = Board.new
+game = Game.new(player_1,player_2,board)
+board.display
+game.play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+attr_accessor:cells
+
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  def full?
+    @cells.all?{|element| element != " "}
+  end
+
+  def turn_count
+    count = 0
+    @cells.each do |element|
+      count += 1 if element != " "
+    end
+    count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+ def taken?(input)
+    position(input) != " "
+  end
+
+
+  def valid_move?(input)
+    return false if input.to_i < 1 || input.to_i > 9
+    taken?(input) == false
```
randallreedjr: You can simplify checking the valid integer range, and make it more semantic, by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F). With that simplification, you can combine your two conditions.
```
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"),board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+  @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def over?
+    if won? || draw?
+        true
+        else 
+        false
+    end
```
```
def over?
  won? || draw?
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+        
+        if (p1 == "X" && p2 == "X" && p3 == "X") || (p1 == "O" && p2 == "O" && p3 == "O")
+            return win_combination # return the win_combination indexes that won.
+        end
+    end
+    return false
+  end
+
+def full?
+    @board.cells.all? do |space|
```
randallreedjr: Doesn't `@board` already have a `full?` method?

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  def draw?
+    if !won? && full?
+        true
+    end
+  end
+
+
+  def winner
+    win = won?  
+    return  @board.cells[win[0]] if win
```
randallreedjr: Good. You could also make the assignment in the `if` statement:
```
def winner
  @board.cells[win[0]] if win = won?
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+require_relative '../player.rb'
+
+class Player::Human < Player
+  def move(board)
+    puts "please make a move from 1-9"
+    input = gets.chomp
```
```
def move(board)
  puts "please make a move from 1-9"
  gets.chomp
end
```

randallreedjr:
@ShiLi007 Great job! I left you a few notes, feel free to respond with any questions or comments. Closing this PR for now.

PR #120, opened 2016-03-08
Done

Chris-Sexton:


bin/tictactoe
```diff
@@ -1,3 +1,59 @@
...
+      else
+        draws +=1
+      end
+    end
+    puts "X won #{x_wins} times, O won #{o_wins} times, and the players tied #{draws} times."
+    puts "Thanks for playing! Would you like to play again? Enter yes or no"
+    response = gets.chomp
+  end
+end
+puts "I hope you enjoyed playing Steven's Tic Tac Toe. Play again soon!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    cells[position.to_i-1]
+  end
+
+  def full?
+    cells.include?(" ") ? false : true
```

```
!cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def turn_count
+    cells.count("X") + cells.count("O")
```
randallreedjr: You can also combine these counts using a block.

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+        xes << index
+      elsif pos == "O"
+        oes << index
+      end
+    end
+  if WIN_COMBINATIONS.any? {|combo| combo.all? {|pos| xes.include?(pos)}}
+    "X"
+  elsif WIN_COMBINATIONS.any? {|combo| combo.all? {|pos| oes.include?(pos)}}
+    "O"
+  end
```
randallreedjr: This seems to reuse a lot of code from `won?`. Perhaps there is a way to share the functionality between the two methods?

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def taken?(position)
+    cells[position.to_i-1] == " " ? false : true
+  end
+
+  def valid_move?(position)
+    if position.match(/[1-9]/)
```
randallreedjr: Since we are checking integer values, you could make it more semantic, by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
`position.between?(1,9)`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    taken?(position) ? false : true
+    else
+      return false
```
randallreedjr: Avoid returning boolean literals from `if` or `else` statements, relying instead on the implicit return value of the boolean expression. By removing that code, you can simplify this method by combining your two conditions.
```
def valid_move?(position)
  position.to_i.between?(1,9) && !taken?(position)
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  def move(board)
+    puts "Where would you like to move? Enter board position 1 - 9"
+    input = gets.chomp
```
```
def move(board)
  puts "Where would you like to move? Enter board position 1 - 9"
  gets.chomp
end

randallreedjr:
@Chris-Sexton Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #121, opened 2016-03-08
Done

mirandalily:


bin/tictactoe
```diff
@@ -1,3 +1,41 @@
...
+    until game.won?
+      game.play
+    end
+    puts "Great game! Want to play again? Yes or No?"
+    input = gets.chomp
+      if input == "no" || input == "No" || input == "NO"
+        puts "Thanks for playing!"
+      end
+    end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,41 @@
...
+      puts "Thanks for playing!"
+    end
+  elsif input == "2"
+    game = Game.new(Human.new("X"), Human.new("O"))
+    until game.won?
+      game.play
+    end
+    puts "Great game! Want to play again? Yes or No?"
+    input = gets.chomp
+      if input == "no" || input == "No" || input == "NO"
```
randallreedjr: These comparisons could potentially be simplified to `input.downcase == 'no'`.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def full?
+    cells.all? do |space|
+      space == "X" || space == "O"
+    end
+  end
+
+  def turn_count
+      cells.count{|space| space == "O" || space == "X"}
```
randallreedjr: Good choice of iterators for both `full?` and `turn_count`.

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+  ]
+
+  def initialize(player_1=Human.new("X"), player_2=Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def turn_count
+    @board.count{|space| space == "O" || space == "X"}
```
randallreedjr: You should be able to use `@board.turn_count` without reimplementing the method.

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

`@board.turn_count.even? ? player_1 : player_2`

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def draw?
+    if !won? && full?
+      true
+    end
+  end
+
+  def winner
+    won? ? board.cells[won?.first] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(board)
+    puts "What move would you like to make (1-9)?"
+    move = gets.chomp
```
```
def move(board)
  puts "What move would you like to make (1-9)?"
  gets.chomp
end

File: lib/board.rb [hunk 8]
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def taken?(location)
+    p = location.to_i-1
+    if cells[p] != " " && cells[p] != ""
+      true
```
```
def taken?(location)
  p = location.to_i - 1
  cells[o] != " " && cells[p] != ""
end
```

randallreedjr:
@mirandalily Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #122, opened 2016-03-09
aprietof and fossbender done

fossbender:
Technically has some AI

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,47 @@
 #!/usr/bin/env ruby
-
+require 'pry'
 require_relative '../config/environment'
+
+game_type = ''
+first_player = ''
+
+# Greet the user with a message
+puts "Greetings from TicTacToe.\n"
+
+
+# prompt the user for game type until user enters valid response
+until game_type.match(/[0|1|2]/)
+  puts "Enter the number for the type of game you'd like:\n0 - Computer plays against itself\n1 - You play against the computer\n2 - You play against another person"
+  game_type = gets.strip # game_type is still a string here
+end
+
+
+# let the user know their response was received
+print "\nYou selected a #{game_type}-player game. "
+
+
+case game_type
+
+  when "0"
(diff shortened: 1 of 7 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
...
-
+require 'pry'
 require_relative '../config/environment'
+
+answer = "y"
+until answer == "n"
+  game = Game.new.game_type
+  game.start
+  answer = game.repeat?
+end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def cells_filled?
+    # returns an array of the tokens that have been played
+    cells.find_all {|cell| cell == 'X' || cell == 'O'}
+  end
+
+  def turn_count
+    # returns the number of turns taken based on filled cells
+    cells_filled?.size
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,177 @@
...
+  # defaults to two human players, X and O, with am empty board
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  # returns the correct player, X, for the third move
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Human < Player # inherits from Player
+
+  def move(board)
+    # asks the user for input and returns it
+    puts "Please enter a move from 1 to 9"
+    input = gets.strip
```
```
def move(board)
  puts "Please enter a move from 1 to 9"
  gets.strip
end

randallreedjr:
@aprietof @fossbender Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #123, opened 2016-03-09
Project submitted

aplee29:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToe_CLI.new.start
```
randallreedjr: Perfect CLI class, well done!

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+  def position(user_input)
+    cells[user_input.to_i - 1]
+  end
+
+  def full?
+    cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    cells.count {|cell| cell == "X" || cell == "O"}
```
randallreedjr: Good choice of iterators for both `full?` and `turn_count`.

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+      (board.cells[win_combo[0]] == "O" && board.cells[win_combo[1]] == "O" && board.cells[win_combo[2]] == "O")
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    won? ? board.cells[won?.first] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

If you prefer the single-line approach, you can use
```
```
which will still return `nil` if the condition fails.

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def taken?(user_input)
+    position(user_input) == "X" || position(user_input) == "O"
+  end
+
+  def valid_move?(user_input)
+    (1..9).to_a.include?(user_input.to_i) && !self.taken?(user_input)
```
randallreedjr: You could also use [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F), which may be more semantic.
````
user_input.to_i.between?(1,9)
```

randallreedjr:
@aplee29 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #124, opened 2016-03-09
Done

esimonian:


bin/tictactoe
```diff
@@ -1,3 +1,36 @@
...
+        game =  Game.new(Computer.new("X"), Human.new("O"))
+      game.play
+    else
+      puts "Invalid input"
+    end
+  when "2"
+    game = Game.new()
+    puts "X goes first"
+    game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,36 @@
...
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+  def position(input)
+     cells[input.to_i-1]
+  end
+  def full?
+    cells.all?{|token| token == "X" || token == "O"}
+  end
+  def turn_count
+    cells.count{|token| token == "X" || token == "O"}
```

lib/game.rb
```diff
@@ -0,0 +1,59 @@
...
+  [0,4,8],
+  [6,4,2]
+ ]
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,59 @@
...
+   end
+ end
+  def draw?
+   !won? && board.full?
+ end
+ def over?
+  won? || draw?
+ end
+  def winner
+     won? ? board.cells[won?.first] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  def move(input = nil)
+   puts "Please enter 1-9:"
+   input = gets.strip
```
```
def move()
  puts "Please enter 1-9:"
  gets.chomp
end
```

randallreedjr:
@esimonian Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #125, opened 2016-03-09
Done

agdavid:


bin/tictactoe
```diff
@@ -1,3 +1,18 @@
...
+puts "     :+:         :+:         :+:     "
+puts "    +:+         +:+         +:+      "
+puts "   +#+         +#+         +#+       "
+puts "  +#+         +#+         +#+        "
+puts " #+#         #+#         #+#         "
+puts "###         ###         ###          "
+puts "\n"
+puts "-------------------------------------"
+
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize()
+    #set attribute :cells 
+    #equal to our 'board' array with 9 empty spaces
+    @cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    #to be either an "X" or "O token"
+    #returns true, otherwise false
+    self.cells.all? {|position| position == "X" || position == "O"}
+  end
+
+  def turn_count
+    #instance method that counts each 'board' array position
+    #adding only those that are either "X" or "O token"
+    #returns sum
+    self.cells.count {|position| position == "X" || position == "O"}
```

lib/game.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end  
+
+  #state of the Game methods
+  def current_player
+    board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

`board.turn_count.even? ? self.player_1 : self.player_2`

lib/game.rb
```diff
@@ -0,0 +1,194 @@
...
+  def draw?
+    !self.won? && board.full?
+  end
+
+  def winner
+    #if #won? method returns a winning combo
+    #make won?[0] the index value
+    #and return the item in board.cells[index value]
+    if won?
+      board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(board=nil)
+    puts "Please select a position between 1-9: "
+    position = gets.chomp
```
```
def move(board=nil)
  puts "Please select a position between 1-9: "
  gets.chomp
end
```

randallreedjr:
@agdavid Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #126, opened 2016-03-10
Refactored

Shmuwol:


bin/tictactoe
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+TicTacToeCLI.new.call
```
randallreedjr: Perfect bin file.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
+    cells.all? {|cell| cell != " "}
+  end
+
+  def turn_count
+    cells.count {|cell| cell != " "}
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+      combo.all? { |position| board.cells[position] == player_2.token }
+    end
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
def winner
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,98 @@
...
+  attr_accessor :token, :combos
+
+  def initialize(token)
+    @token = token
+    @combos = Game::WIN_COMBINATIONS
+    @opponents_token = "O" if @token == "X"
+    @opponents_token = "X" if @token == "O"
+  end
+
+  def move(board)
```
randallreedjr: Pretty smart approach, not bad!

randallreedjr:
@Shmuwol Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

lib/game.rb
Shmuwol: Not sure I understand this

Shmuwol: can you elaborate please?

randallreedjr: As written, your `winner` method will execute `won?` twice - first when the `if` statement is evaluated. and second when selecting the index in `cells` using `won?.first`. The issue with this relates to performance, in that `won?` has to iterate through all of the WIN_COMBINATIONS.

To avoid this, you can capture the return value of `won?` and store it in `won`. If `won?` returns a value, `won = won?` will evaluate to that return value, which is 'truthy'. On the other hand, if `won?` fails to return a value, because no winning combination was detected, it will return `nil`; therefore the statement `won = won?` will also return `nil`, which is 'falsey'.

So, this revised approach is saying, if `won?` detects a winner and returns a value, pass that to board.cells to retrieve the winner's token; otherwise, return `nil`.

Shmuwol: like this? 
```ruby
  def winner
    if won = won?
      board.cells[won.first]
    end
  end
```

randallreedjr: Yep!

Shmuwol: 👍 
any explanation why `board.cells[won.first] if won = won?` doesn't work?

randallreedjr: I would expect it would work. What error are you seeing?

Shmuwol:
```ruby 
Game play plays through an entire game
     Failure/Error: game.play
     NameError:
       undefined local variable or method `won' for #<Game:0x007f9e20a8b810>
       Did you mean?  won
                      won?
     # ./lib/game.rb:43:in `winner'
     # ./lib/game.rb:63:in `play'
     # ./spec/04_game_spec.rb:334:in `block (3 levels) in <top (required)>'
```

PR #127, opened 2016-03-12
Project submitted

cernanb:


bin/tictactoe
```diff
@@ -1,3 +1,36 @@
...
+      player2 = Player::Human.new("O")
+  end
+
+  game = Game.new(player1, player2)
+  game.board.display
+  game.play
+
+  puts "Would you like to play again or exit? [play again][exit]"
+  input = gets.strip
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+#require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
```
randallreedjr: `reset!` will take care of this for you; no need to do the double assignment.

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    reset!
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+  def position(space)
+    self.cells[space.to_i-1]
+  end
+
+  def full?
+    @cells.none? {|cell| cell == " "}
+  end
+
+  def turn_count
+    @cells.select {|cell| cell == "X" || cell == "O" }.size
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

`board.turn_count.even? ? self.player_1 : self.player_2`

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+    combo_found
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
+    won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

If you prefer the single-line approach, you could also use
```
board.cells[won[0]] if won = won?
```
which will still return `nil` if the assignment fails.

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+class Player::Human < Player
+  attr_accessor :token
+
+  def move(board)
+    puts "Please enter a move: "
+    input = gets.strip
+    input
```
```
def move(board)
  puts "Please enter a move: "
  gets.strip
end
```

randallreedjr:
@cernanb Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #129, opened 2016-03-14
Project submitted

tyleresmith:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to TicTicToe!"
+game = Game.new
+game.play
```

lib/board.rb
```diff
@@ -0,0 +1,104 @@
+require 'pry'
+require_relative "player.rb"
+
+class Board
+	attr_accessor :cells
+
+	def initialize
+		@cells = []
```
randallreedjr: #reset! will take care of this, no need to do a double assignment.

lib/board.rb
```diff
@@ -0,0 +1,104 @@
...
+      end
+    end
+    a
+  end
+
+	def reset!
+		@cells = []
+		9.times do 
+			@cells << " "
+		end
```

lib/board.rb
```diff
@@ -0,0 +1,104 @@
...
+
+	#keeps track of turn ccount based on occupied cells
+	def turn_count
+		count = 0
+		@cells.each do |cell|
+			if cell == "X" || cell == "O"
+				count +=1
+			end
+		end
+		count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+  def player_2
+    @player_2
+  end
+
+  def player_1
+    @player_1
+  end
+
+  def current_player
+    if ((@board.turn_count)+2) % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+    @board.full? && won? == nil
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won? != nil
+      @board.cells[won?.first]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "../player.rb"
+
+class Human < Player
+
+  def move(board)
+    pos = gets.chomp
+    pos
```
```
def move(board)
  gets.chomp
end
```

randallreedjr:
@tyleresmith Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #130, opened 2016-03-15
Project submitted

haugenc:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
...
+    when "wargame"
+      wins = (1..100).count {|i| game = Game.new(Player::Computer.new("X"),Player::Computer.new("O")) ; game.play}
+      puts "\nComputer won #{wins} games out of 100!"
+    else
+      puts "You have entered an invalid selection."
+  end
+
+  puts "\nWould you like to play again? (y/n):"
+  continue = gets.chomp.downcase
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+  #///CLASS METHODS///#
+  def initialize
+    @cells = Array.new(9," ")
```

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+    #the game is won
+    WIN_COMBINATIONS.find do |win|
+      win.all? {|i| board.cells[i] == player_1.token} || win.all? {|i| board.cells[i] == player_2.token}
+    end
+  end
+  def draw?
+    @board.full? && !won?
+  end
+  def winner
+    @board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,33 @@
+class Player::Computer < Player
+
+  POS_RANK = {4=>4,0=>3,2=>3,6=>3,8=>3,1=>2,3=>2,5=>2,7=>2} #Rank of board position
```
randallreedjr: Nice approach.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  def full?
+    #is board full.  Returns true or false
+    @cells.all?{|cell| cell == "X" || cell == "O"}
+  end
+  def turn_count
+    @cells.count{|cell| cell =="X" || cell == "O"}
+  end
+  def taken?(input)
+    position(input) != " " ? true : false
```

`position(input) != " "`

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  def turn
+    move = current_player.move(board)
+    board.valid_move?(move) == true ? board.update(move,current_player) : turn
```
randallreedjr: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.
```
board.valid_move?(move) ? board.update(move,current_player) : turn
```

randallreedjr:
@haugenc Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #131, opened 2016-03-15
Done

pacosta613:


bin/tictactoe
```diff
@@ -1,3 +1,65 @@
...
+end
+
+def two_player
+  puts "Friends, get ready to duel"
+  game = Game.new(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+  game.play
+  opening_statments
+end
+
+play_game
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    if won? 
+      @board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Player::Human < Player
+  def move(board)
+  puts "Please choose numbers 1-9"
+  input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def full?
+    @cells.all?{|full| full == "X" || full == "O"}
+  end
+
+  def turn_count
+    @cells.count {|full| full == "X" || full == "O"}
+  end
+
+  def taken?(pos)
+    position(pos) == "X" || position(pos) == "O" ? true:false
```
```
def taken?(pos)
  position(pos) == "X" || position(pos) == "O"
end
```

randallreedjr:
@pacosta613 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #133, opened 2016-03-17
Done

gnfisher:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+CLI.new.call
```
randallreedjr: Perfect bin file!

lib/board.rb
```diff
@@ -0,0 +1,97 @@
...
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    !@cells.include?(" ")
+  end
+
+  def turn_count
+    @cells.select {|c| c != " "}.size
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def current_player
+    turn_count % 2 == 0 ? 'X' : 'O'
```
randallreedjr: Use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

`turn_count.even? ? 'X' : 'O'`

lib/board.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def won?
+    WIN_COMBINATIONS.detect do |combo|
+      arr = combo.map {|c| @cells[c]}.sort
+      !arr.include?(" ") && arr.first == arr.last
+    end
+  end
+
+  def winner
+   won? ? @cells[won?[0]] : nil
```
randallreedjr: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

spec/04_game_spec.rb
```diff
@@ -1,340 +1,322 @@
...
-      expect(Game::WIN_COMBINATIONS).to include_array([0,1,2])
-      expect(Game::WIN_COMBINATIONS).to include_array([3,4,5])
-      expect(Game::WIN_COMBINATIONS).to include_array([6,7,8])
-      expect(Game::WIN_COMBINATIONS).to include_array([0,3,6])
-      expect(Game::WIN_COMBINATIONS).to include_array([1,4,7])
-      expect(Game::WIN_COMBINATIONS).to include_array([2,5,8])
-      expect(Game::WIN_COMBINATIONS).to include_array([0,4,8])
-      expect(Game::WIN_COMBINATIONS).to include_array([6,4,2])
-    end
+
```
randallreedjr: Why change all these tests?

randallreedjr:
@gnfisher Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

spec/04_game_spec.rb
gnfisher: I believe I needed to because after integrating AI some of how the game worked changed; rather than rewrite them I just muted them.

PR #134, opened 2016-03-18
Done

aback1227:


aback1227:
Unfinished AI - still working on it.

bin/tictactoe
```diff
@@ -1,3 +1,40 @@
...
+end
+
+game = Game.new(player_1, player_2)
+game.numbered_board
+game.play
+
+puts "Play again? (type 'y' or 'n')"
+input = gets.strip
+end
+puts "Goodbye!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+class Board
+
+	attr_accessor :cells
+
+	def initialize
+		reset!
```
randallreedjr: Being more consistent about whitespace, especially indentation, would make your code more readable.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+		puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+	end
+	
+	def position(input)
+		self.cells[input.to_i-1]
+	end
+
+	def full?
+		if self.cells.all?{|cell| cell == "X" || cell == "O"}
+			true
```

```
def full?
  self.cells.all?{|cell| cell == "X" || cell == "O"}
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+		else
+			false
+		end
+	end
+
+	def turn_count
+		count = 0
+		self.cells.each do |cell|
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+]
+  
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1 
+    @player_2 = player_2 
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+      @board.taken?(combo[0]+1)
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    won? ? @board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

randallreedjr:
@aback1227 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

lib/board.rb
aback1227: For some reason, the indentations/spacing change when I push to Github. My indentations should be two spaces. Wondering if this is a common thing... or just me?

randallreedjr: I've seen it before. What setup are you using? And what text editor?

aback1227: I'm using Sublime 3 on a Mac.

randallreedjr: Hm. Did you set the preferences to use 2 spaces instead of tabs?

aback1227: Ah okay didn't even realize it was doing that. Thank you!

PR #135, opened 2016-03-18
Done

hudbuz:


bin/tictactoe
```diff
@@ -1,3 +1,11 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+puts "Welome to TicTactoe!"
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+
+
+class Board
+
+  attr_accessor :cells, :token, :move, :player
+
+ 
+
+    def initialize
+      @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+
+
+    def reset!
+      @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    end   
+
+    def position(input)
+      move = input.to_i
+      move -= 1
+      @cells[move]
+    end
+
+    def full?
+      true if !@cells.detect {|x| x == " "}
```

```
cells.all? {|cell| cell != " "}`
```

Finally, avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+
+
+    def turn_count
+      counter = 0
+      @cells.each do |x|
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def taken?(input)
+      position(input) != " " ? true : false
+    end
+
+    def valid_move?(input)
+
+      valid_numbers = ["1","2","3","4","5","6","7","8","9"]
+      if taken?(input) == false && valid_numbers.include?(input)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,177 @@
...
+      @player_2.board = @board
+    end
+
+  end
+
+
+  def current_player
+    if @board.turn_count == 0
+      return @player_1
+    elsif @board.turn_count % 2 != 0
```

lib/game.rb
```diff
@@ -0,0 +1,177 @@
...
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [0,4,8], [2,4,6]]
+  def initialize(player_1 = "X", player_2 = "O", board = [" ", " ", " ", " ", " ", " ", " ", " ", " "])
+    if board == [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+      @board = Board.new
+    else
+      @board = board
+    end
+    if player_1 == "X"
+      @player_1 = Player::Human.new(player_1)
+      @player_1.board = @board
```
randallreedjr: Seems like a `board` should belong to a `game`, not a `player`. Don't both players use the same board?

lib/game.rb
```diff
@@ -0,0 +1,177 @@
...
+  end
+  
+
+  
+
+
+  def won?
+    status = false
+
+    Game::WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
    end
  end
```

randallreedjr:
@hudbuz Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #136, opened 2016-03-19
so far...

jbok4:


bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+      player1 = Player::Human.new("X")
+      player2 = Player::Human.new("O")
+  end
+
+  game = Game.new(player1, player2)
+  game.play
+
+  puts "Would you like to play again or exit? [play][exit]"
+  input = gets.strip
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+
+  def current_player
+    if board.turn_count % 2 == 0
```

randallreedjr:
@jbok4 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #137, opened 2016-03-20
Done

tttra:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.run
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board 
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  def turn_count
+    @cells.count {|cell| cell != " "}
+  end
+
+  def taken?(user_input)
+    @cells[user_input.to_i - 1] != " " 
+  end
+
+  def valid_move?(user_input)
+    user_input.to_i.between?(1,9) && @cells[user_input.to_i - 1] == " "
```
randallreedjr: This second condition seems redundant with `taken?`.

lib/game.rb
```diff
@@ -0,0 +1,157 @@
...
+  [6,4,2]]
+
+  def initialize(player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

randallreedjr:
@tttra Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #140, opened 2016-03-22
Project submitted

LonghornRach:


bin/tictactoe
```diff
@@ -1,3 +1,37 @@
...
+  puts "Would you like to play again? (Y/N)"
+  answer = gets.strip
+  if answer == "Y"
+    play
+  else
+    puts "Goodbye"
+  end
+end
+
+play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+  def initialize (player_1 = Human.new("X"), player_2 = Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+    @winner = winner
+  end
+
+  def current_player
+    turn = self.board.turn_count + 1
+    if turn % 2 == 0
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  
+  def move(input=nil)
+    input = gets.strip
```
```
def move(input=nil)
  gets.strip
end

randallreedjr:
@LonghornRach Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #141, opened 2016-03-23
Project submitted

ldgarber:


bin/tictactoe
```diff
@@ -1,3 +1,38 @@
...
+  end
+
+  Game.new(player_1, player_2).play
+
+  puts "Would you like to play again? Y/N"
+  if gets.chomp.upcase != "Y"
+    go_again = false
+    puts "Thanks for playing!"
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+  attr_accessor :cells
+
+  def initialize()
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    return false if (input.to_i <= 0 || input.to_i > 9)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+  def initialize(player_1=Human.new("X"), player_2=Human.new("O"), board=Board.new) 
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+    @winner = nil
+  end
+
+  def current_player
+    #if board#turn_count_moves is even, it's player_1. odd, it's player_2 
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def over?
+    self.won? || self.draw?
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |combination|
```

```ruby
  def won?
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Human < Player
+  def move(board)
+    move = gets.chomp
+    move
```
```
def move(board)
  gets.chomp
end

File: lib/board.rb [hunk 7]
```diff
@@ -0,0 +1,57 @@
...
+    @cells[input.to_i - 1]
+  end
+
+  def full? 
+    @cells.any? {|cell| cell == " "} ? false : true
```

`cells.none? {|cell| cell == " "}`

randallreedjr:
@ldgarber Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #143, opened 2016-03-24
Done

e-baker:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
...
+  when "2"
+    player_1 = Human.new("X")
+    player_2 = Human.new("O")
+    game = Game.new(player_1, player_2)
+    game.play
+  end
+
+  puts "Would you like to play another game or exit?"
+  input = gets.strip
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def position(num)
+    return self.cells[num.to_i-1]
+  end
+
+  def full?
+    x = self.cells.count("X")
+    o = self.cells.count("O")
+    x + o == 9 ? true : false
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def turn_count
+    return self.cells.count("X") + self.cells.count("O")
```

```ruby
def turn_count
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(board)
+    puts "What position would you like to occupy?"
+    input = gets.chomp
```
```
def move(board)
  gets.chomp
end

File: lib/board.rb [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def taken?(position)
+    self.cells[position.to_i-1] == "X" ? true : self.cells[position.to_i-1] == "O" ? true : false
```

Here, you can just use the `||` (OR) operator.
```
def taken?(position)
  self.cells[position.to_i-1] == "X" || self.cells[position.to_i-1] == "O"
end
```

randallreedjr:
@e-baker Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #144, opened 2016-03-25
Project submitted

wsaxe:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+start
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+require 'pry'
+class Board
+  @cells = []
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+require 'pry'
+class Board
+  @cells = []
```
randallreedjr: Initializing `@cells` outside of an instance method context makes this a 'class instance variable' instead of a typical instance variable, which is probably not what you intended here. Class instance variables are only accessible through class methods. Instead, initialize your variable in the `initialize` or `reset` method.

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+  def cells
+    @cells
+  end
+  def cells=(array)
+    @cells = array
+  end
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+  def position(input)
+    @cells[input.to_i-1]
+  end
+  def full?
+    select_if_empty = @cells.select { |cell| cell == " " }
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    select_if_empty == [] ? (true) : (false)
+  end
+  def turn_count
+    select_if_empty = @cells.select { |cell| cell == " " }
+    9-select_if_empty.length
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    [6,4,2]  # Top Right Diagonal
+  ]
+  attr_accessor :board, :player_1, :player_2
+  def initialize(player1 = Player::Human.new("X"), player2 = Player::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player1
+    @player_2 = player2
+  end
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+  def position(input)
+    @cells[input.to_i-1]
+  end
+  def full?
+    select_if_empty = @cells.select { |cell| cell == " " }
+    select_if_empty == [] ? (true) : (false)
```

`select_if_empty == []`

randallreedjr:
@wsaxe Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #145, opened 2016-03-25
Project submitted

viparthasarathy:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
...
+  Game.new(Computer.new("X"), Computer.new("O")).play
+when "1"
+  puts "Enter 1 to go first, or 2 for the computer to go first."
+  starter = gets.chomp
+  starter == "2" ? Game.new(Computer.new("X"), Human.new("O")).play : Game.new(Human.new("X"), Computer.new("O")).play
+when "2"
+  Game.new(Human.new("X"), Human.new("O")).play
+else
+  "Invalid number of players. Shutting down..."
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def full?
+    self.cells.none? {|c| c == " "}
+  end
+
+  def position(pos)
+    self.cells[pos.to_i-1]
+  end
+
+  def turn_count
+    9 - self.cells.select { |c| c == " "}.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  ]
+
+  def initialize(player_1=Human.new("X"), player_2=Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def over?
+    true if won? || draw?
+  end
+
+  def won?
+    x_cells = self.board.cells.each_index.select{|i| self.board.cells[i] == "X"}
+    o_cells = self.board.cells.each_index.select{|i| self.board.cells[i] == "O"}
+    WIN_COMBINATIONS.each do |combination|
```

```ruby
  def won?
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+    false
+  end
+
+  def draw?
+    true if self.board.full? && !won?
+  end
+
+  def winner
+    self.board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(board)
+    puts "Please select an empty position on the board."
+    choice = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
+  end
+
+  def over?
+    true if won? || draw?
```
```
def over?
  won? || draw?
end
```

randallreedjr:
@viparthasarathy Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

viparthasarathy:
@randallreedjr Thanks for the code review. I'll take a look into refactoring it the upcoming week.

PR #146, opened 2016-03-25
Pass tests.  Logic version of CPU AI in progress

marcelgalang:


marcelgalang:
Started minimax AI but stopped after seeing Avi's updated notes.

PR #147, opened 2016-03-26
Test branch

marcelgalang:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.run
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells, :board
+
+  def initialize
+    # @board= @cells
+    @cells= Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    @cells[input.to_i - 1 ]
+  end
+
+  def full?
+     @cells.include?(" " ||"") ? false : true
+  end
+
+  def turn_count
+    count = 0
+    cells.each do |cell|
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,187 @@
...
+    # def initialize(player_1, player_2, board)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+
+  end
+
+  def current_player
+    counter = 0
+    board.cells.each do |turn|
```
randallreedjr: `#turn_count` would provide you with this number, no?

lib/game.rb
```diff
@@ -0,0 +1,187 @@
...
+      if turn == "X" || turn == "O"
+        counter += 1
+      end
+    end
+      counter % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,187 @@
...
+    end
+  end
+
+  def won?
+
+    WIN_COMBINATIONS.detect do |win|
+
+    if position_taken?(win[0]) && position_taken?(win[1]) && position_taken?(win[2]) && board.cells[win[0]] ==
+      board.cells[win[1]] && board.cells[win[0]] == board.cells[win[2]]
+      [win[0],win[1], win[2]]
```
randallreedjr: `[win[0],win[1],win[2]]` equals `win`, doesn't it?

lib/game.rb
```diff
@@ -0,0 +1,187 @@
...
+
+  def position_taken?(location)
+    !(board.cells[location].nil? || board.cells[location] == " ")
+  end
+
+  def winner
+    if win = won?
+      return board.cells[win[0]]
+    else
+       nil
```
randallreedjr: This `else` clause is unnecessary; an `if` statement will return nil if the condition is not met.

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+require 'pry'
+
+class Human < Player
+
+  def move(input)
+    input = " "
+    input = gets.chomp
+    input
```
```
def move(input)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    puts "\n"
+    puts "\n"
+  end
+
+  def position(input)
+    @cells[input.to_i - 1 ]
+  end
+
+  def full?
+     @cells.include?(" " ||"") ? false : true
```

`!@cells.include?(" " || "")`

randallreedjr:
@marcelgalang Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #148, opened 2016-03-28
Done

bedwardsphoto:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+TicTacToeCLI.new.call
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+   
+    def initialize
+      @cells=[]
```
randallreedjr: `reset!` will take care of setting `@cells`, no need to also set it equal to an empty array.

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    def over?
+      won? || draw?
+    end
+
+    def won?
+      WIN_COMBINATIONS.detect do |combo|
+        if board.cells[combo[0]] == "X" && board.cells[combo[1]]=="X" && board.cells[combo[2]] == "X"
+          return combo
```
randallreedjr: You don't need an explicit return here, just the condition. `detect` will return the element for you.

randallreedjr:
@bedwardsphoto Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #149, opened 2016-03-30
Done

mcclains2003:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+start_game
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def cells
```
randallreedjr: With `attr_accessor :cells`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    @cells
+  end
+
+  def reset!
+    self.cells = Array.new(9, " ")
+  end
+
+  def full?
+    @cells.each do |token|
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+      if token == " "
+        return false
+      end
+    end
+  end
+
+  def turn_count
+    counter = 0
+    @cells.each do |space|
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+      return false
+    else
+      return false
+    end
+  end
+
+  def won?
+    winner = false
+
+    WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    return winner
+  end
+
+  def winner
+    if won? == false
+      return nil
+    else 
+      @board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(position)
+    puts "Where would you like to place your icon?"
+    token_placement = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+    [0,4,8], #top left to bottom right
+    [6,4,2]  #bottom left to top right
+  ]
+
+  def current_player
+    (self.board.turn_count.even? ? self.player_1 : self.player_2)
+  end
+
+  def over?
+    (self.draw? || self.won? ? true : false)
```
randallreedjr: Avoid using the ternary operator to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. 

```
def over?
  self.draw? || self.won?
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+          winner = [combo[0],combo[1],combo[2]]
+      end
+    end
+
+    return winner
+  end
+
+  def winner
+    if won? == false
+      return nil
```
randallreedjr: In this case, you don't need to explicitly return `nil`; an `if` statement will return `nil` when the condition is not met.
```
def winner
  if won?
    @board.cells[won?[0]]
  end
end
```

randallreedjr:
@mcclains2003 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #150, opened 2016-03-31
done

LeahEarl:


bin/tictactoe
```diff
@@ -1,3 +1,11 @@
...
 
 require_relative '../config/environment'
+
+puts "Welcome to Global Thermonuclear War!"
+sleep(3)
+puts "Apologies, we are having technical difficulties. Let's play Tic-Tac-Toe in the meantime."
+sleep(3)
+puts "Welcome to Tic-Tac-Toe!"
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  #-----checks for open spaces/" "-----#
+  def full?
+    !cells.include?(" ")
+  end
+
+  #-----counts cells containing markers-----#
+  def turn_count
+    cells.select {|m| m != " "}.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,122 @@
...
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  #-----use results from won? to return player token-----#
+  def winner
+    board.position(won?[0]+1) if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,45 @@
...
+  def make_combo(board, token)
+    Game::WIN_COMBINATIONS.detect do |combo|
+      (board.cells[combo[0]] == token && board.cells[combo[1]] == token && !board.taken?(combo[2]+1)) ||
+      (board.cells[combo[1]] == token && board.cells[combo[2]] == token && !board.taken?(combo[0]+1)) ||
+      (board.cells[combo[0]] == token && board.cells[combo[2]] == token && !board.taken?(combo[1]+1))
+    end
+  end
+
+  def best_move(board)
+    winning(board) || blocking(board) || corners(board) || sides(board)
```
randallreedjr: 👍 Good method, very readable.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Human < Player
+
+  #-----gets pos for player move-----#
+  def move(board = Board.new)
+    pos = gets.chomp
+    pos
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  #-----checks numbered position for !" "-----#
+  def taken?(num)
+    cells[num.to_i - 1] == "X" || cells[num.to_i - 1] == "O" ? true : false
```

def taken?(num)
  cells[num.to_i - 1] == "X" || cells[num.to_i - 1] == "O"
end
```

randallreedjr:
@LeahEarl Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #152, opened 2016-04-01
Done

RMcNeely:


lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def update(num, player)
+      cells[(num.to_i-1)] = player.token
+  end
+
+  def reset!
+    self.cells.clear
+    self.cells=(Array.new(9, " "))
+  end
+  #binding.pry
```
randallreedjr: It's good practice to remove these comments before submitting your pull requests.

bin/tictactoe
```diff
@@ -1,3 +1,32 @@
...
+    puts "MACHINE MADNESS!"
+    game = Game.new(Player::Computer.new("X"), Player::Computer.new("O"))
+  end
+
+  game.play
+
+  puts "Would you like to play again?"
+  puts "Stop or one more game"
+play_again = gets.strip
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  def taken?(num)
+    cells[(num.to_i - 1)] == " " ? false : true
+  end
+
+  def update(num, player)
+      cells[(num.to_i-1)] = player.token
+  end
+
+  def reset!
+    self.cells.clear
```
randallreedjr: `cells.clear` is not necessary here, as the assignment will occur on the next line.

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  def position(num)
+    cells[num.to_i - 1]
+  end
+
+  def full?
+    turn_count == 9
+  end
+
+  def turn_count
+    cells.count("X") + cells.count("O")
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def valid_move?(num)
+     taken?(num) == false && num.to_i.between?(1, 9)
```
randallreedjr: Instead of comparing to `false`, you can negate the result of the boolean operation with the `!` (bang) operator.
```
!taken?(num)
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count.even? ? player_1 : player_2
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |combination|
```

```ruby
  def won?
      @board.cells[combo[0]].taken?
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+
+class Player::Human < Player
+  def move(board)
+    #binding.pry
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def taken?(num)
+    cells[(num.to_i - 1)] == " " ? false : true
```

```
def taken?(num)
  cells[(num.to_i - 1)] ! = " "
end
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    win_combo = won?
+     if win_combo
+       if @board.cells[win_combo[0]] == " "
+           nil
```
randallreedjr: You do not need to explicitly return `nil` from an `if` statement; `if` will return `nil` if the condition fails.
```
  @board.cells[win_combo[0]]
end
```

randallreedjr:
@RMcNeely Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #153, opened 2016-04-03
Project submitted

asherfsullivan:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
...
+  player_1 = Player::Human.new("X")
+  player_2 = Player::Computer.new("O")
+else players == 2
+  player_1 = Player::Human.new("X")
+  player_2 = Player::Human.new("O")
+end
+game = Game.new(player_1, player_2)
+until game.won?
+  game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def initialize(cells = nil)
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def reset!
+    cells.clear
+    initialize
```
randallreedjr: `initialize` is not typically called directly, but rather invoked via `new` when an object is instantiated.

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+  ]
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,6 @@
+class Player::Human < Player
+  def move(board)
+    puts "Please enter 1-9:"
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

randallreedjr:
@asherfsullivan Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #154, opened 2016-04-04
Done

tddycks:


bin/tictactoe
```diff
@@ -1,3 +1,17 @@
...
+Prompt the user for what kind of game they want to play, 0,1, or 2 player.
+Ask the user for who should go first and be "X".
+Use the input to correctly initialize a Game with the appropriate player types and token values.
+When the game is over, the CLI should prompt the user if they would like to play again and allow them to choose a new configuration for the game as described above. If the user doesn't want to play again, exit the program.
+=end
+
+puts "Greetings! Word on the street has it that you're pretty good at Tic Tac Toe. Let's see what you got!"
+
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+
+class Board
+
+  attr_accessor :cells, :player
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+      board.cells[combo[0]] != " "
+    end
+  end
+
+def draw?
+  board.full? && !won?
+end
+
+def winner
+  board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

randallreedjr:
@tddycks Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #156, opened 2016-04-05
Project submitted

vannak1:


bin/tictactoe
```diff
@@ -1,3 +1,29 @@
...
+  else
+    puts "Starting a game with #{player} human players."
+    game = Game.new(Player::Computer.new("X"), Player::Human.new("O"), Board.new)
+    game.play
+  end
+else
+  puts "Starting a game with 0 human players."
+  game = Game.new(Player::Computer.new("X"), Player::Computer.new("O"), Board.new)
+  game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+class Board
+  attr_accessor :cells
+
+  def initialize(cells = nil)
+    if cells == nil
+      @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    
+    self.cells[position]
+  end
+
+  def full?
+    self.cells.all?{|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    count = self.cells.count{|cell| cell == "X" || cell == "O"}
```
randallreedjr: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def turn_count
  self.cells.count {|cell| cell == "X" || cell == "O"}
end
```

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+      puts "Move ##{self.board.turn_count+1}"
+      self.board.display
+      self.board.update(current_move, player)
+      puts "#{player.token} moved #{current_move}"
+      self.board.display
+    end
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game_status.rb
```diff
@@ -0,0 +1,35 @@
...
+    def draw?
+      !won? && board.full?
+    end
+
+    def winner
+      array_to_check = WIN_COMBINATIONS.collect do |combo|
+        [board.cells[combo[0]],board.cells[combo[1]],board.cells[combo[2]]]
+      end
+
+      if array_to_check.include?(["X", "X", "X"])
```
randallreedjr: It would be good to find a way to reuse this logic from `#won?` instead of reimplementing it.

lib/game_status.rb
```diff
@@ -0,0 +1,35 @@
...
+      array_to_check = WIN_COMBINATIONS.collect do |combo|
+        [board.cells[combo[0]],board.cells[combo[1]],board.cells[combo[2]]]
+      end
+
+      if array_to_check.include?(["X", "X", "X"]) 
+        "X" 
+      elsif array_to_check.include?(["O", "O", "O"])
+        "O"
+      else
+        nil
```
randallreedjr: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
if array_to_check.include?(["X","X","X"])
  "X"
elsif array_to_check.include?(["O", "O", "O"])
  "O"
end
```

randallreedjr:
@vannak1 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #158, opened 2016-04-07
Project submitted

phaltoe:


bin/tictactoe
```diff
@@ -1,3 +1,50 @@
...
+
+  puts "Lesgoooo!!!" if player_choice == "y"
+
+  puts "See you next time!!!\n" if player_choice == "n"
+
+  until player_choice == "y" || player_choice == "n"
+    puts "Invalid Entry, try again!\n"
+    player_choice = gets.strip
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,59 @@
...
+  def draw?
+    @board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    return board.cells[won?.first] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+Human = Class.new(Player)
+
+class Human
+  def move(board)
+    input = gets.strip
```
```
def move(board)
  gets.strip
end
```

randallreedjr:
@phaltoe Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #159, opened 2016-04-07
Done

stlachman:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
...
+  player_1 = Player::Human.new("X")
+  player_2 = Player::Computer.new("O")
+else players == 2
+  player_1 = Player::Human.new("X")
+  player_2 = Player::Human.new("O")
+end
+game = Game.new(player_1, player_2)
+until game.won?
+  game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+class Board
+  attr_accessor :cells, :token
+
+
+
+  def reset!
+   @cells = Array.new(9, " ")
+  end
+
+  def initialize
```
randallreedjr: It is customary to declare `initialize` before any other instance methods in a class.

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+@player_1 = player_1
+@player_2 = player_2
+@board = board
+end
+
+
+def current_player
+  turns = @board.cells.select {|piece| piece == "X" || piece == "O"}
+  if turns.count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+     @board.cells[combination[0]] == @board.cells[combination[1]] && @board.cells[combination[1]]== @board.cells[combination[2]] && @board.cells[combination[0]] != " "
+end
+end
+
+def draw?
+ @board.full? && !won?
+end
+
+def winner
+  @board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+ def display
+  puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+  puts "-----------"
+  puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+  puts "-----------"
+  puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+ end
+
+def position(input)
+cells[input.to_i - 1]
```
randallreedjr: Being consistent with the whitespace in your code, especially indentation, will make your code more readable.

randallreedjr:
@stlachman Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #160, opened 2016-04-08
Done

adriandave13:


bin/tictactoe
```diff
@@ -1,3 +1,34 @@
...
+  when "2"
+    game = Game.new(Player::Human.new(token), Player::Human.new(opponent_token), Board.new)
+  end
+
+  game.play
+
+  puts "Would you like to play again: 'Y' or 'N'?"
+  play = gets.chomp
+
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  attr_accessor :player_1, :player_2, :board
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    (not won?) && full?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
+      position = won?[0]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "-----------"
+    puts " #{@cells[6..8].join(" | ")} "
+  end
+
+  def position(index)
+    @cells[index.to_i - 1]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
```

```
def full?
  !@cells.include?(" ")
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,47 @@
...
+    combination = Game::WIN_COMBINATIONS.detect do |combination|
+      (board.cells[combination[0]] == board.cells[combination[1]] && board.cells[combination[0]] == token && !board.taken?(combination[2] + 1)) ||
+      (board.cells[combination[1]] == board.cells[combination[2]] && board.cells[combination[1]] == token && !board.taken?(combination[0] + 1)) ||
+      (board.cells[combination[0]] == board.cells[combination[2]] && board.cells[combination[0]] == token && !board.taken?(combination[1] + 1))
+    end
+
+    if combination
+      combination.detect {|index| !board.taken?(index + 1)}
+    else
+      nil
```
randallreedjr: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
if combination
  combination.detect {|index| !board.taken?(index + 1)}
end
```

randallreedjr:
@adriandave13 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #161, opened 2016-04-10
Done

devinburnette:


bin/tictactoe
```diff
@@ -1,3 +1,60 @@
...
+      play_again?
+    end
+  else
+    puts "Not a valid player token.. Must select X or O... "
+    exit
+  end
+else
+  puts "Not a valid number of human players."
+  exit
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  attr_accessor :cells, :player
+
+  def initialize()
+    reset!
+  end
+
+  def reset!
+    @cells = []
+    9.times{@cells << " "}
```
randallreedjr: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+  ]
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def update(occupy, player)
+     cells[occupy.to_i - 1] = player.token
+  end
+
+  def full?
+    true unless cells.any?{|x| x == " "}
```
randallreedjr: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. You can reverse (or negate) the boolean with the `!` (bang) operator.

```
def full?
  !cells.any?{|x| x == " "}
end
```

Of course, instead of using `!any?`, you could just use `none?` instead.

randallreedjr:
@devinburnette Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #162, opened 2016-04-11
ttt-ai-mp

miclp:
Working TTT game with AI.

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,7 +1,5 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
-
 
 def tic_tac_toe
 
@@ -10,7 +8,7 @@ def tic_tac_toe
     puts "Would you like to play a 0, 1, or 2 person game?"
     game_type = gets.chomp
 
-    case game_type  #need to account for bad input for game_type
+    case game_type
     when "0"
       player_1 = Computer.new("X")
       player_2 = Computer.new("O")
@@ -27,13 +25,12 @@ def tic_tac_toe
     end
 
     game = Game.new(player_1, player_2)
-    game.play 
+    game.play
 
     puts "would you like to play again?  (y/n)"
     again = gets.chomp
     return nil if again == "n"
   end
-
 end
 
 tic_tac_toe
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,40 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+def tic_tac_toe
+
+  while true
+    puts "Welcome to Tic Tac Toe"
+    puts "Would you like to play a 0, 1, or 2 person game?"
+    game_type = gets.chomp
+
+    case game_type  #need to account for bad input for game_type
+    when "0"
+      player_1 = Computer.new("X")
+      player_2 = Computer.new("O")
+    when "1"
+      player_1 = Human.new("X")
+      player_2 = Computer.new("O")
+    when "2"
+      player_1 = Human.new("X")
+      player_2 = Human.new("O")
+    else
+      puts "bad input, defaulting to a 1 player game"
+      player_1 = Human.new("X")
+      player_2 = Computer.new("O")
+    end
+
+    game = Game.new(player_1, player_2)
+    game.play 
+
+    puts "would you like to play again?  (y/n)"
+    again = gets.chomp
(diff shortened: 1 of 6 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,79 @@
...
+    # game.play
+
+    puts "would you like to play again?  (y/n)"
+    again = gets.chomp
+    return nil if again == "n"
+  end
+end
+
+tic_tac_toe
+puts "bye"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def full?
+    self.cells.all? do |element|
+      element != " " && element != "" && element != nil
+    end
+  end
+
+  def turn_count
+    counter = 0
+    self.cells.each do |element|
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,126 @@
...
+    [2, 5, 8],
+    [0, 4, 8],
+    [6, 4, 2]
+  ]
+
+  # current_player assumes that the first player was "X"
+  # it uses Board's #turn_count to calculate whose turn it is.
+  def current_player
+    num_moves = self.board.turn_count
+    num_moves % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  def over?
+    self.board.full?
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |combination| #iterate over WIN_COMBINATIONS
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+    end
+    false
+  end
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  ####### setter and getter for winning_token
```
randallreedjr: Since you are already using `attr_accessor` for other attributes, and you are using standard getter and setter methods, it would be preferable to add `winning_token` as an `attr_accessor` for consistency.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+
+class Human < Player
+
+  def move(board)
+    puts "Where would you like to move?  Your token is #{self.token}"
+    desired_move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  # Accepts a player-perspective position as a string (1-9).  Casts it .to_i internally
+  # optionally, can be called with perspective = "array perspective"
+  def taken?(position, perspective = "player perspective")
+    if perspective == "array perspective"
+      current = self.cells[position]
+    elsif perspective == "player perspective"
+      current = self.cells[position.to_i - 1]
+    end
+    return true if current != " " && current != ""
```

```
def taken?(position, perspective = "player perspective")
  ...
  current != " " && current != ""
end
```

lib/game.rb
```diff
@@ -0,0 +1,126 @@
...
+    @winning_token
+  end
+  #######
+
+  # **OPTIONAL** #winning_player returns the Player object of the winner as found by #won?
+  def winning_player
+    if self.won?
+      return player_1 if player_1.token == self.winning_token
+      return player_2 if player_2.token == self.winning_token
+      return nil # if no winner
```
randallreedjr: I _thin_ this last `nil` is not necessary. If the condition fails, the `if` statement will automatically return `nil`.

lib/game.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+  end
+
+  # #winner returns the token of the winning player, as found by #won?
+  def winner
+    self.won? ? winning_token : nil
```
randallreedjr: Similarly, here you can use if instead of the ternary.
```
def winner
  winning_token if self.won?
end
```

randallreedjr:
@miclp Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

miclp:
@randallreedjr thank you for all the detailed feedback.  I may have questions for you as I look more closely but for now I think I'm all set!

PR #163, opened 2016-04-11
Done

jasonpaulso:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
...
+      game.play_cli
+    when '3'
+      game = Game.new(Player::Computer.new("X"), Player::Computer.new("O"))
+      game.play_cli
+    when "q"
+      exit
+    else
+      puts 'What was that? (You might have entered an invalid response.)'
+    end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,42 @@
+require "pry"
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def reset!
+    @cells.clear
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,42 @@
...
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells.clear
```
randallreedjr: No need to call clear as the contents of cells will be overwritten by the assignment.

lib/board.rb
```diff
@@ -0,0 +1,42 @@
...
+    @cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+  def turn_count
+    @cells.count{|token| token == "X" || token == "O"}
+  end
+  def taken?(position)
+    !(@cells[position.to_i - 1].nil? || @cells[position.to_i - 1] == " ")
+  end
+  def valid_move?(position)
+    return if (position.to_i - 1 < 0)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+    [6,4,2]
+  ]
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+  def turn_count
+    self.board.cells.count{|token| token == "X" || token == "O"}
```
randallreedjr: Isn't this method already implemented as `self.board.turn_count`?

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+  def current_player
+    self.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+    self.board.cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+  def draw?
+    won?.nil? && full?
+  end
+  def over?
+    draw? || won? 
+  end
+  def winner
+    won?.each {|num| return self.board.cells[num][0]} unless won?.nil?
```
randallreedjr: Why use `each` and then `return` right away? I don't think an iterator is necessary here.
```
def winner
  self.board.cells[0][0] if won?
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Player::Human < Player
+  def move(board=nil)
+    puts "Hello, Player #{token}, what is your move(1-9)?"
+    move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+      position_3 = self.board.cells[win_index_3]
+
+      position_1 == "X" && position_2 == "X" && position_3 == "X" || position_1 == "O" && position_2 == "O" && position_3 =="O"
+    end
+  end
+  def full?
+    self.board.cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+  def draw?
+    won?.nil? && full?
```
randallreedjr: You could also use `!won?`; having two methods with `?` looks a little funny.

randallreedjr:
@jasonpaulso Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

jasonpaulso:
Thanks Randall, this is very helpful!

PR #164, opened 2016-04-13
Done

joshnevius:
Passed all TTT tests. Completed AI with easy sample strategy.

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,33 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Let's play Tic Tac Toe!" 
+puts "Please enter 1 or 2 players, or enter 0 for if you just want to watch."
+count = gets.chomp
+
+ 
+case count
+  
+  when "0"
+    first_player = Player::Computer.new("X") 
+    second_player = Player::Computer.new("O")
+  when "1"
+    puts "Enter 1 - 'X' - to go first or 2 - 'O' - for the computer to go first."
+    input = gets.chomp
+    if input == "1"
+      first_player = Player::Human.new("X")
+      second_player = Player::Computer.new("O") 
+    elsif input == "2"
+      first_player = Player::Computer.new("X")
+      second_player = Player::Human.new("O")
+    end
+  when "2"
+    first_player = Player::Human.new("X")
+    second_player = Player::Human.new("O")
+  else
+    puts "Guess not? Maybe next time!"
+
+end
+
+Game.new(first_player, second_player).play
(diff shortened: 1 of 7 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,33 @@
...
+    end
+  when "2"
+    first_player = Player::Human.new("X")
+    second_player = Player::Human.new("O")
+  else
+    puts "Guess not? Maybe next time!"
+
+end
+
+Game.new(first_player, second_player).play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize(cells = nil)
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize(cells = nil)
+    @cells = Array.new(9, " ")  
+  end
+
+  def reset!
+    initialize.clear
```
randallreedjr: This is a little backwards. `initialize` is typically not called directly, only invoked on object instantiation through `new`.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Please enter 1 through 9."
+    puts "\n"
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def turn_count
+    @cells.count do |count|
+      count == "X" || count == "O"
+    end
+  end
+
+  def taken?(char)
+    position(char) == "X" || position(char) == "O" ? true : false
```

```
def taken?(char)
  position(char) == "X" || position(char) == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    if win_combo = won?
+      @board.cells[win_combo.first]
+    else
+      nil
```
randallreedjr: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
if win_combo = won?
  @board.cells[win_combo.first]
end
```

randallreedjr:
@joshnevius Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

PR #165, opened 2016-04-14
Done

motti770:


config/player.rb
```diff
@@ -0,0 +1,9 @@
+require 'pry' 
+class Player
```
randallreedjr: This class should not be defined twice.

randallreedjr:
@motti770 Are you still working on this? I don't see any executable file in your submission.

PR #166, opened 2016-04-15
Project submitted

nothingisfunny:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
...
+		game = Game.new(player_1 = Player::Human.new("X"), player_2 = Player::Computer.new("O"), board = Board.new)
+		game.play
+	else input == "n"
+		game = Game.new(player_1 = Player::Computer.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+		game.play
+	end
+elsif input == "2"
+	game = Game.new(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+	game.play
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+	attr_accessor :cells, :turns_taken
+	def initialize
+		@cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+		return self.cells[input.to_i - 1]
+	end
+
+	def full?
+		self.cells.include?(" ") ? false : true
+	end
+
+	def turn_count
+		turns = 0
+		self.cells.each do |i|
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+		end
+		return turns
+	end
+
+	def taken?(position)
+		self.cells[position.to_i-1] != " " 
+	end
+
+	def valid_move?(position)
+		!self.taken?(position) && position.to_i-1 >= 0 && position.to_i-1 <=9
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+		end
+	end
+
+	def over?
+		board.full? || self.won? || self.draw?
+	end
+
+	def won?
+			a = []
+			WIN_COMBINATIONS.each do |i|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

bin/tictactoe
```diff
@@ -1,3 +1,25 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+puts "Welcome to Tic Tac Toe!"
+puts "press 0 for a 0 players game"
+puts "press 1 for a 1 players game"
+puts "press 2 for a 2 players game"
+input = gets.strip
+if input == "0"
+	game = Game.new(player_1 = Player::Computer.new("X"), player_2 = Player::Computer.new("O"), board = Board.new)
```
randallreedjr: Being consistent with your whitespace, particularly using 2 spaces for indentation, will make your code more readable.

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+require_relative '../player.rb'
+
+class Player
+	class Human < Player
+		def move(board)
+			puts "Input 1-9:"
+			input = gets.strip
+			return input
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+		"-----------", 
+		" #{cells[6]} | #{cells[7]} | #{cells[8]} "
+	end
+
+	def position(input)
+		return self.cells[input.to_i - 1]
+	end
+
+	def full?
+		self.cells.include?(" ") ? false : true
```
```
def full?
  !self.cells.include?(" ")
end
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+		a = board.cells.select{|item| item == "X"}
+		b = board.cells.select{|item| item == "O"}
+		if self.won?
+			if a.length > b.length
+				return "X"
+		    else
+			    return "O"
+			end
+		else
+			nil
```
randallreedjr: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
if self.won?
  if a.length > b.length
    return "X"
  else
    return "O"
  end
end
```

randallreedjr:
@nothingisfunny Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

nothingisfunny:
@randallreedjr Hi, thanks for taking a look and leaving comments, they are really usefull! Honestly, I didn't expect that someone will check the code, and I still didn't finish improving the AI, so I'll work on code some more taking into account your notes.

PR #167, opened 2016-04-17
Done

johnram528:


bin/tictactoe
```diff
@@ -1,3 +1,34 @@
...
+    player_1 = Player::Human.new("X")
+    player_2 = Player::Human.new("O")
+  when "0"
+    player_1 = Player::Computer.new("X")
+    player_2 = Player::Computer.new("O")
+  else
+    puts "Sorry, try again."
+end
+
+Game.new(player_1, player_2).play
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+class Board
+  attr_accessor :cells, :position
+
+  @cells = []
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def reset!
+    self.cells.clear
```
randallreedjr: No need to call `clear` as the contents of `cells` will be overwritten by the assignment.

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    border = "-----------"
+    puts top_row, border, middle_row, border, bottom_row
+  end
+
+  def position(num)
+    self.cells[num.to_i-1]
+  end
+
+  def turn_count
+    count = self.cells.count {|cell| cell == "X" || cell == "O"}
```
randallreedjr: You don't need to use an intermediate assignment here. You can just return `self.cells.count {|cell| cell == "X" || cell == "O"}`.

lib/players/human.rb
```diff
@@ -0,0 +1,15 @@
+class Player::Human < Player
+
+  def move(board)
+    puts "Where do you want to move next?(1-9)"
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    count
+  end
+
+  def full?
+    if self.turn_count == 9
+      true
```
```
def full?
  self.turn_count == 9
end
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+     WIN_COMBINATIONS.detect do |combinations|
+       if combinations.all? {|position| board.cells[position] == "X"}
+        @winner = "X"
+        return combinations
+       elsif combinations.all? {|position| board.cells[position] == "O"}
+        @winner = "O"
+        return combinations
+       else
+        @winner = nil
+        false
```
randallreedjr: In this case, returning the result of `@winner = nil` will be falsey, so `false` is redundant.

randallreedjr:
@johnram528 Great job! I left you a few notes, feel free to respond with any questions or comments. Sorry it took me so long to provide you with feedback, hopefully it’s still useful to you! Closing this PR for now.

johnram528:
Very helpful, thanks!

PR #271, opened 2016-08-08
Done

BriOD:


bin/tictactoe
```diff
@@ -1,3 +1,14 @@
...
+puts "Welcome to tic tac toe! How many players wish to play?"
+num_players = gets.strip.to_i
+if num_players == 0
+  game = Game.new(Players::Computer.new("X"), Players::Computer.new("O"), Board.new)
+elsif num_players == 1
+  game = Game.new(Players::Human.new("X"), Players::Computer.new("O"), Board.new)
+elsif num_players == 2
+  game = Game.new
+end
+game.play
```
mendelB: Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,100 @@
...
+    [2,4,6], # Right top to left bottom
+  ]
+
+  def initialize(player1 = Players::Human.new("X"), player2=Players::Human.new("O"), board = Board.new)
+    @player_1 = player1
+    @player_2 = player2
+    @board = board
+  end
+
+  def current_player
```
mendelB: You can use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

lib/game.rb
```diff
@@ -0,0 +1,100 @@
...
+    
+    # if !won? || draw?
+    #   binding.pry
+    # elsif over? && won?
+    #   @board.cells[won?[0]]
+    # end
+
+    if won?
+      @winner = @board.cells[won?.first]
+    end
```
mendelB: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    cells.all? {|cell| cell != " "}
+
+  end
+
+  def turn_count
+    cells.count {|cell| cell == "X" || cell == "O"}
+
+  end
+
+  def taken?(user_input)
```
mendelB: In such a case, you can avoid specifying `true` or `false` and rather rely on the expression itself which is a boolean:
`position(user_input) == " "` will itself return `true` or `false` based on the validity of that expression

lib/players/computer.rb
```diff
@@ -0,0 +1,32 @@
...
+        move = "5"
+      # second move, take the top left corner
+      elsif board.turn_count == 1
+        move = "1"
+      # third move, check top right and bottom left corner
+      elsif board.turn_count == 2
+        move = [3, 7].detect{|i| !board.taken?(i)}.to_s
+      # Check for any combinations of 2 matching tokens and an empty square  
+      else
+        Game::WIN_COMBINATIONS.detect do |combo|
```
mendelB: You seem to be calling `board.position(i+1)` quite often, it might be worth encapsulating that to a helper method e.g.
``` ruby
def position(i)
 board.position(i+1)
end
```
Same with `board.taken?(i+1)`

Also, instead of setting `move` to `nil` and then conditionally resetting it, it might be worth just using the explicit return within the appropriate conditionals, so that if you ever reach the point where you're resetting `move` just `return` it. (even though we try to avoid using explicit return if possible this would dry up the method considerably)

lib/players/computer.rb
```diff
@@ -0,0 +1,32 @@
...
+          # Check if you have any chances to win
+          if combo.select{|i| board.position(i+1) == token}.size == 2 && combo.any?{|i| board.position(i+1) == " "}
```
mendelB: Instead of `select` and then querying the size, why not use the count method, similarly to how you used it in the `turn_count` method?

mendelB:
Hey @BriOD pretty awesome stuff!

I left a couple of refactoring notes around the code that I hope will prove helpful.
I'd highly recommend refactoring accordingly, and @ me when you're done and I can give this another once over.

Good luck! And be sure to reach out with any questions!

PR #272, opened 2016-08-09
Done

SnowboardTechie:


lib/game.rb
```diff
@@ -0,0 +1,83 @@
+class Game
+  WIN_COMBINATIONS = [[0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[6,4,2]]
+
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+  end
+
+  def get_players
```
mendelB: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
+    !@cells.include?(" ")
+  end
+
+  def turn_count
```
mendelB: I like this approach to this!

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
```
mendelB: Kinda interesting how you just used `include?` and circumvented the usual iteration method that would be implemented here. Nice!

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def turn_count
+    9 - @cells.count(" ")
+  end
+
+  def taken?(pos)
+    position(pos) == "X" || position(pos) == "O"
+  end
+
+  def valid_move?(pos)
```
mendelB: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+    if (1..9).cover?(pos.to_i)
```
mendelB: You can simplify checking the valid integer range, and make it more semantic, by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  def won?
+    WIN_COMBINATIONS.any? {|combo| @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.cells[combo[0]] != " "}
+  end
+
+  def draw?
+    over? && !won?
+  end
+
+  def winner
```
mendelB: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+    if @board.valid_move?(move)
+      @board.update(move, current_player)
+    else
+      turn
+    end
+  end
+
+  def play
+    turn unless over?
+    if over?
```
mendelB: Wouldn't it be impossible to reach this point in the code *unless* `over?` is indeed `true`?

lib/players/computer.rb
```diff
@@ -0,0 +1,83 @@
...
+            else
+              game.board.cells[move - 1] = " "
+            end
+
+          end
+        end
+
+      
+
+      if @choice == nil
```
mendelB: A slightly more elegant alternative here would be to use the bang operator `if !@choice` or the `#nil?` method `if @choice.nil?`

lib/players/computer.rb
```diff
@@ -0,0 +1,83 @@
...
+            game.board.cells[move - 1] = " "
+          else
+            game.board.cells[move - 1] = " "
+          end
+        end
+      end
+
+      end
+
+      if @choice == nil
```
mendelB: Here you could utilize the `||=` method which would only reassign `@choice` if it were `nil`
`@choice ||= moves.sample`

lib/players/computer.rb
```diff
@@ -0,0 +1,83 @@
...
+        return 0
+      end
+    end
+
+    def find_move(game)
+      moves = game.get_available_moves
+
+      @choice = nil
+
+      moves.each do |move|
```
mendelB: It seems that you repeat the same exact logic for both `"X"` and `"O"` which is not very 'DRY'
It might be a good idea to abstract this to a method e.g.
``` ruby
def make_choice(game, move, token)
  game.board.cells[move - 1]  = token
  winner = (game.winner == token)
  game.board.cells[move - 1] = " "
  winner || nil
end
```

`@choice ||= (make_choice(game, move, "X") || make_choice(game, move, "O")`

(^ This is all just a quick rough idea - you could implement this however you want)

mendelB:
Hey @SnowboardTechie @Traci7822 this is really neat!

I left a few comments with some refactoring ideas. I'd highly recommend refactoring accordingly and @ me when you've finished and I can give this one more quick review.

Good luck! Any questions just let me know!

PR #273, opened 2016-08-09
Project submitted

ghost:


lib/game.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  # Game State methods:
+  def current_player
+    self.board.turn_count % 2 == 0 ? @player_1 : @player_2
```
mendelB: Here, you could use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

mendelB:
Hey @bging4 I left you one small comment but I've gotta say, you really nailed this one!!!
Kudos for some excellent refactoring work!

lib/game.rb
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  def winner
+    if winning_combo = won?
+      board.cells[winning_combo.first]
+    end
+  end
+
+  # Managing the game methods
+  def start
```
mendelB: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

PR #274, opened 2016-08-11
Done

KerryAlsace:


lib/board.rb
```diff
@@ -0,0 +1,87 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    # @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+    @cells = Array.new(9, " ")
+  end
+
+  def position(position_number)
```
Lukeghenco: Is there a way we could abstract this a lay deeper?

maybe with 

def position (position_number)
  @cells[position.to_i - 1]
end

lib/board.rb
```diff
@@ -0,0 +1,87 @@
...
+    puts " #{self.position("4")} | #{self.position("5")} | #{self.position("6")} "
+    puts "-----------"
+    puts " #{self.position("7")} | #{self.position("8")} | #{self.position("9")} "
+  end
+
+  def full?
+    self.cells.all? {|character| character == "X" || character == "O"}
+  end
+
+  def turn_count
```
Lukeghenco: Ruby has an iterator call count that I believe would work well here

@cells.count{ | i | i == "X" || i == "0" }

lib/board.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def taken?(position_number)
+    self.position(position_number) == "X" || self.position(position_number) == "O"
+  end
+
+  def valid_move?(position_number)
+    !self.taken?(position_number) && position_number.to_i.between?(1, 9)
+  end
+
+  def update(position_number, player)
```
Lukeghenco: we could do this with less lines and still maintain readability
```ruby 
self.cells[position_number.to_i - 1] = player.token 
self.display
```

lib/board.rb
```diff
@@ -0,0 +1,87 @@
...
+    counter = 0
+    self.cells.each do |position|
+      if position == "X" || position == "O"
+      counter += 1
+      end
+    end
+    counter
+  end
+
+  def taken?(position_number)
```
Lukeghenco: is position number always equal to the number the user inputs or do you at some point change it to be `-1`? just noticed that sometimes you use position number - 1 and sometimes you don't? This is confusing.

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+      self.board.cells.values_at(win_combo[0]) == self.board.cells.values_at(win_combo[1]) &&
+      self.board.cells.values_at(win_combo[1]) == self.board.cells.values_at(win_combo[2])
+    end
+  end
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
```
Lukeghenco: it isn't necessary to set this to be a variable we could just do this
```ruby
if won?
    self.board.cells[won?.first]
end
```

PR #277, opened 2016-08-21
Done

Juanriv45:


bin/tictactoe
```diff
@@ -1,3 +1,33 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
Lukeghenco: @Juanriv45 for lines 4-33 it would be great to abstract this into a CLI class within our lib folder. keep the bin file simple

something like 

```CLI.new.start_game```

File: lib/board.rb [hunk 2]
```diff
@@ -0,0 +1,59 @@
+class Board
+  attr_accessor :cells
+  attr_reader :corners
```
Lukeghenco: @Juanriv45  is corners needed or are we just adding more complexity than we really need?

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  @cells = []
```
Lukeghenco: @Juanriv45 we are setting @cells to be [] in the initialize method. This isn't going to be needed.

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
Lukeghenco: @Juanriv45 an optional way to do this is 

```ruby
@cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
Lukeghenco: And again 
```ruby 
@cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+    puts "*******************************************"
+  end
```
Lukeghenco: @Juanriv45 a line space between methods makes your code a lot easier to read.

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  def position (cell_number)
+    @cells[cell_number.to_i - 1]
+  end
+  def full?
+   @cells.all? do |cell_number|
+     cell_number == "X" || cell_number == "O"
+    end
+  end
+  def turn_count
```
Lukeghenco: @Juanriv45 for lines 27-35 we could simplify this a lot with the Ruby iterator `count`

example 

```ruby
@cells.count { |token| token == "X" || token == "O" }
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+        counter += 1
+      end
+    end
+    return counter
+  end
+  def taken?(cell_number)
+     spot = cell_number.to_i - 1
+    !(@cells[spot].nil? || @cells[spot] == " ")
+  end
+  def corners_taken?
```
Lukeghenco: @Juanriv45 I don't believe we need this method.

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+    [1,3,7,9].all? do |corner|
+        taken?(corner)
+      end
+    end
+
+  def valid_move?(cell_number)
```
Lukeghenco: @Juanriv45 for lines 47-52 you have a few methods that you could use for this already.

example

```ruby 
cell_number.to_i.between?(1, 9) $$ !taken(cell_number)
```

PR #278, opened 2016-08-22
Project submitted

johnfewell:


johnfewell:
Done.

bin/tictactoe
```diff
@@ -1,3 +1,45 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
+
+def choose_game
```
Lukeghenco: @johnfewell I think abstracting lines 5 - 45 into a CLI class in the lib folder would be the better way to do this. The application could then just call 

```CLI.new.choose_game```

on line 5 and end of file

File: lib/board.rb [hunk 2]
```diff
@@ -0,0 +1,48 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
```
Lukeghenco: @johnfewell you could make line 6 just call your `reset!` method.

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+    cells.clear
```
Lukeghenco: @johnfewell line 10 is not necessary since you are setting @cells to be a new array.

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+   def position(input)
+     @cells[input.to_i-1]
+   end
+
+   def full?
+     @cells.all? { |c| c == "X" || c == "O" }
+   end
+
+   def turn_count
+     cells.count{|token| token == "X" || token == "O"}
```
Lukeghenco: great use of the count method here.

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+   end
+
+   def taken?(input)
+     position(input) != " "
+   end
+
+   def valid_move?(input)
+     if input.to_i.is_a?(Integer) && input.to_i.between?(1, 9)
```
Lukeghenco: the .to_i should take care of is_a?(Integer)

PR #279, opened 2016-08-23
Project submitted

Lani1321:


bin/tictactoe
```diff
@@ -1,3 +1,42 @@
-#!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
```
Lukeghenco: @Lani1321 @klovae a great design pattern would have you abstract line 2 - line 42 into a CLI class in the lib folder and then you would just call a new instance of that class. `CLI.new.new_game`

bin/tictactoe
```diff
@@ -1,3 +1,42 @@
-#!/usr/bin/env ruby
```
Lukeghenco: @Lani1321 @klovae  why are we removing this? this is to make the bin file know that you are using ruby code within the file.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+
+class Board
+  attr_accessor :cells
+
+  def initialize           
+    @cells = Array.new(9, " ")
```
Lukeghenco: @Lani1321 @klovae since our `reset!` method does the same thing. We could just use that here instead.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+ def full?
+    @cells.all?{|token| token == "X" || token == "O"}
+  end
+  
+ def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def turn_count
```
Lukeghenco: @Lani1321 @klovae  for lines 31-36 we could use the ruby method `count`

example 

```ruby
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def won?
+    @winning_combo = WIN_COMBINATIONS.detect do |combo|
+      @board.taken?(combo[0] + 1) && @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]]
+    end
+    @winning_combo.nil? ? false : true
+  end
+
+  def winner
+    won? ? @board.cells[@winning_combo[0]] : nil
```
Lukeghenco: @Lani1321 @klovae  good use of ternary operator.

lib/players/computer.rb
```diff
@@ -0,0 +1,23 @@
+module Players
+  class Computer < Player
+    def move(board)
+      move = nil
+      if !board.taken?(5)
+        move = "5"
+      elsif !board.taken?(1)
+        move = [1, 3, 7, 9].detect{|index| !board.taken?(index)}
+
+      else
```
Lukeghenco: @Lani1321 @klovae  this looks really great :)

bin/tictactoe
klovae: Whoops that was a total accident! We'll add it back in.

Lukeghenco: I assumed :)

PR #281, opened 2016-08-23
Done

shalstein:


bin/tictactoe
```diff
@@ -1,3 +1,56 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+#require "pry"
```
Lukeghenco: 2 Things

1: I would remove the pry from your final PR (makes it looks a lot cleaner)

2. I think abstracting line 5 - 56 into a separate class in the `lib` folder would be a great idea, and then on line 5 you could just call 
```ruby
CLI.new.start
```

bin/tictactoe
```diff
@@ -1,3 +1,56 @@
...
+  player_amount = ""
+  until player_amount == "0" || player_amount == "1" || player_amount == "2"
+    puts "Welcome to Tic-Tac-Toe!"
+    puts "Please select the number of players."
+    puts "0 / 1 / 2 "
+    player_amount = gets.strip
+  end
+  #binding.pry
+  player1 = " "
+  until player1 == "x" || player1 ==  "o"
```
Lukeghenco: you could make this simpler too

```ruby
until player1 != " "
```

bin/tictactoe
```diff
@@ -1,3 +1,56 @@
...
+    puts "Please select a token for the first player."
+    puts "x / o"
+    player1 = gets.chomp
```
Lukeghenco: I would change this to 
```ruby 
player1 = gets.strip.downcase
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
+class Board
+
+  attr_accessor :cells
+
+
+  def initialize
+    @cells = Array.new(9, " ")
```
Lukeghenco: we could just call the reset! method here since it already does this for us.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(int)
+    int = int.to_i
```
Lukeghenco: we can just remove line 25 and change line 26 to be 
```ruby 
cells[int.to_i - 1]
```

I would also look at changing int to be input all over your code. It looks more readable

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    cells[int-1]
+  end
+
+  def full?
+    cells.all? {|cell| cell != " "}
+  end
+
+  def turn_count
```
Lukeghenco: great job using the map iterator, but `count` would work better here 

```ruby
cells.count{|token| token == "X" || token == "O"}
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    count = 0
+    cells.map { |cell|  count += 1 if cell != " "  }
+    count
+  end
+
+  def taken?(str)
```
Lukeghenco: line 39 arguments should be the input passed to it. 
I would use this for the entire method 
```ruby
def taken?(input)
    position(input) == " "
end
```

the position method is already converting the input to an integer for us.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    int = str.to_i-1
+    cells[int] != " " ? true : false
+  end
+
+  def valid_move?(str)
+    #int = Integer(str) rescue false
+    int = str.to_i
+    int.between?(1,9) ? !taken?(int) : false
```
Lukeghenco: again change str to input and use something like this
```
input.to_i.between?(1, 9) && !taken?(input)
```
this will return an implicit true or false from Ruby

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+   [3,4,5],
+   [6,7,8],
+   [0,3,6],
+   [1,4,7],
+   [2,5,8],
+   [0,4,8],
+   [2,4,6]
+ ]
+   def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)#binding.pry
+     @board = []
```
Lukeghenco: not sure why we are using this, but then declaring it again on 24. This is not needed.

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+
+   def won?
+     WIN_COMBINATIONS.any? do |num_array|
+       board.cells[num_array[0]] != " " &&
+       board.cells[num_array[0]] == board.cells[num_array[1]] && board.cells[num_array[1]] == board.cells[num_array[2]]
+     end
+   end
+
+   def draw?
+     if !won? && board.full?
```
Lukeghenco: lines 45 - 48 are not needed if we just change line 44 to be 
```ruby
!won? && board.full?
``` 
this will return an implicit true or false

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+      player = WIN_COMBINATIONS.detect do |num_array|
+        board.cells[num_array[0]] != " " &&
+        board.cells[num_array[0]] ==  board.cells[num_array[1]] && board.cells[num_array[1]] == board.cells[num_array[2]]
+      end
+      player == nil ? nil : board.cells[player.first]
+    end
+
+    def turn
+      #binding.pry
+      mv = "yata"
```
Lukeghenco: I would just set this to be an empty string or nil

lib/board.rb
peacestone: I think you meant

position(input) != " "

bin/tictactoe
peacestone: If someone will input any other character on the keyboard the program breaks (I tried it to be sure). That's why I specified untill == o  || == x

Lukeghenco: ok cool

lib/board.rb
Lukeghenco: Yes I did :) great catch :)

PR #282, opened 2016-08-24
Done

em379:


bin/tictactoe
```diff
@@ -1,3 +1,27 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
Lukeghenco: I would abstract lines 5 - 27 into a CLI class with methods in the `lib` folder and on line 5 just call 
```ruby 
CLI.new.start_game
```

bin/tictactoe
```diff
@@ -1,3 +1,27 @@
...
+puts "Welcome to Tic Tac Toe!"
+
+puts "What kind of Game would you like to play? O, 1, or 2 player."
+
+input = gets.strip
```
Lukeghenco: I would make the input an Integer so that you do not have to worry about Uppercase or Lowercase inputs

```ruby 
input = gets.strip.to_i
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
```
Lukeghenco: great job using another method that does what you need :)

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+  def full?
+    cells.all? { |cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    cells.count { |cell| cell == "X" || cell == "O"}
+  end
+
+  def taken?(user_input)
+    if position(user_input) == "X" || position(user_input) == "O"
```
Lukeghenco: we could simplify this to be 
```ruby 
position(user_input) != " "
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
+class Game
+
+  attr_accessor :board, :player_1, :player_2
+
+  WIN_COMBINATIONS = [
+  top_row_win = [0, 1, 2],
```
Lukeghenco: there is no real need to have an array of variables names that contain arrays.

PR #283, opened 2016-08-25
Done

jmb521:


bin/tictactoe
```diff
@@ -1,3 +1,53 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+# 0 player is computer vs computer
```
Lukeghenco: For lines 4-53 we should separate this into a CLI class that lives in our lib folder. We would then be able to call this Class instance method on line 4 instead 

```
CLI.new.start_game
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
Lukeghenco: why are we setting @cells there and then call reset! that does that again for us?

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    reset!
+  end
+  def cells
+    @cells
+  end
+
+  def reset!
+
+     @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
Lukeghenco: We could use some Ruby magic here and use `Array.new(9, " ")` instead of explicitly setting that array.

lib/board.rb
```diff
@@ -0,0 +1,81 @@
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+    reset!
+  end
+  def cells
```
Lukeghenco: do we need a getter method for cells when we have attr_accessor already?

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(pos)
+    current_pos = pos.to_i
+
+    cells[current_pos-1]
```
Lukeghenco: lets simplify this into one line. 

pos is a string but we can `pos.to_i - 1` on line 33 instead of creating a new variable that stores pos as an integer

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  end
+
+  def full?
+    @cells.each do |is_full|
```
Lukeghenco: we have a Ruby Iterator `.all?` that would work well for this. 

```ruby 
@cells.all? { |token| token == "X" || token == "O" }
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+      if is_full.include?(" ")
+        return false
+      end
+    end
+  end
+
+  def turn_count
+    # count = 0
```
Lukeghenco: why don't we just count full space instead that would simplify this code and implicitly return the number for us 

```ruby 
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+    counted = @cells.count{|x| x == " "}
+
+    full_count = 9-counted
+
+  end
+
+  def taken?(x)
```
Lukeghenco: we can condense lines 54-58 into a one liner 
```ruby 
position(x) == "X" || position(x) == "O"
```
this will implicitly return true or false in ruby, so we don't need to make this use an if/else statement

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+      false
+    end
+  end
+
+  def valid_move?(is_valid)
+    validated = is_valid.to_i
+
+    if validated.between?(1,9)
+
+      if taken?(validated)
```
Lukeghenco: we can also just make a one line here that checks with an `&&`

```ruby 
is_valid.to_i.between?(1,9) && !taken?(is_valid.to_i)
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def taken?(x)
+    if position(x) == "X" || position(x) == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(is_valid)
```
Lukeghenco: think about changing the argument name to input, since that is what is being passed in.

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+      else
+        true
+      end
+    end
+  end
+
+  def update(pos, x_or_o)
+    pos = pos.to_i
+    pos -= 1
+    @cells[pos] = x_or_o.token
```
Lukeghenco:
```ruby
@cells[pos.to_i - 1]
```

lib/board.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      if taken?(validated)
+        false
+      else
+        true
+      end
+    end
+  end
+
+  def update(pos, x_or_o)
```
Lukeghenco: so our second argument here should be player

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def board
```
Lukeghenco: do we not have access to this from `attr_accessor :cells` ?

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+    @board
+
+  end
+
+  def player_1
```
Lukeghenco: We have lines 28-33 given to us from `attr_accessor`

lib/game.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def draw?
+    @board.full? && !won?
+
+  end
+
+  def winner
+      if winning = won?
+      @winner = @board.cells[winning.first]
```
Lukeghenco: we could just change this to be 
```
@winner = @board.cells[won?.first]
```

PR #284, opened 2016-08-27
Done

bwatson78:


bin/tictactoe
```diff
@@ -1,3 +1,43 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
Lukeghenco: For lines 5-43. I would abstract this into a CLI class in the lib folder and give them methods you can call. 

On line 5 we could just do this now

```ruby 
CLI.new.start_game
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    reset!
```
Lukeghenco: good job reusing methods 👍

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+    puts ""
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    full_board = @cells.all? do |space|
```
Lukeghenco: we could turn this into a small one liner, since Ruby implicitly returns the last line of code in a method.
```
@cells.all? { |space| space == "X" || space == "O" }
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+      space == "X" || space == "O"
+    end
+    full_board
+  end
+
+  def turn_count
+      @cells.count{|token| token == "X" || token == "O"}
```
Lukeghenco: Good use of the #count method here

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def taken?(index)
```
Lukeghenco: we could use our #position to good use here

example 
```ruby 
!(position(index) == " " || position(index) == "")

File: lib/board.rb [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+    number = index.to_i - 1
+    !(@cells[number].nil? || @cells[number] == " " || @cells[number] == "")
+  end
+
+  def valid_move?(move)
```
Lukeghenco: we could use our taken? method here that is now using #position, and we no longer need to set move to be `.to_i` as a passed in argument for taken.

I would look at using the `between?` Ruby Method for this `|| location < 1 || location > 9`

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    if taken?(location) || location < 1 || location > 9
+      false
+    else
+      true
+    end
+  end
+
+  def update(cell, player)
+    position = cell.to_i - 1
+    @cells[position] = player.token
```
Lukeghenco: just shorten this down to one line
`@cells[cell.to_i - 1] = player.token`

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
+
+  def over?
+    if won? || draw? || @board.full?
```
Lukeghenco: draw? should already be checking for a full board. Also conditionals will already implicitly return true or false, so we don't need to set up an else statement here

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+      position_3 = @board.cells[win_index_3]
+      a_win = [position_1, position_2, position_3].uniq.size == 1 && position_1 != " "
+      if a_win
+        the_answer = win_combination
+      end
+    end
+    the_answer
+  end
+
+  def draw?
```
Lukeghenco: ruby already implicitly return true or false, explicitly returning true or false is not needs, as well as the else statement

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+    if !won? && @board.full?
+      return true
+    else
+      return false
+    end
+  end
+
+  def winner
+    winning_arr = won?
```
Lukeghenco: I would simplify this.

```
won? ? @board.cells[won?[0] : nil 
```

PR #285, opened 2016-09-01
Project submitted

KentLeong:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
Lukeghenco: For lines 5-21. Lets add a new cli.rb file in the lib directory and turn this into a class with callable instance methods.

Then we can use something like this on line 5

CLI.new.start_game

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+require "pry"
+class Board
+  attr_accessor :cells
+
+  def initialize()
+    reset!
```
Lukeghenco: Good use of an existing method here

PR #288, opened 2016-09-07
Project submitted

zszalas2:


lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    @cells[user_input.to_i - 1]
+  end
+
+  def turn_count
+    cells.count{|token| token == "X" || token == "O"}
+  end
+
+# revisit
+  def full?
+    if turn_count == 9
```
Lukeghenco: we could simplify this to be 
```ruby
def full?
  turn_count == 9
end
```

since this will return a true or false.

lib/cli.rb
```diff
@@ -0,0 +1,45 @@
...
+      when "2"
+          game = Game.new(Players::Human.new(player_1), Players::Human.new(player_2)).play
+      
+    end
+
+    def end_game
+      puts "Game is over. Would you like to play again? Y/N"
+       input = gets.chomp
+       case input
+         when "Y"
```
Lukeghenco: You should use some ruby magic to make sure that the input is always capitalized so if a user types y/n it still works.

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+    ]
+
+    def initialize(player_1= Players::Human.new, player_2= Players::Human.new("O"), board = Board.new)
+      @board = board
+      @player_1 = player_1
+      @player_2 = player_2
+    end
+
+    def current_player
+      if @board.turn_count.even?
```
Lukeghenco: I would use a `?` ternary operator here and make this one line of code

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+           player_1
+        else
+           player_2
+      end
+    end
+
+    def over?
+      if @board.full? || won? || draw?
```
Lukeghenco: this should work fine 
```ruby 
def over?
    won? || draw?
end
```

because our draw method is already checking for board to be full

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+          true
+        else false
+      end
+    end
+
+    def draw?
+      if @board.full? && !won?
```
Lukeghenco: we don't need to use these if statements for boolean returns. just the basic comparison left to right will return a true or false implicitly

PR #289, opened 2016-09-10
Project submitted

jminterwebs:


lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    @cells = Array.new(9, " ")
```
Lukeghenco: you should just call a method that already does this for you.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    @cells[self.index(location)]
+  end
+
+  def full?
+    !(@cells.include?(" " || nil))
+  end
+
+  def turn_count
+      count = 9
```
Lukeghenco: for lines 34-40 we could use a simple Ruby iterator called count 

like 

```cells.count{ |token| token == "X" || token == "0" }```

File: lib/game.rb [hunk 3]
```diff
@@ -0,0 +1,190 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+  end
+
+  def current_player
+    if self.board.turn_count % 2 == 0
```
Lukeghenco: we could just use a ternary operator here (?) and have it return the value of self.player_1 or self.player_2

lib/game.rb
```diff
@@ -0,0 +1,190 @@
...
+    [false, nil]
+  end
+
+  def won?
+    # WIN_COMBINATIONS.each do |win|
+    #   ref = self.board.cells[win[0]]
+    #   return true if (ref == "X" || ref == "O") && self.board.cells[win[1]] == ref && self.board.cells[win[2]] == ref
+    # end
+    # false
+    self.win_array[0]
```
Lukeghenco: just move win_array method contents inside of here

lib/game.rb
```diff
@@ -0,0 +1,190 @@
...
+    # return false if self.won?
+    # self.board.cells.each do |position|
+    #     return false if position == " "
+    # end
+    # true
+    return true if !self.won? && self.board.full?
+  end
+
+  def over?
+    return true if self.won? || self.draw?
```
Lukeghenco: no need for an explicit return here. 

couldn't we just do this 
```
def over? 
    self.won? || self.draw?
end
```

lib/game.rb
```diff
@@ -0,0 +1,190 @@
...
+      #     self.board.update(move, player_1)
+      #     valid = true
+      #   else
+      #     puts "Invalid move. Instructions for playing Tic-Tac-Toe can be found here: https://en.wikipedia.org/wiki/Tic-tac-toe"
+      #   end
+      # end
+    elsif current_player == self.player_2
+      # self.cli.player_turn_prompt(current_player, 1)
+      self.player_turn_prompt(current_player, 1)
+      # valid = false
```
Lukeghenco: I would really move out all of this old code. It makes it hard to read what is going on.

lib/game.rb
lair001: Actually, this was a deliberate refactoring. #won? and #winner were identical except for the return value so I implemented #win_array to keep things DRY.

lib/board.rb
jminterwebs: fixed

lib/board.rb
jminterwebs: fixed

lib/game.rb
jminterwebs: fixed

lib/game.rb
jminterwebs: cleaned up comments

bin/tictactoe
```diff
@@ -1,3 +1,16 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require 'pry'
+
+puts "Greetings!"
+
+cli = Cli.new
```
Lukeghenco: Great job on abstracting this class, but all of the code should be located in the cli. All that should be on this page is `Cli.new.setup_game`

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    location.to_i - 1
+  end
+
+  def position(location)
+
+    @cells[self.index(location)]
+  end
+
+  def full?
+    !(@cells.include?(" " || nil))
```

lib/cli.rb
```diff
@@ -0,0 +1,192 @@
...
+    #     puts "In a single player game, Player 1 will be human while Player 2 will be computer."
+    #     valid = true
+    #   else
+    #     puts "Invalid selection!"
+    #     puts "Please enter 0, 1 or 2."
+    #     num_players = gets.chomp.to_i
+    #   end
+    # end
+    go_first = ->(first_player) do
+      # if num_players == 0 && first_player == 1
```
Lukeghenco: I would remove all of this commented out code if it is unnecessary

snsavage:
Hi Luke,

The code that you reviewed seems to be for Samuel's project.  I think that something got messed up when we were choosing groups at the beginning of the project.  David and I completed our project together and Samuel worked with a different partner.  Is there a way that this can be fixed on GitHub and/or the Learn website?  

This is the correct repo for David and I:
https://github.com/snsavage/ttt-with-ai-project-v-000 <https://github.com/snsavage/ttt-with-ai-project-v-000>

Thanks.

Scott   

Scott Savage
snsavage@gmail.com <mailto:snsavage@gmail.com>
snsavage.com <https://www.snsavage.com/>

lair001:
Hi Luke, these comments are not on the pull request from my branch of the
project.  However, some of them are still relevant and I'll make another
pull request later.

PR #290, opened 2016-09-12
Project submitted

kkterai:


lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def reset!
+    self.cells = [" "," "," "," "," "," "," "," "," "]
+  end
+
+  def initialize
+    reset!
+  end
+
+  def display
```
Lukeghenco: I believe this would be much cleaner for you 
```

File: lib/game.rb [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+      Game.new(Players::Computer.new("X"), Players::Computer.new("O"), Board.new)
+    else
+      puts "I didn't understand that."
+      start
+    end
+  end
+
+  def self.start_1_player
+    puts "Who should go first and be 'X'? (Human/Computer)"
+    input = gets.strip.upcase
```
Lukeghenco: I would move this input stuff to the cli portion of your program. The models shouldn't really contain inputs and outputs.

lib/players/computer.rb
```diff
@@ -0,0 +1,122 @@
+require 'pry'
+
+module Players
```
Lukeghenco: this class should have access to the game board and be able to get the win_combinations from that

lib/game.rb
kkterai: Hi Luke,

Thanks for your feedback!

So, to clarify your recommendation below, do you mean that we should put
the entire method into the CLI?

Kenlyn

lib/players/computer.rb
kkterai: Hi Luke,

Can you please help me to see what you see here? As far as I can tell, this
class *does* have access to the game board and is able to get the
win_combinations from it.

   - In environment.rb
   <https://github.com/kkterai/ttt-with-ai-project-v-000/blob/master/config/environment.rb>,
   it says  "require_all 'lib'" so I don't understand it as an environment
   issue.
   - The board is also referenced throughout the class, including
   #win(board) which iterates through win_combinations.

Thanks in advance for your help in understanding this and we'll revise with
your response. :-) Best, Kenlyn

PR #295, opened 2016-09-13
Done!

radditude:
Done!


bin/tictactoe
```diff
@@ -1,3 +1,61 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```
radditude: I see what you mean! I made a CLI class and did some refactoring so it's more readable.

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+    attr_accessor:cells
+    
+    def initialize
+        self.reset!
```
radditude: Thanks :)

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+        @cells.count {|cell| cell != " "}
+    end
+    
+    def taken?(input)
+        cell_value = position(input)
+        cell_value == " " ? false : true
+    end
+    
+    def valid_move?(input)
+        if input.to_i.between?(1, 9) && !taken?(input)
```
radditude: Fixed

lib/players/human.rb
```diff
@@ -0,0 +1,24 @@
+module Players
+    class Human < Player
+        
+        attr_accessor:input
+        
+        def receive_input
```
radditude: I definitely see the problem here, but I'm having trouble implementing the solution. Whenever I move the "input = gets.strip" line to the CLI class, the program works just fine when I run it via bin/tictactoe. However, when I try to run the test suite, rspec hangs when it hits the tests involving the Players::Human class. The tests don't fail, rspec just hangs until I ctrl+C out of it.

Currently I have two different #move methods - one that passes spec and one that works when I run bin/tictactoe. Any ideas on how I can resolve the rspec issue?

radditude:
Implemented refactoring as suggested by Luke

PR #297, opened 2016-09-15
Done

rgylling:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
Lukeghenco: for lines 4-30 I would convert this into a CLI class with callable instance methods that lives in the `lib` folder. Then on line 5 you could just call this: 
```ruby 
CLI.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
```
Lukeghenco: Good use of existing method 👍

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  def full?
+    !@cells.include?(" ") ? true : false
+  end
+
+  def turn_count
+    @cells.count{|cell| cell != " "}
+  end
+
+  def taken?(input)
+    if position(input) == "X" || position(input) == "O"
```
Lukeghenco: a conditional statement will return true or false implicitly by default. This removes the need for a if statement with explicit returns.

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+      return true
+    else
+      false
+    end
+  end
+
+  def valid_move?(input)
+    if input.to_i.between?(1,9) && !taken?(input)
```
Lukeghenco: another example of implicit boolean returns

lib/players/computer.rb
```diff
@@ -0,0 +1,137 @@
+module Players
+  class Computer < Player
+
+    def computer_options(board)
+        move = ["7", "1", "9", "3", "4", "8", "6", "2", "5"]
+        moves = move.find {|slot| !board.taken?(slot)}
+        moves
+    end
+
+    def move(board)
```
Lukeghenco: I would try to find a way to iterate over the WIN_COMBINATIONS for best possible win. This works but it is extremely verbose and a lot to read.

lib/players/human.rb
```diff
@@ -0,0 +1,24 @@
+module Players
+  class Human < Player
+
+    attr_accessor:input
+
+    def user_input
```
Lukeghenco: I would move these two method to lives inside of you CLI or GAME class. It seems out of place here.

lib/board.rb
rgylling: fixed this  :)

lib/board.rb
rgylling: fixed this  :)

lib/players/human.rb
rgylling: I put these methods here to pass the tests
( Players::Human
  inherits from Player
  #move
    asks the user for input and returns it
)

lib/board.rb
rgylling: Thanks luke :)

bin/tictactoe
rgylling: done

lib/players/computer.rb
rgylling: done

PR #298, opened 2016-09-15
Done

jinstrider2000:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+### Different game initializations available ###
+
+def start
```
Lukeghenco: I would convert this method into a CLI class with smaller methods that lives in the `lib` folder. You will then be able to just call this on line 5 

```ruby 
CLI.new.start 
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    self.cells = Array.new(9, " ")
```
Lukeghenco: we could just call our `reset!` method here

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def reset!
+    self.cells = Array.new(9, " ")
+  end
+  
+  # This #display method I essentially copy and pasted from the earlier tic-tac-toe projects before I learned about Enumberables and iterative methods like #each
+  def display
```
Lukeghenco: it is generally good practice to leave inputs and outputs in a common interface file (like the new CLI class) where user interaction is expected.

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+    if current_player.class == Players::Computer
+      puts input
+    end
+    self.board.update(input, current_player)
+    self.board.display
+  end
+
+  def play
+    puts "GAME START!"
```
Lukeghenco: I would move this to our CLI class as well.

PR #302, opened 2016-09-27
Project submitted

marcnjaramillo:


bin/tictactoe
```diff
@@ -1,3 +1,31 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+class Board
+  attr_accessor :cells
+
+  def initialize()
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def position(input)
+    if @cells[input.to_i] == "O" || @cells[input.to_i] == " " || @cells[input.to_i] == "X"
+     @cells[input.to_i-1]
+   end
+  end
+
+  def full?
+    @cells.any? {|space| space == " "} ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def taken?(input)
+    if @cells[input.to_i - 1] == "X" || @cells[input.to_i - 1] == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+     @cells[input.to_i-1]
+   end
+  end
+
+  def full?
+    @cells.any? {|space| space == " "} ? false : true
+  end
+
+  def taken?(input)
+    if @cells[input.to_i - 1] == "X" || @cells[input.to_i - 1] == "O"
```
Lukeghenco: we could remove this if statement and it would implicitly return true or false already.

Lukeghenco: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.

Lukeghenco (commented):
Hi Marc,

Great job on your TTT with AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: luke.ghenco@flatironschool.com or on slack: (@luke).

Regards,

Luke Ghenco

PR #303, opened 2016-09-28
Project submitted

bc86:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+	def position(input) 
+		@cells[input.to_i - 1]
+	end
+
+	def full?
+		@cells.all? { |full| full != " " } 
+	end
+
+	def turn_count
+		counter = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+	]
+
+	def initialize( player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new )
+		@player_1 = player_1
+		@player_2 = player_2
+		@board = board
+	end
+
+	def current_player
+		board.turn_count.even? ? @player_1 : @player_2
```
Lukeghenco: Good use of Ternary operator here.

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+	end
+
+	def winner
+		if winning_combo = won?
+			the_winner = board.cells[winning_combo.first]
+		end
+	end
+
+	def turn
+		puts "To Make Your Move Pick a Number Between 1-9"
```
Lukeghenco: I would refrain from putting inputs outputs here if we could redirect this to the cli somehow.

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+      		turn
+    	end
+    	if won?
+      		puts "Congratulations #{winner}!"
+    	elsif draw?
+      		puts "Cats Game!"
+    	end
+	end
+
+	def start
```
Lukeghenco: this method should live in the CLI

PR #304, opened 2016-09-28
Project submitted

christinalcole:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.new.start
```
Lukeghenco: great job on placing this in a different file, but all user interaction, inputs and outputs should be in a CLI class. So this would look like 
```ruby 
CLI.new.start
``` 
instead

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+require 'pry'
+class Board
+
+	attr_accessor :cells
+
+	def initialize
+		reset!
```
Lukeghenco: great use of an existing method.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+		puts " #{self.cells[0]} | #{self.cells[1]} | #{self.cells[2]} "
+		puts "-----------"
+		puts " #{self.cells[3]} | #{self.cells[4]} | #{self.cells[5]} "
+		puts "-----------"
+		puts " #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+	end
+
+#converts user's input to array index, returns value of the board cell at that index
+	def position(position)
+		position = position.to_i - 1
```
Lukeghenco: for this method I would just simplify this to be 
```ruby
self.cells[position.to_i - 1]
``` 
as setting the variable doesn't really increase the readability of the code.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+		self.cells[position]
+	end
+
+#returns true for a full board
+	def full?
+	    self.cells.all? {|value| (value == "X" || value == "O")}
```
Lukeghenco: Great use of the all? iterator here

lib/game.rb
```diff
@@ -0,0 +1,163 @@
...
+  def win
+     WIN_COMBINATIONS.detect do |combo|
+      self.board.cells[combo[0]] == self.board.cells[combo[1]] &&
+      self.board.cells[combo[1]] == self.board.cells[combo[2]] &&
+      self.board.taken?(combo[0]+1)  # combo[0] + 1 required becaue of #position indexing
+    end
+  end
+
+#won? monitors board status for a win
+  def won?
```
Lukeghenco: you should just move the logic from your win method into here. A method that exists just to call one other method is not needed here and doesn't serve a real purpose.

lib/game.rb
```diff
@@ -0,0 +1,163 @@
...
+        self.turn
+      end
+      if self.won?
+        puts "Congratulations #{self.winner}!"
+      else
+        puts "Cats Game!"
+      end
+    end
+
+#start manages the logic of running a game: 0, 1, or 2 players, who goes first/is 'X', initialize Game
```
Lukeghenco: the next several methods should be abstracted away in a CLI class. The Game class should be unconcerned about the inputs/outputs.

PR #305, opened 2016-10-03
Project submitted

vicision:


Rakefile
```diff
@@ -0,0 +1,13 @@
+require_relative './config/environment'
+
+puts "Welcome to TicTacToe"
```
Lukeghenco: this code should be place in a CLI class

bin/tictactoe
```diff
@@ -1,3 +1,156 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(cell)
+    @cells[cell.to_i - 1]
+  end
+
+  def full?
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+        if current_player.is_a?(Players::Human)
+          puts "Invalid move. Try again, #{current_player.token}!"
+        else
+          puts "(Invalid entry)"
+        end
+        turn
+      end
+    end
+
+    def play
```
Lukeghenco: this should be placed in the CLI since it contains a lot outputs. Generally a model class does not output or take inputs only a CLI class would do this.

lib/board.rb
vicision: Good point, thanks. I've refactored this to your suggestion.

Rakefile
vicision: This was in the Rakefile simply for testing, the bin file had this code for purposes of the actual game but it has now been moved to a CLI class in lib/tic_tac_toe_cli.rb. I've removed the welcome message and commented out the rest of the code in the Rakefile but have kept the file for future reference.

bin/tictactoe
vicision: Thanks for this suggestion. I've created the ```lib/tic_tac_toe_cli.rb``` file and placed ```call``` in there along with the other methods relating to the CLI logic, which ```call``` calls. Is that an acceptable way of implementing it?

vicision:
@Lukeghenco Hi Luke, sorry it's taken longer than I thought: struggled for a couple days trying to refactor the outputs into the CLI instead of Game class, and then was hit with power/internet outages. Should be a lot better now, though! I've refactored my code a ton. Let me know what you think. Thanks!

vicision:
@Lukeghenco By the way, due to refactoring my game now fails a handful of rspec tests (only because its output doesn't match what the tests request, really). Everything works great though, and I'd passed them all before implementing the CLI part, so I'm assuming that's OK.

PR #306, opened 2016-10-05
Done

ayafushimi:


bin/tictactoe
```diff
@@ -1,3 +1,87 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+# metod
+def new_game
```
Lukeghenco: refer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    i = input.to_i - 1
+    cells[i]
```
Lukeghenco: it would be ok do just use `cells[input.to_i - 1]`

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def update(input, player)
+    i = input.to_i - 1
+    cells[i] = player.token
```
Lukeghenco: see comment on line 22

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def turn_count
+    cells.count{|c| c!=" "}
+  end
+
+  def taken?(input)
+    position(input) != " "
+  end
+
+  def valid_move?(input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    if win_cell
+      cells[win_cell[0]]
+    else
+      nil
+    end
+  end
+
+  def turn
+    player = current_player
```
Lukeghenco: there is no need to set this to a variable we can just call 
```ruby 
puts "#{current_player.token} turn!"
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+      turn
+    end
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cats Game!"
+    end
+  end
+
+  # def start
```
Lukeghenco: you can remove commented out code

lib/players/human.rb
```diff
@@ -0,0 +1,17 @@
+module Players
+  class Human < Player
+
+    def move(board, game)
```
Lukeghenco: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

ayafushimi: Thank you for your advice!
But I can't understand how to use "until" loop without local variable.
I'd like my program to keep asking until getting valid input.

My code is:
```
    def move(board, game)
      puts "Please select a number."
      puts "The upper left is '1', the bottom right is '9'."
      input = gets

      until board.valid_move?(input)
        puts "Please select a number."
        puts "The upper left is '1', the bottom right is '9'."
        input = gets
      end

      input
    end
```

And I tried following code, but it didn't work.
```
    def move(board, game)
      puts "Please select a number."
      puts "The upper left is '1', the bottom right is '9'."
      gets.chomp

      until board.valid_move?(gets.chomp)
        puts "Please select a number."
        puts "The upper left is '1', the bottom right is '9'."
        gets.chomp
      end

    end
```

PR #307, opened 2016-10-06
Project submitted

BeejLuig:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+play_again = true
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,156 @@
...
+      else
+        @player_2.win_count = @player_2.win_count + 1
+      end
+    elsif self.draw?
+      puts "Cats Game!"
+      @@draw_count += 1
+    end
+  end
+
+  def play_again?(input)
```
Lukeghenco: I would also move code from lines 73 - 137 into that CLI class. It is generally bad practice to mix game logic with user inputs and outputs.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+    puts " 1 | 2 | 3 "
+    puts "------------"
+    puts " 4 | 5 | 6 "
+    puts "------------"
+    puts " 7 | 8 | 9 "
+    puts "\n"
+  end
+
+  def full?
+    @cells.include?(" ") == false
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def taken?(pos)
+    self.position(pos) != " " ? true : false
+  end
+
+  def turn_count
+    @cells.select {|x| x == "X" || x == "O"}.length
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def update(pos, player)
+    pos = pos.to_i - 1
+    @cells[pos] = player.token
+  end
+
+  def valid_move?(input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,156 @@
...
+    self.play_again?(gets.chomp)
+  end
+
+  def winner
+    if self.won?
+      @winning_combo[0]
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

PR #308, opened 2016-10-07
Done

scottibe:


["I'm thinking...What a senseless war Vietnam was.
```diff
@@ -0,0 +1,16 @@
+["I'm thinking...What a senseless war Vietnam was. Just Boy Genius    McNamara's little adventure in the jungles of Southeast Asia.",
```
Lukeghenco: was this meant to be in here?

lib/game.rb
```diff
@@ -0,0 +1,316 @@
...
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+
+
+  def start
```
Lukeghenco: I would move this user interaction logic into a CLI class and then just call `CLI.new.start` on line 5 of your bin file.

PR #312, opened 2016-10-16
Project submitted

yeungn:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```
Lukeghenco: I would move this to the `reset!` method instead

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    binding.pry
```
Lukeghenco: I would remove this code from your app

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    cells[input.to_i-1]
+  end
+
+  def full?
+    !cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,110 @@
...
+      turn
+    end
+    if won?
+      puts "Congratulations #{self.winner}!"
+    else draw?
+      puts "Cats Game!"
+    end
+  end
+
+  def start
```
Lukeghenco: I would move this into a separate class called `class CLI` and abstract this away .

lib/players/computer.rb
```diff
@@ -0,0 +1,47 @@
+module Players
+  class Computer < Player
+
+    # WIN_COMBINATIONS = [
```
Lukeghenco: I would remove your old commented out code.

PR #313, opened 2016-10-16
Project submitted

Richard-Frost:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+
+ TicTacToeCLI.new.call
```
Lukeghenco: Great job extracting this into a separate class 👍

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} " 
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def position_taken?(input)
+    if cells[input.to_i-1] == " "
```
Lukeghenco: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. This should be all you need to do 

```ruby 
def position_taken?(input)
  cells[input.to_i - 1] == " "
end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+        puts "That is not a valid move please try again"
+        turn
+      else
+        @board.update(input, current_player)
+      end
+    end
+
+  def play
+    board.display
+    until over? == true || self.won? == true || self.draw? == true
```
Lukeghenco: our `over?` method already checks for `won?` and `draw?` so if `over?` doesn't return true neither will the other two statements.

the best way would be to refactor this to be 
```ruby
until over?
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+      turn
+      puts " "
+      puts " "
+      board.display
+    end
+    
+    if won? == true
```
Lukeghenco: again `won?` will return true already and a if statement will only move into the block if it is true.

Lukeghenco (commented):
Hi Richard,

Great job on your TTT with AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: luke@flatironschool.com or on slack: (@luke).

Regards,

Luke Ghenco

PR #314, opened 2016-10-18
Done

marineb:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+puts "Welcome to this awesome game of Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+  
+  def position(new_move)
+    cells[new_move.to_i-1]
+  end
+  
+  def full?
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+  
+  def current_player
+    @board.turn_count.even? ? player_1 : player_2
+  end
+  
+  def welcome
```
Lukeghenco: this logic should be moved to our new CLI class

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    @board.turn_count == 9 && !won?
+  end
+  
+  def over?
+    #if board is full, OR if someone won
+    won? || draw?
+  end
+    
+  def winner
+    if something = won?
```
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+      @winner = @board.cells[something.first]
+    end
+  end
+  
+  def turn
+    player = current_player
```
Lukeghenco: current_player already returns something, so we don't need to set this to a variable.

Lukeghenco (commented):
Hi Marine,

Great job on your TTT with AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: luke@flatironschool.com or on slack: (@luke).

Regards,

Luke Ghenco

marineb:
Hi @Lukeghenco. Thanks for the feedback! I've made the updates. Let me know what you think. 

Also I wanted to say for full disclosure, I did look at some of the solution for the end of my computer's player logic. I didn't know this was going to get code reviewed. :o  

Thank you again!

PR #316, opened 2016-10-19
Done

rmullig2:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
-#!/usr/bin/env ruby
+#!/usr/bin/ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic-Tac-Toe"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    cells_populate
+  end
+
+  def reset!
+    @cells.clear
+    cells_populate
```
Lukeghenco: I would remove the cells_populate method and just do this here, since this is the purpose of this method 

Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements. It's more concise and will prevent errors from repeatedly typing `" "`.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    !@cells.any? { |cell| cell == " " }
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb.bak
```diff
@@ -0,0 +1,51 @@
+class Board
```
Lukeghenco: I am not sure why there is two Board files in this application

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+  def is_winner?(player)
+    WIN_COMBINATIONS.each { |combination|
+       if combination.all? { |position| @board.cells[position] == player }
+         return true
+       end }
+     false
+   end
+
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

and you no longer need the is_winner? method.

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+    is_winner?("X") || is_winner?("O")
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
```
Lukeghenco: Instead of calling `Game#is_winner?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

rmullig2:
Luke,


I have made the changes you suggested. Interesting use of the first 
method. I would not have thought of that.


Thanks,


Ray

On Tue, Oct 25, 2016 at 11:15 AM, Luke Ghenco wrote:

@Lukeghenco commented on this pull request.


Hi Ray,


Great job on your TTT with AI project. I did a code review on Github, 
and left some comments for you. You should have received some 
notifications from Github. Please notify me when you have completed the 
fixes or if you have any questions by email: luke@flatironschool.com 
<mailto:luke@flatironschool.com>  or on slack: (@luke 
<https://github.com/luke> ).


Regards,


Luke Ghenco


In bin/tictactoe 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
>
+
+puts "Welcome to Tic-Tac-Toe"


Prefer to encapsulate CLI logic into a class. For example, the only code 
in your bin executable should be:


#!/usr/bin/env ruby

TicTacToeCLI.new.call



And then you'd have a file: lib/tic_tac_toe_cli.rb with:


class TicTacToeCLI
   def initialize
   end

   def call
    # Your game logic
   end
end



In lib/board.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> @@ -0,0 +1,51 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = []


I would just call self.reset! here and delegate to that method to create 
the default empty board.




In lib/board.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> @@ -0,0 +1,51 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    cells_populate
+  end
+
+  def reset!
+    @cells.clear
+    cells_populate


I would remove the cells_populate method and just do this here, since 
this is the purpose of this method


Another way to instantiate an Array with i values is Array.new(9, " ") 
(see Array.new <http://ruby-doc.org/core-2.2.0/Array.html#method-c-new> 
), where 9 is the number of elements in the array and " " is the default 
value for the elements. It's more concise and will prevent errors from 
repeatedly typing " ".




In lib/board.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> +    9.times { @cells << " " }
+  end
+
+  def display
+    puts "-----------"
+    puts "-----------"
+  end
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?


Use the most expressive and efficient iterator for a problem. 
Board#full? is asking if all elements are not empty strings. You can use 
logic that is symmetrical to those semantics with the #all? 
<http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F> 
iterator.


cells.all? {|cell| cell != " "}


This is also more efficient in that the loop will stop after the 
iteration for which a cell is not empty.




In lib/board.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> +    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts "-----------"
+  end
+
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+  def full?
+    !@cells.any? { |cell| cell == " " }
+  end
+
+  def turn_count


Use the highest level and most expressive iterator possible to solve a 
problem. count 
<http://ruby-doc.org/core-2.2.0/Array.html#method-i-count>  on the cells 
is perfect and the rest of this method is not needed. Entire body could 
read:


def turn_count
end





In lib/board.rb.bak 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> @@ -0,0 +1,51 @@
+class Board


I am not sure why there is two Board files in this application




In lib/game.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> +  end
+
+  def over?
+    won? || draw?
+  end
+
+  def is_winner?(player)
+    WIN_COMBINATIONS.each { |combination|
+       if combination.all? { |position| @board.cells[position] == 
player }
+         return true
+       end }
+     false
+   end
+
+
+  def won?


Here you can use Enumerable#detect 
<http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect>  to 
return the winning combination or nil if none is found.


   def won?
       @board.taken?(combo[0] + 1)
     end
   end



and you no longer need the is_winner? method.




In lib/game.rb 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
:
> +       if combination.all? { |position| @board.cells[position] == 
> player }
+         return true
+       end }
+     false
+   end
+
+
+  def won?
+    is_winner?("X") || is_winner?("O")
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner


Instead of calling Game#is_winner? twice, consider using an if statement 
assignment. For example:


if won = won?
   board.cells[won.first]
end



If won? returns true, then and only then will the local variable won be 
assigned, otherwise, the method will just abort with a false return 
value. But if won is assigned, we can use that variable instead of 
calling won? again within the if statement.


—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub 
<https://github.com/learn-co-students/ttt-with-ai-project-v-000/pull/316#pullrequestreview-5670543> 
, or mute the thread

PR #317, opened 2016-10-19
Done.

snsavage:


lib/board.rb
```diff
@@ -0,0 +1,69 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = board_array
```
Lukeghenco: I would call our `reset!` method here

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+    !taken?(position) && position.to_i.between?(1,9)
+  end
+
+  def turn_count
+    @cells.count {|cell| cell == "X" || cell == "O"}
+  end
+
+  private
+  def board_array
+    [" "," "," "," "," "," "," "," "," "]
```

Lukeghenco: I would also move this into the reset! method

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

this will check if the spot is taken too instead of having to iterate over the board once before moving into WIN_COMBINATIONS

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+        end
+      end
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
```
Lukeghenco: Instead of iterating over win_combinations again consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #318, opened 2016-10-20
Done

storywr:


bin/tictactoe
```diff
@@ -1,3 +1,53 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+loop do
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,75 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize(cells = nil)
+    @cells = cells || Array.new(9, " ")
```
Lukeghenco: I would just call the `reset!` method here.

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+  def position(move)
+    move = input_to_index(move)
+    @cells[move]
+  end
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+    @cells.include?(" ") == false
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+    counter = 0
+    @cells.each do |cell|
+      counter += 1 if cell != " "
+    end
+    counter
+  end
+
+  def taken?(move)
+    move = input_to_index(move)
+    if @cells[move] == " "
```
Lukeghenco: couldn't we use the `||` operator here instead of multiple if / elsif statements?

lib/board.rb
```diff
@@ -0,0 +1,75 @@
...
+      false
+    elsif @cells[move] == nil
+      false
+    else
+      true
+    end
+  end
+
+  def valid_move?(move)
+    if taken?(move)
```
Lukeghenco: see comment on line 48

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+      if @board.valid_move?(input + 1) == true
+        @board.cells[input] = current_player.token
+        @board.display
+        break
+      end
+    end
+  end
+
+  def won?
+      WIN_COMBINATIONS.each do |win_combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  def draw?
+    !won? && @board.full?
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
+    x_count = @board.cells.select{|i| i == "X"}
```
Lukeghenco: a better way would be to use our won? method here 

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #319, opened 2016-10-20
Done

anfjord:


bin/tictactoe
```diff
@@ -1,3 +1,26 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+    def initialize
+      self.cells = [" "," "," "," "," "," "," "," "," "]
+    end
+
+    def test
```
Lukeghenco: this method seems unneeded ?

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
+
+  attr_accessor :cells
+
+    def initialize
+      self.cells = [" "," "," "," "," "," "," "," "," "]
```
Lukeghenco: we should just call our `reset!` method here

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def test
+      puts "this is a test!!"
+    end
+
+    def reset!
+      self.cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    def display
+      puts" #{self.cells[0]} | #{self.cells[1]} | #{self.cells[2]} "
+      puts"-----------"
+      puts" #{self.cells[3]} | #{self.cells[4]} | #{self.cells[5]} "
+      puts"-----------"
+      puts" #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+    end
+
+    def position(number)
+      position_index = number.to_i - 1
```
Lukeghenco: cells[number.to_i - 1] should be sufficient here.

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+      cells[position_index]
+    end
+
+    def full?
+
+      cells.all?{|x| x != " "}
+    end
+
+    def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+      turns = cells.collect{|x| x == "X" || x == "O"}
+      turns.count {|x| x == true}
+    end
+
+    def taken?(location)
+      value = cells[(location.to_i - 1)]
```
Lukeghenco: you don't have to set this to a variable before checking it with a conditional statement as you do on line 42

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          attr_accessor :board, :player_1, :player_2
+
+    def initialize(player_1=Player::Human.new("X"),player_2=Player::Human.new("O"),board=Board.new)
+      @player_1 = player_1
+      @player_2 = player_2
+      @board = board
+    end
+
+    def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+      else
+        player_2
+      end
+    end
+
+    def over?
+      won? || draw?
+    end
+
+    def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+      end
+        return false
+    end
+
+    def draw?
+      !won? && board.full?
+    end
+
+    def winner
```
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+      move_location = current_player.move(board)
+      if !board.valid_move?(move_location)
+      turn
+      end
+      board.update(move_location,current_player)
+    end
+
+
+    def play
+      while !over?
```
Lukeghenco: `until over?` might make this more readable

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+        puts "Please enter where you want to put your symbol."
+        turn
+        board.display
+      end
+      if won?
+      winner_letter  = winner
```
Lukeghenco: this line seems uneeded

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+      puts "Congratulations #{winner_letter}!"
```
Lukeghenco: just change this to 
```ruby
puts "Congratulations #{winner}!"
```

PR #320, opened 2016-10-22
Done

amcooper:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+		puts "-----------"
+		puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+	end
+
+	def position(cell)
+		@cells[ cell.to_i - 1 ]
+	end
+
+	def full?
+		!@cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+			  cells[combo[1]] == cells[combo[2]]
+		end
+	end
+
+	def draw?
+		@board.full? && !won?
+	end
+
+	def winner
+		combo = won?
```
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+		!combo ? nil : @board.cells[combo[0]]
+	end
+
+	def turn
+		current_turn = ''
+		player = current_player
```
Lukeghenco: we don't have to current_player to a variable since it already returns a value

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+		end
+
+		if draw?
+			puts "Cats Game!"
+		else
+			puts "Congratulations #{winner}!"
+		end
+	end
+
+  def self.input_validation(input,valid_array)
```
Lukeghenco: this should be in our CLI class, inputs and outputs should be regulated to one location and not sprinkled in our other classes.

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+    while !valid_array.include?(input)
+    	print "\nInvalid input. Try again. "
+    	input = gets.chomp
+    end
+    input
+  end
+
+	def self.start
```
Lukeghenco: see note on line 75

lib/game.rb
amcooper: I'll go ahead and change this.

My reasoning: I assigned the return value of `current_player` to a variable so that I didn't have to call that method twice, because I assumed calling the method required more processing than allocating and deallocating the variable.

amcooper:
@Lukeghenco OK Luke the fixes are complete. Thanks for your help!

PR #321, opened 2016-10-25
Done

andresduarte:


lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def position(pos)
+    self.cells[(pos.to_i) - 1]
+  end
+
+  def full?
+    self.cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end

File: lib/board.rb [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+    self.cells.select {|cell| cell == "X" || cell == "O"}.size
+  end
+
+  def taken?(pos)
+    position(pos) != " " && position(pos) != nil ? true : false
+  end
+
+  def valid_move?(pos)
```
```
pos.to_i.between?(1,9) && !taken(pos)
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+    self.board.turn_count.even? ? self.player_1 : self.player_2
+  end
+
+
+
+  def draw?
+    self.board.full? && !self.won? ? true : false
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+      if e.all? { |pos| self.board.position(pos + 1) == self.player_1.token} || e.all? { |pos| self.board.position(pos + 1) == self.player_2.token}
+        @token_win = self.board.cells[e[0]]
+        return true
+      end
+    end
+    false
+  end
+
+  def over?
+    self.won? || self.draw? ? true : false
```
Lukeghenco: we don't need the 2nd half of this code `? true : false` those are already implicit returns with a conditional statement like this.

PR #322, opened 2016-10-26
Done

cdudhat:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,55 @@
...
+puts "Welcome to Tic Tac Toe!"
+
+#Defines the primary game play including selection of number of players
+def game_play
+  print "How many players? (0, 1 or 2): "
+  input = gets.strip
+  blank = Board.new #Sample blank board to familiarize the player
+  blank.cells = [1, 2, 3, 4, 5, 6, 7, 8, 9] #Entering position values to familiarize the player
+  if input == "0"
```
Lukeghenco: this might be a good place for a switch statement

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  #initialize would create a new board aka cells representing the board
+  def initialize
+    reset!
+  end
+
+  #Defining the display board
+  def display
+    def dash
```
Lukeghenco: why are we defining an inner method here?

lib/game.rb
```diff
@@ -0,0 +1,120 @@
...
+
+  #Initialized a new game with two players and a new board
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  #Determines the current character by judging the number of plays made
+  def current_player
```
Lukeghenco: we could define this simply with a ternary operator `?` and remove setting the variable

lib/game.rb
```diff
@@ -0,0 +1,120 @@
...
+    if self.board.turn_count.even?
+      current_player = @player_1
+    else
+      current_player = @player_2
+    end
+  end
+
+  #define a method to check for winning combinations
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,120 @@
...
+    (!won? && self.board.full?)
+  end
+
+  #Checks if the game is over
+  def over?
+    (won? || draw?)
+  end
+
+  #Determines the winner character
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
cdudhat: Reason: Carried over from "Intro to Ruby"
Resolution: Issue addressed
Comment: Agreed & Closed

lib/game.rb
cdudhat: Reason: None
Resolution: Issue addressed
Comment: Agreed & Closed

bin/tictactoe
cdudhat: Resolution: Issue addressed
Comment: Agreed & Closed

PR #325, opened 2016-10-29
Project submitted

adamfriedl:


bin/tictactoe
```diff
@@ -1,3 +1,77 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+# game = Game.new
+# until game.won?
+# 	game.play
+# end
+
+def greeting
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

adamfriedl: Awesome, thanks Luke. I see how that makes it a lot cleaner.

PR #326, opened 2016-10-31
Done

mahdiASC:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+def top
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,55 @@
...
+  puts "Welcome to Tic-Tac-Toe (Object-Oriented Version)"
+  puts "======================"
+end
+
+def pick_setup
+  x = 3
```
Lukeghenco: It would be preferable to set the variable `x` to be named something more appropriate like `input` and it is much better to set `input = nil`

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    def full?
+        @cells.all?{|cell| cell!=" "}
+    end
+
+    def turn_count
+        @cells.count{|cell| cell!=" "}
+    end
+
+    def taken?(pos)
+        cell=@cells[pos.to_i-1]
```
Lukeghenco: why are we setting a variable here? why not just have 

```ruby
@cells[pos.to_i - 1] != " "
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+        cell!=" "
+    end
+
+    def valid_move?(pos)
+        index=pos.to_i
```
Lukeghenco: similiar situation to comment on line 37

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+        [6,4,2]
+    ]
+
+    def initialize(play1 = Players::Human.new("X"), play2 = Players::Human.new("O"), board=Board.new)
+        @board = board
+        @player_1=play1
+        @player_2=play2
+    end
+
+    def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+
+        if temp == player_1.token
+            player_1
+        else
+            player_2
+        end
+    end
+
+    def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    def draw?
+        board.full? && !won?
+    end
+
+    def over?
+        draw? || won?
+    end
+
+    def winner
```
Lukeghenco: instead of repeating the logic of the `#won?` you should consider using something like this:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+        if temp.nil?
+            nil
+        else
+            @board.cells[temp[0]]
+        end
+    end
+
+
+    def top
```
Lukeghenco: It is  preferable to see the CLI logic separated from the game logic in its own class (e.g. CLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+      puts "Welcome to Tic-Tac-Toe (Object-Oriented Version)"
+      puts "======================"
+    end
+
+    def turn
+        x =""
```
Lukeghenco: see comment on line 72

lib/game.rb
mahdiASC: That shortens the code considerably. Ty!

lib/game.rb
mahdiASC: Didn't know about using variable assignment as a conditional statement. Handy trick, thanks for bringing it to my attention. I'll be making a bunch of changes and comitting based of your feedback.

lib/game.rb
mahdiASC: Could you expand on this. I'm not sure how best to separate the CLI logic from the game logic. Are you referring to the puts? I've updated most of the project based on your suggestions. The game logic was moved to another file "lib/TicTacToeCLI.rb"

PR #327, opened 2016-11-02
Done

jasoncasares:


bin/tictactoe
```diff
@@ -1,3 +1,10 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require 'pry'
+
+new_game = GamePlaySelection.new
```

```ruby
#!/usr/bin/env ruby
       
GamePlaySelection.new.call
```

And then you'd have a file: `lib/game_play_selection.rb` with:

```ruby
class GamePlaySelection
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,87 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    # @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def position(position_number)
+    case position_number
```
Lukeghenco: a better way to approach this would be to use something like this 
```ruby 
self.cells[position_number - 1]
```

lib/board.rb
```diff
@@ -0,0 +1,87 @@
...
+    puts " #{self.position("4")} | #{self.position("5")} | #{self.position("6")} "
+    puts "-----------"
+    puts " #{self.position("7")} | #{self.position("8")} | #{self.position("9")} "
+  end
+
+  def full?
+    self.cells.all? {|character| character == "X" || character == "O"}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

jasoncasares:
Changes have been made

PR #328, opened 2016-11-02
Project submitted

jonbf:


bin/tictactoe
```diff
@@ -1,3 +1,24 @@
 #!/usr/bin/env ruby
 
-require_relative '../config/environment'
+require_relative '../config/environment.rb'
+Game.new.start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+require "pry"
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def reset!
+    @cells = [' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ', ' ']
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(n)
+    cells[n.to_i - 1]
+  end
+
+  def full?
+    !cells.include?(' ')
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def turn_count
+    cells.select{|c| c == "X" || c == "O"}.length
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def taken?(n)
```
Lukeghenco: what is `n` ? Is it a number? Be descriptive of your argument and variable names. This makes your code much more readable.

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    index = n.to_i - 1
+    !(cells[index].nil? || cells[index] == " ")
+  end
+
+  def valid_move?(n)
+    !taken?(n.to_i) && n.to_i.between?(1,9)
+  end
+
+  def update(n, player)
+    index = n.to_i - 1
```
Lukeghenco: not entirely necessary to set this as a variable since you are not reusing it in this method.

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+  def current_player
+    if turn_count.even?
+      @player_1
+    else
+      @player_2
+    end
+  end
+
+  def won?
+    @@WIN_COMBINATIONS.each do |wc|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  def draw?
+    return full? && !won?
+  end
+
+  def over?
+    won? || draw? || full?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+    won? ? board.cells[won?[0]] : nil
+  end
+
+  def turn
+    n = current_player.move(board)
```
Lukeghenco: Think about variable names. Not entirely sure what this is trying to tell me.

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+    end
+
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cats Game!"
+    end
+  end
+
+  def start
```
Lukeghenco: This should belong in a CLI class.

PR #331, opened 2016-11-08
Done

mo6709:


bin/tictactoe
```diff
@@ -1,3 +1,66 @@
 #!/usr/bin/env ruby
-
+require 'pry'
 require_relative '../config/environment'
+
+def start_game
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.start_game
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    @cells[index.to_i-1] 
+  end
+
+  #Check if the entier @cells array occupied 
+  def full?
+  	@cells.all?{ |cell| cell == "X" || cell == "O"}
+  end
+
+  #Count the number of turns that has been played
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+			 tokens = @board.cells.values_at(win_combo[0], win_combo[1], win_combo[2]).uniq
+			 tokens.size == 1 && tokens[0] != " "
+	  end 
+	end
+
+	def draw?
+		!self.won? && @board.full?	
+	end
+
+	def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

bin/tictactoe
mo6709: done plus more abstraction in the TicTacToeCLI class.

lib/board.rb
mo6709: done.

lib/game.rb
mo6709: done. thanks for the insight.

luke:
Guys can you avoid using @luke as github sends me notifications :)

PR #332, opened 2016-11-08
Done

eskayio:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,32 @@
...
+puts "Welcome to Tic Tac Toe! How many human players [0,1,2] will participate?: "
+input = gets.strip
```
Lukeghenco: we can combine this into one step 
```ruby
input = gets.strip.to_i
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+#data and logic of the Tic Tac Toe game board
+#board state held in @cells(Array)
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "] #holds the current state of the game @cells
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def reset!
+   self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "] #clears all tokens from the game @cells
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  def display
+    puts " #{@cells[0]} " + "|" + " #{@cells[1]} " + "|" + " #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} " + "|" + " #{@cells[4]} " + "|" + " #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} " + "|" + " #{@cells[7]} " + "|" + " #{@cells[8]} "
+  end
+
+  def position(pos_from_user)
+    position = pos_from_user.to_i - 1 #decreases num from user to match array indices
```
Lukeghenco: as this variable is only used once you might as well just set up your method to look like this
```@cells[pos_from_user.to_i - 1]```
A variable would only be helpful if used more than once.

File: lib/board.rb [hunk 6]
```diff
@@ -0,0 +1,54 @@
...
+    @cells[position] #returns value from position
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+    full = @cells.any? {|position| position == " "} #if any position us empty returns true
+    full ? false : true #if any position is empty, full == true and #full? returns false
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def taken?(pos_from_user)
+    self.position(pos_from_user) != " " #returns true if the position is occupied (not empty)
+    # if X or O --> true
+    # if empty --> false
+  end
+
+  def valid_move?(pos_from_user) #returns true for user input between 1-9 that is not taken
+    if pos_from_user.to_i.between?(1,9) && self.taken?(pos_from_user) == false
```
Lukeghenco: if we remove the if statement this will just be a conditional statement that return `true` implicitly if both sides are true.

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def current_player
+    board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+      #if yes, returns the matching combo. If not, returns nil.
+    end
+    combo_match
+  end
+
+  def draw?
+    !won? && board.full? #if board is full and has no player has won, returns true
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

eskayio:
Thanks Luke.

I didn't see any notifications from GitHub aside from this one but was able
to get to your feedback using the link at the bottom of your email.

Thanks so much! I'll make the changes and submit another pull request.

Samir

lib/game.rb
eskayio: Luke, I can't figure out why it's necessary to add 1 to combo[0] parameter in Board#taken call. Can you help me understand that?

eskayio:
Luke, I've made all the recommended changes and run 'learn submit' which I believe pushes and submits a pull request. Please let me know if I need to manually submit the pull request though. Thanks again for the feedback!

PR #333, opened 2016-11-14
Done

Jschles1:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeCLI.new.start
```
cernanb: Great job encapsulating the CLI logic into a separate class.

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  # Resets the state of the cells in the board to a 9 element array of " ".
+  def reset!
+    @cells = Array.new(9, " ")
```
cernanb: Try to access your instance variables using your settors/gettors via attr_accessor. 

`self.cells = Array.new(9, " ")`

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  # Takes in user input and returns the value of the board cell.
+  def position(position)
+    @cells[position.to_i-1]
+  end
+
+  # Returns true for a full board,returns false for an in-progress game.
+  def full?
+    !@cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```
cernanb: Consider the use of [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+  def draw?
+    @board.full? && !(self.won?)
+  end
+
+  def over?
+    self.won? || self.draw?
+  end
+
+  def winner
+    self.won?.nil? ? nil : @board.cells[self.won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,14 @@
+module Players
+  # Inherits from Player.
+  class Human < Player
+
+    # Asks the user for input and returns it.
+    def move(board)
+      print "Where would you like to move (1-9)? "
```
```
def move(board)
  gets.chomp
end
```

PR #334, opened 2016-11-15
Done

Logicae:


lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+    attr_accessor :cells
+
+    def initialize
+      @cells = Array.new(9, " ")
+    end
+
+    def reset!
+      @cells.clear
```
Lukeghenco: the code on line 11 already replaces the current @cells array so this line is redudant.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+    attr_accessor :cells
+
+    def initialize
+      @cells = Array.new(9, " ")
```
Lukeghenco: I would just call `reset!` here and delegate to that method to create the default empty board.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+    def display
+      puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+      puts "-----------"
+      puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+      puts "-----------"
+      puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+    end
+
+    def position(board_choice)
+      choice = @cells[board_choice.to_i - 1]
```
Lukeghenco: setting this as a variable is unneeded because a ruby method already returns the value of `@cells[board_choice.to_i - 1]` implicitly

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+      choice
+    end
+
+    def full?
+      @cells.all? {|location| location == "X" || location == "O"}
+    end
+
+    def turn_count
```

```ruby
def turn_count
end
```

lib/cli.rb
```diff
@@ -0,0 +1,39 @@
+class CLI
+
+  def start
+    puts "Hello, welcome to this tic-tac-toe game."
+    puts "Please press '1' for single-player mode, '2' for dual-player mode, or '3' for computer mode?"
+
+    response = gets.strip.to_i
+
+    if !(response == 1 || response == 2 || response == 3)
```
Lukeghenco: checkout the ruby method `between?`

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+                      [2,4,6]
+                              ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+      board.display
+    end
+    if won?
+      puts "Congratulations #{winner}!"
+    else draw?
+      puts "Cats Game!"
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def draw?
+    board.full? && !(won?)
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #335, opened 2016-11-16
Done

kjleitz:


bin/tictactoe
```diff
@@ -1,3 +1,59 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    self.cells.count { |cell| cell != " " }
+  end
+
+  def taken?(pos)
+    self.position(pos) != " "
+  end
+
+  def valid_move?(pos)
+    index = self.input_to_index(pos)
+    index.between?(0, 8) && self.position(pos) == " "
```
cernanb: You should be able to take advantage of your `taken?` method here. 

```
index.between?(0, 8) && !taken?(index) 
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count.even? ? self.player_1 : self.player_2
+  end
+
+  def won?
+    cells = self.board.cells
+    WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+  def over?
+    self.won? || self.draw?
+  end
+
+  def draw?
+    !self.won? && self.board.full?
+  end
+
+  def winner
+    self.won? && self.board.cells[self.won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

bin/tictactoe
kjleitz: Good point! I'll work on this.

lib/board.rb
kjleitz: Solid point! I'll fix this.

lib/game.rb
kjleitz: Ha! Dunno why I didn't think of that. Thanks! I'll fix this.

lib/game.rb
kjleitz: Very cool! I'll change this.

PR #338, opened 2016-11-17
Done

JuanGongora:


bin/tictactoe
```diff
@@ -1,3 +1,127 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+class TicTac_CLI
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells.clear
```
Lukeghenco: @cells.clear is unneeded as line 6 already resets it for you.

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(take)
+    @cells[take.to_i - 1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    (@cells.all? { |turn| (turn == "X" || turn == "O") }) ? true : false
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    @cells.each { |turn| counter += 1 if turn == "X" || turn == "O" }
+    counter
+  end
+
+  def taken?(index)
+    (@cells[index.to_i - 1] == " " || nil) ? false : true
+  end
+
+  def valid_move?(index)
+    (index.to_i.between?(1, 9) && (@cells[index.to_i - 1] == " " || @cells[index.to_i - 1] == nil)) ? true : false
```
Lukeghenco: a ternary operator is not needed for an expression statement like this. The implicit return value for an expression is true or false so if both sides of the && statement return true `valid_move?` returns true else false.

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def turn_count
+    counter = 0
+    @cells.each { |turn| counter += 1 if turn == "X" || turn == "O" }
+    counter
+  end
+
+  def taken?(index)
+    (@cells[index.to_i - 1] == " " || nil) ? false : true
```
Lukeghenco: see comment on line 40

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+    WIN_COMBINATIONS.detect do |win|
+      win.all? {|i| board.cells[i] == "X"} || win.all? {|i| board.cells[i] == "O"}
+    end
+  end
+
+  def draw?
+    (won? || !over?) ? false : true
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def won?
+    WIN_COMBINATIONS.detect do |win|
+      win.all? {|i| board.cells[i] == "X"} || win.all? {|i| board.cells[i] == "O"}
+    end
+  end
+
+  def draw?
+    (won? || !over?) ? false : true
```
Lukeghenco: as noted in the other class. All you need to do is let the expression return the results 

```ruby 
def draw?
  !won? && over?
end 
```

PR #340, opened 2016-11-18
Done

bigrocketgames:


bin/tictactoe
```diff
@@ -1,3 +1,65 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+def start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    index = input.to_i-1
```
Lukeghenco: as this is a variable that is only be used once there is no need to set it. This would be preferred. 
```ruby
@cells[input.to_i - 1]
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+    return @cells[index]
+  end
+
+  def update(input, player)
+    @cells[input.to_i-1] = player.token
+  end
+
+  def valid_move?(input)
```
```
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+    if !taken?(input) && input.to_i.between?(1, 9)
+      true
+    elsif taken?(input)
+      false
+    end
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+    @cells.each do |move|
+      if (move == "X" || move == "O")
+        count += 1
+      end
+    end
+
+    return count
+  end
+
+  def taken?(input)
```
Lukeghenco: this can be really simplified using the ternary operator. 

```ruby 
@cells[input.to_i - 1] != " " ? true : false
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+    if @cells[input.to_i-1] == " "
+      false
+    elsif @cells[input.to_i-1] = "X" || @cells[input.to_i-1] == "O"
+      true
+    end
+  end
+
+  def full?
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    input = character.move(@board)
+    if @board.valid_move?(input) == true
+      @board.update(input, character)
+      @board.display
+    else
+      turn
+    end
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    if (@board.turn_count.odd?)
+      return @player_2
+    else
+      return @player_1
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def over?
+    if (won? || draw? || @board.full?)
+      return true
+    else
+      return false
+    end
+  end
+
+  def winner
```
Lukeghenco: def winner
  !won? || draw? ? nil : @board.cells[won?.first]
end

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+        return true
+      else
+        return false
+      end
+    else
+      return false
+    end
+  end
+
+  def over?
```
Lukeghenco: your draw? method already checks for @board.full?

This is also a good time to use the ternary operator

bigrocketgames:
Luke,

Thank you for the feedback on this project.  I have made the requested
fixes and pushed them to GitHub.  The feedback was very helpful.  It
cleaned up my code and will help me to create cleaner code in the future
as I start I continue to learn and apply the feedback I get.

-- 
Nathan Robbert
Big Rocket Games LLC.

PR #341, opened 2016-11-24
Project submitted

jcpny1:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
+puts 'Hi! Welcome to the wonderful world of Tic-Tac-Toe!'
+
+begin
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts '-----------'
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts '-----------'
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def full?
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end
+  end
+
+  def winner
+    combo = won?
+    @board.position(combo[0]+1) if combo
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

PR #342, opened 2016-11-30
Done

wakenight123:


lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+    if position(input) == "" || position(input) == " "
+      false
+    elsif position(input) == nil
+      false
+    elsif position(input) == "X" || position(input) == "O"
+      true
+    end
+  end
+
+  def valid_move?(input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+    [2,4,6]   #7
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+    elsif board.turn_count == 1
+      @player_2
+    elsif board.turn_count == 0
+      @player_1
+    elsif board.turn_count % 2 != 0
+      @player_2
+    end
+  end
+
+  def won?
```
cernanb: Consider using [`Enumerable#detect`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect) to return the winning combination or `nil` if none is found.

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+         (position3 == "X")) ||
+         ((position1 == "O") && (position2 == "O") && (position3 == "O"))
+
+        return win_combo
+      end
+    end
+    false
+  end
+
+  def winner
```
cernanb: Rather than iterating on the WIN_COMBINATIONS array, consider the following:

```ruby
if won = won?
  board.cells[won.first]
end
```

If `won?` returns true, then and only then will the local variable `won` be assigned, otherwise, the method will just abort with a false return value. But if `won` is assigned, we can use that variable within the if statement.

lib/game.rb
```diff
@@ -0,0 +1,147 @@
...
+    WIN_COMBINATIONS.each do |win_combo|
+      if win_combo == won?
+        return board.cells[win_combo[0]]
+      end
+    end
+    nil
+  end
+
+  def draw?
+    if (won? == true)
```
cernanb: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.

`if won?`

wakenight123:
Hi Cernan!

Thank you so much for your very detailed analysis of my code. I have made
changes as you suggested, but I still don't understand a few things:

1) Separating the CLI logic from the game logic. Which lines should I move
out of the game class?

2) Where are you suggesting I place the self.reset! method? "*I would just
call self.reset! here and delegate to that method to create the default
empty board."*

3) Can you explain this line of code you suggested:

Thank you again so much for your advice and time!

Benny Hung
646 315 4451

PR #343, opened 2016-12-02
Done

brownielox:


bin/tictactoe
```diff
@@ -1,3 +1,47 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  attr_accessor :cells
+
+  @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def position(user_input)
+    num = user_input.to_i - 1
+      self.cells[num]
+  end
+
+  def full?
+    switch = true
+    self.cells.each {|item| item != " " ? switch : switch = false}
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+    switch
+  end
+
+  def turn_count
+    count = 0
+    self.cells.each {|item| item != " " ? count += 1 : count}
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+    count
+  end
+
+  def taken?(location)
+    self.cells[location.to_i - 1] == " " ? false : true
+  end
+
+  def valid_move?(move)
+    move = move.to_i
+    move < 10 && move > 0 && taken?(move) == false ? true : false
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+    [6,7,8],
+    [0,3,6],
+    [1,4,7],
+    [2,5,8],
+    [2,4,6],
+    [0,4,8],
+  ]
+
+  def current_player
+    self.board.turn_count.odd? ? self.player_2 : self.player_1
```
cernanb: Awesome job using `.odd?` in your current_player method.

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def over?
+    self.draw? || self.won?
+  end
+
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+        outcome = true
+      end
+    end
+    outcome
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
```
cernanb: Consider using `won?` in this method to take advantage of that built out functionality. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+      if the_board[combo[0]] != " " && the_board[combo[0]] == the_board[combo[1]] && the_board[combo[0]] == the_board[combo[2]]
+        winner = the_board[combo[0]]
+      end
+    end
+    winner
+  end
+
+  def turn
+    input = nil
+    while self.board.valid_move?(input) == false
```
cernanb: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+
+  attr_accessor :cells
+
```
cernanb: Initializing `@cells` outside of an instance method context makes this a 'class instance variable' instead of a typical instance variable, which is probably not what you intended here. Class instance variables are only accessible through class methods. Instead, initialize your variable in the `initialize` or `reset` method.

PR #344, opened 2016-12-03
Done

gd875:


bin/tictactoe
```diff
@@ -1,3 +1,36 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+  attr_accessor :cells
+
+
+
+  def initialize
+    self.reset!
```
cernanb: Awesome job calling `.reset` in your `initialize` method!

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  def full?
+    @cells.all? do |pos|
+      pos == "X" || pos == "O"
+    end
+  end
+
+  def turn_count
+    count = 0
+    @cells.each do |pos|
+      count +=1 if (pos != " ")
```

```ruby
def turn_count
  cells.count { |pos| #logic here }
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+    count
+  end
+
+  def taken?(pos)
+    @cells[pos.to_i-1] != " "
+  end
+
+  def valid_move?(pos)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+      @player_1 = player_1
+      @player_2 = player_2
+      @board = board
+      @the_winner=""
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+      @player_1
+    else
+      @player_2
+    end
+  end
+
+  def won?
+    winning_combo = nil
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def over?
+    if won? || draw?
+      true
+    else
+      false
+    end
+  end
+
+  def winner
```

```ruby
if winning_combo = won?
  board.cells[winning_combo.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+class Players
+
+  class Human < Player
+    def move(board)
+      desired_move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #345, opened 2016-12-07
Done

Atmajr:


bin/tictactoe
```diff
@@ -1,3 +1,85 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(cell)
+    get_cell = cell.to_i - 1
+    @cells[get_cell]
+  end
+
+  def full?
+    if @cells.include?(" ") then
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+      return false
+    else
+      return true
+    end
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+    return counter
+  end
+
+  def taken?(location)
+    location = location.to_i - 1
+    !(@cells[location].nil? || @cells[location] == " ")
+  end
+
+  def valid_move?(location)
+    location = location.to_i
```
```
location.to_i.between?(1,9) && !taken(location)
```

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+  def over?
+    if ((@board.full? != false) || self.won? || self.draw?) then
+      return true
+    else
+      return false
+    end
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |win_combination|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+    def move(board)
```
```
def move(board)
  gets.chomp
end
```

PR #347, opened 2016-12-09
Done

cbrazil:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+  class Human < Player
+
+    def move(input)
+      puts "It's your turn #{token}! What's your move? (Enter a number between 1 and 9)"
+      input = gets.strip
```
```
def move(board)
  puts "It's your turn #{token}! What's your move? (Enter a number between 1 and 9)"
  gets.strip
end
```

cbrazil: Hi Cernan! Thanks for the comments and edits. I implemented the changes you recommended and pushed the changes to Github.

Thanks!
Callie

PR #348, opened 2016-12-12
Done

briecoyle:


bin/tictactoe
```diff
@@ -1,3 +1,33 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(user_input)
+    self.cells[user_input.to_i - 1]
+  end
+
+  def full?
+    self.cells.include?(" ") ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+  def turn_count
+    9 - self.cells.count(" ")
+  end
+
+  def taken?(user_input)
+    self.position(user_input) != " "
+  end
+
+  def valid_move?(user_input)
+    user_input.to_i.between?(1, 9) && taken?(user_input) == false
```
cernanb: Use the implicit boolean return value of a method rather than checking for equality to true or false. 
```
```

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count%2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def won?
+    WIN_COMBINATIONS.find do |combo|
+      self.board.cells[combo[0]] == self.board.cells[combo[1]] &&
+      self.board.cells[combo[1]] == self.board.cells[combo[2]] && self.board.cells[combo[0]] != " "
+    end
+  end
+
+  def winner
+    self.board.cells[self.won?[0]] if self.won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+module Players
+  class Human < Player
+    def move(user_input, board = nil)
+      user_input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #349, opened 2016-12-16
Done

aml7733:


bin/tictactoe.rb
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env ruby
+
+require_relative '../config/environment'
+
```
cernanb: Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
+class Board
+  attr_accessor :cells, :turn_count, :exes, :oes
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+  def exes
+    @exes = cells.map.with_index.map { |mark, i| i if mark == "X" }
+  end
+
+  def oes
+    @oes = cells.map.with_index.map { |mark, i| i if mark == "O" }
+  end
+
+  def full?
+    return false if @cells.any? { |e| e == " " }
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    true
+  end
+
+  def taken?(pos)
+    self.position(pos) != " "
+  end
+
+  def turn_count
+    turn_count = 0
+    @cells.each { |e| turn_count += 1 if e != " " }
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    turn_count
+  end
+
+  def valid_move?(move)
+    if self.taken?(move) || !((1..9).include?(move.to_i))
```
cernanb: Because `#taken?` and `.include?` return boolean values implicitly, you don't have to return true or false. 

```
!((1..9).include?(move.to_i) && !taken(input)
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+      @player_2
+    end
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def won?
+    exes = @board.cells.map.with_index.map { |mark, i| i if mark == "X" }
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  def draw?
+    if @board.full? && won? == false
+      true
+    else
+    false
+    end
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #350, opened 2016-12-17
Done

pbortnick:


bin/tictactoe
```diff
@@ -1,3 +1,27 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    reset!
```
cernanb: Awesome job calling `reset` in your intialize method!

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def full?
+    @cells.all? do |position|
+      position != " "
+    end
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  def taken?(input)
+    if @cells[input.to_i - 1] ==  " "
+      false
+    else
+      true
+    end
+  end
+
+  def valid_move?(input)
+    !taken?(input) && input.to_i > 0 && input.to_i < 10
```
`input.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [2,4,6], [0,4,8]]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    if board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+  def draw?
+    if won?
+      false
+    elsif board.full? && !won?
+      true
+    end
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+  class Human < Player
+
+    def move(input)
+      puts "Please enter a cell 1-9"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #351, opened 2016-12-17
Project submitted

madifana:


bin/tictactoe
```diff
@@ -1,3 +1,34 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    self.reset!
```
cernanb: Great job calling your `.reset` method here!

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  def taken?(move)
+    if position(move) == "X" || position(move) == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(position)
+    if position.to_i.between?(1, 9) && !taken?(position.to_i)
```
cernanb: You should be able to use the implicit return values from these method calls 

```
position.to_i.between?(1, 9) && !taken?(position.to_i)
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count%2 == 0 ? player_1 : player_2
```

`@board.turn_count.even? ? player_1 : player_2`

PR #352, opened 2016-12-17
Done

ratdog19336:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
+class Board
+  attr_accessor :cells, :board, :token
+
+# Board object = #<Board:0x000000018f0d30 @cells=[" ", " ", " ", " ", " ", " ", " ", " ", " "]>
+# Each method calls on the board object and analyzes it based on an input
+
+  def initialize
+    @cells = Array.new(9, " ")
+    #create the board self and populates the @cells
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+        board.display
+        puts "Congratulations #{winner}!"
+      elsif draw?
+        board.display
+        puts "Cats Game!"
+      end
+    end
+
+    def current_player
+      #self is the game object, and the current_player instance method is looking against the whole self --> board is calling the board object and also why you are able to run the board.turn_count method since youre calling the instance method on the board class (i.e we would have access to all the board methods).  This method looks to check the board to determine whose turn it is, knowing that X always goes first and therefore a divisor determines if it's player_1(X) or player_2(O) and then returns that object, denoted as X and O by the object's token
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+    def draw?
+      if board.full? && !won?
+        true
+      else won?
+        false
+      end
+    end
+
+    def winner
+      if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+require 'pry'
+
+module Players
+
+  class Human < Player
+    attr_accessor :input
+
+    def move(game)
+      @input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #353, opened 2016-12-18
Project submitted

vanitanaidu:


bin/tictactoe
```diff
@@ -1,3 +1,33 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
+class Board
+  attr_accessor :cells
+
+    def initialize
+      @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def full?
+      self.cells.all? { |token| token == "X" || token == "O" }
+    end
+
+    def turn_count
+      counter = 0
+      self.cells.each do |current_move|
+      if current_move == 'X' || current_move == 'O'
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+      new = self.cells[new_location]
+        if new == "X" || new == "O"
+          true
+        else
+          false
+        end
+    end
+
+    def valid_move?(location)
+      new_location = location.to_i - 1
```
cernanb: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value. You can also take advantage of the `taken` method you have defined above. 
```
location.to_i.between?(1,9) && !taken(location)
```

lib/game.rb
```diff
@@ -0,0 +1,145 @@
...
+        if current_turn == "X" || current_turn == "O"
+          counter += 1
+        end
+      end
+      counter
+    end
+
+
+    def current_player
+
```

lib/game.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    def over?
+       self.won? || self.draw?
+
+    end
+
+
+    def won?
+        WIN_COMBINATIONS.any? do |win_combination|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,145 @@
...
+    def draw?
+      if !self.won? && board.full?
+      true
+        else
+        false
+      end
+    end
+
+  def winner
+      WIN_COMBINATIONS.detect do |win_combination|
```
cernanb: Consider using your `won?` method to determine the winning combo and use that combo to grab the winner's token. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,16 @@
+module Players
+
+
+  class Human < Player
+
+    def move(board)
+      board = gets.chomp
+      puts "Where would you like to go? Please enter number 1-9"
```
```
def move(board)
  gets.chomp
end
```

vanitanaidu:
Hi Cernan,

Happy New Year! Sorry for the late reply. I wasn't aware that I was getting
a code review for this lab.

I've made amendments to the code based on your feedback and I pushed it to
github. However, your feedback for the `lib/game.rb` file's `turn_count`
method might not give me the correct return value.I believe the test wants
a number, however a `even?` or `odd?` ruby method will return a true/false
value.

Please let me know if I should make further changes to my code. Thank You!



Vanita

Vanita

PR #354, opened 2016-12-21
finishes cli element

warrend:


lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  def position(value)
+    cells[value.to_i - 1]
+  end
+
+  def full?
+    cells.all? { |board| board === "X" || board === "O" }
+  end
+
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+  # cells = ["X", "O", "X", "O", "X", "X", "O", "O", "X"]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+      @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.cells[combo[1]] != " "
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

cernanb (commented):
Hi Daniel,

Great job on your TTT - AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: cernan@flatironschool.com or on slack: (@cernanb).

Regards,

Cernan Bernardo

PR #355, opened 2016-12-23
Done

krishl:


lib/board.rb
```diff
@@ -0,0 +1,48 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+
+  def initialize
+    reset!
```
cernanb: Awesome job calling `reset!` in your intialize method!

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+  def turn_count
+    cells.count { |cell| cell != " " }
+  end
+
+  def taken?(move)
+    position(move) != " "
+  end
+
+  def valid_move?(move)
+    !taken?(move) && move.to_i.between?(1,9) ? true : false
```
cernanb: You should be able to utilize the implicit return values of the method calls. 

```
!taken?(move) && move.to_i.between?(1,9)
```

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+  def won?
+    WIN_COMBINATIONS.detect { |win_combination| board.cells[win_combination[0]] == board.cells[win_combination[1]] && board.cells[win_combination[0]] == board.cells[win_combination[2]] && board.cells[win_combination[0]] != " "}
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
```

```ruby
if winner = won?
  board.cells[winner[0]]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "Where would you like to make your move? Enter a valid number from 1 to 9."
+      board = gets.chomp
```
```
def move(board)
  puts "Where would you like to make your move? Enter a valid number from 1 to 9."
  gets.chomp
end
```

PR #356, opened 2016-12-27
Done

MichaelBurak:


lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def full?
+    cells.all?{|token| token == "X" || token == "O"}
+  end
+
+  def turn_count
+  cells.count{|token| token == "X" || token == "O"}
+  end
+
+  def taken?(input)
+  if cells[input.to_i - 1]== "X" || cells[input.to_i - 1] == "O" then true
```
cernanb: Utilize the implicit return value of these equality checks. 

```
cells[input.to_i - 1]== "X" || cells[input.to_i - 1] == "O"
```

PR #357, opened 2016-12-29
Project submitted

msimonborg:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    cells.all? { |cell| cell != ' ' }
+  end
+
+  def turn_count
+    cells.select { |cell| cell != ' ' }.size
```

```ruby
def turn_count
  cells.count { |cell| cell != " " }
end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+      combo.all? { |cell| board.cells[cell] == 'X' } || combo.all? { |cell| board.cells[cell] == 'O' }
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
msimonborg: oh, Enumerable#count!  I didn't know or forgot about that. Thanks!

PR #358, opened 2016-12-29
Done

btcline04:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def position(index)
+    @cells[index.to_i - 1]
+  end
+
+  def full?
+    @cells.all? {|c| c != " "}
+  end
+
+  def turn_count
+    @cells.select {|c| c != " "}.size
```

```ruby
def turn_count
  cells.count { |cell| cell != " " }
end
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+  def draw?
+    board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "Please enter 1-9 for your move:"
+      index = gets.strip
```
```
def move(board)
  puts "Please enter 1-9 for your move:"
  gets.strip
end
```

PR #361, opened 2017-01-04
Project submitted

SCDesigns:


lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+	    end
+	end
+
+	def turn
+	  player = current_player
+      current_move = player.move(@board)
+    if !@board.valid_move?(current_move)
+  	  turn
+    else
+      puts "Turn: #{@board.turn_count+1}\n"
```
cernanb: We would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+class Board
+	attr_accessor :cells
+
+	def initialize
+		reset!
```
cernanb: Awesome job using your `reset` method in `initialize`!

PR #363, opened 2017-01-08
Done

marielsegovia:


lib/board.rb
```diff
@@ -0,0 +1,58 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " ",]
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+  def reset!
+    self.cells.clear
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " ",]
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.all? {|cells| cells == "X" || cells == "O" }
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  def taken?(input)
+    if @cells[input.to_i - 1] == "X" || @cells[input.to_i - 1] == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(input)
+    if (input.to_i - 1).between?(0,8) && !taken?(input)
```
cernanb: Consider using the implicit return values of the two expressions

```
```

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+    [6, 7, 8],
+    [0, 3, 6],#vertical
+    [1, 4, 7],
+    [2, 5, 8],
+    [0, 4, 8],#diagonal
+    [2, 4, 6],
+  ]
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+  def draw?
+    if @board.full? && !won?
+      true
+    else
+      false
+    end
+  end
+
+  def winner
+    WIN_COMBINATIONS.each do |combination|
```
cernanb: Consider using your `won?` method as a helper to return the winner of the game. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+    WIN_COMBINATIONS.detect do |combination|
+      @board.cells[combination[0]] == @board.cells[combination[1]] &&
+      @board.cells[combination[1]] == @board.cells[combination[2]] &&
+      @board.taken?(combination[0] + 1)
+  end
+
+  end
+
+  def draw?
+    if @board.full? && !won?
```
cernanb: You should be able to use the implicit return values of the two expressions here as well

```
@board.full? && !won?
```

PR #364, opened 2017-01-09
Done

senhad:


lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+
+
+attr_accessor :cells
+
+def reset!
+	@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+end 
+
+def full?
+   @cells.all? do |position|
+   position == "X" || position == "O"
+   end
+end 
+
+def turn_count
+   counter = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+  @board = board
+  @player_1 = player_1
+  @player_2 = player_2
+end
+
+def current_player
+  board.turn_count % 2 == 0 ? player_1 : player_2
```

PR #367, opened 2017-01-16
Done

ykumawat:


lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+	attr_accessor :cells
+
+	def reset!
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	end
+
+	def full?
+	   @cells.all? do |position|
+	   position == "X" || position == "O"
+	   end
+	end
+
+	def turn_count
+	   counter = 0
```

```ruby
def turn_count
end
```

PR #369, opened 2017-01-16
Finished (Unbeatable AI (i think!) and wargames)

jonfoster9999:


bin/tictactoe
```diff
@@ -1,3 +1,71 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
+
+
+class Board
+	attr_accessor :cells, :answer_token
+
+	def initialize
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
...
+		@answer_token = []
+	end
+
+	def reset!
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,86 @@
...
+		answers = answers.map { |a| a + 1}
+		answers
+	end
+
+	def position(position)
+		@cells[position.to_i - 1]
+	end
+
+	def full?
+		!self.cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+		@current_player = player_1
+		@winning_symbol
+	end
+
+	def over? 
+		self.won? || self.draw? ? true : false
+	end
+
+	def won?
+		WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+		end
+		false
+	end
+
+	def draw?
+		!self.board.cells.include?(" ") && !self.won? ? true : false
+	end
+
+	def winner
+		if self.won?
```
cernanb: Another alternative to solving this is

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+require_relative '../player.rb'
+
+module Players
+	attr_reader :move
+
+	class Human < Player
+		
+		def move(board, game=Game.new)
+			input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

jonfoster9999:
Hello Cernan!

Thank you for the feedback! I believe I made all the suggested changes and
pushed it up to the repository. Do I make another pull request?

Thanks!
Jonathan

PR #370, opened 2017-01-17
Project submitted

tracypholmes:


lib/tic_tac_toe_cli.rb
```diff
@@ -0,0 +1,82 @@
+class TicTacToeCLI
```
cernanb: Great job encapsulating your CLI logic into a separate class!

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+  # define full here - every element on the board contains "X" or "O"
+  def full?
+    @cells.all? do |mark|
+      mark == 'X' || mark == 'O'
+    end
+  end
+
+  # define turn_count here
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
  cells.count { |occupied_spot| occupied_spot != " " }
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  # taken? method
+  def taken?(input)
+    exxo = position(input)
+    exxo != " " ? true:false
+  end
+
+  # valid_move? method here
+  def valid_move?(input)
+    if position(input) == " " && input.to_i.between?(1, 9)
```
cernanb: Utilize the implicit return value of the two expressions above
```
def valid_move?(input)
  input.to_i.between?(1,9) && position(input) == " "
end
```

Another alternative is to utilize your `#taken?` method
```
def valid_move?(input)
end
```

tracypholmes: I was on the right track based on your first example. I will use it as it is closer to what I already had - I was just too verbose. Thanks!

PR #371, opened 2017-01-17
Done

liahwallace1:


lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif
+      draw?
+      puts "Cat's Game!"
+    end
+  end
+
+  def self.start
+    puts "Welcome to Tic-Tac-Toe with AI!"
```
cernanb: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def update(user_input, player = "X")
+    index = user_input.to_i - 1
+    if valid_move?(user_input)
+      @cells[index] = player.token
+    end
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(user_input)
+    index = user_input.to_i - 1
+    @cells[index]
+  end
+
+  def full?
+    @cells.include?(" ") == false
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def turn_count
+    9 - @cells.count(" ")
```
cernanb: Consider this alternative using Ruby's [`#count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count). Entire body could read:

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def taken?(user_input)
+    self.position(user_input) != " "
+  end
+
+  def valid_move?(user_input)
+    user_input.to_i.between?(1,9) && self.taken?(user_input) == false
```
cernanb: You should be able to use the bang operator to utilize negation. This will enable you to avoid comparing to a boolean value and utilize the implicit return value. 

```
```

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def board
+    @board
+  end
+
+  def current_player
+    if @board.turn_count.even?
```
cernanb: Consider using a ternary operator to make your code a little cleaner. 

```
```

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+  def draw?
+    won? == false && @board.full? == true
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
```
cernanb: You can also utilize the return value of won? to grab the token of the winning player. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter a number 1-9:"
+      user_input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #372, opened 2017-01-21
Project submitted

jilustrisimo:


lib/board.rb
```diff
@@ -0,0 +1,47 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
```
cernanb: Great job calling reset in your intialiez method!

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def reset!
+    @cells = %w( \ )*9
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    !@cells.any?{|c| c == " "}
```

`cells.all? {|c| c != " "}`

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+  # spec requires arguments to be passed in in this order
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    # test calls on player.token, changed true/false statements to player_1/player_2
+      @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #374, opened 2017-01-27
Project submitted

StreamedLine:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
```
cernanb: Awesome job utilizing `reset` in your initialize method!

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    @cells.count { |cell| cell != " "} 
+  end
+
+  def taken?(pos)
+    pos = pos.to_i - 1
+    @cells[pos] != " "
+  end
+
+  def valid_move?(pos)
+    return false if !/^[123456789]$/.match(pos)
```
cernanb: You can probably check both conditions in one line and utilize the implicit return values of those expressions. 

```
pos.to_i.between?(1,9) && !taken(pos)
```

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [0,4,8], [6,4,2]]
+
+  def initialize(player_1 = Players::Human.new('X'), player_2 = Players::Human.new('O'), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def over?
+    board.full?
+  end
+
+  def won?
+    WIN_COMBINATIONS.any? do |combo|
+      #non are empty
```
cernanb: Here you can consider use of [`Enumerable#detect`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect) to return the winning combination or `nil` if none is found.

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+      combo.collect{|e| @board.cells[e]}.uniq.count == 1
+    end
+  end
+
+  def draw?
+    over? && !won?
+  end
+
+  def winner
+    WIN_COMBINATIONS.collect do |combo|
```
cernanb: Here you can take advantage of the `won?` method to provide the winner. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "Please enter move: (1-9)"
+      pos = gets
```
cernanb: You actually don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #376, opened 2017-02-01
Project submitted

gloverab:


lib/board.rb
```diff
@@ -0,0 +1,50 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," ",]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} \n"
+    puts "-----------\n"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} \n"
+    puts "-----------\n"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def reset!
+    self.cells = [" "," "," "," "," "," "," "," "," ",]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def position(input)
+    index = (input.to_i-1)
+    self.cells[index]
+  end
+
+  def full?
+    !self.cells.include?(" ") || false
```

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+    puts "\nSweet. When it's your turn, enter a number 1 through 9 to place your token \non the corrosponding spot of the board, '1' being the top left corner and '9' being the bottom right.\n\n"
+  end
+
+  def current_player
+    if self.board.turn_count.even?
```
cernanb: Consider using a 1 line ternary here to make your code a little more concise and still readable. 

`self.board.turn_count.even? ? @player_1 : @player_2`

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def won?
+    cells = self.board.cells
+    WIN_COMBINATIONS.find do |combo|
+      self.board.taken?(combo[0]+1) && cells[combo[0]] == cells[combo[1]] && cells[combo[0]] == cells[combo[2]]
+    end
+  end
+
+  def winner
+    won? ? self.board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      "Please enter a number 1-9 to make a move"
+      answer = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
gloverab: This is something I hadn't even thought of. Super clever. Thank you!

PR #378, opened 2017-02-04
Project submitted

danpaulgo:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(position)
+    cells[position.to_i-1]
+  end
+
+  def full?
+    empty_cells = cells.detect do |cell|
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+      cell == " "
+    end
+    empty_cells.nil?
+  end
+
+  def turn_count
+    cells.select{ |cell| cell != " " }.length
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def taken?(position)
+    cells[position.to_i-1] != " "
+  end
+
+  def valid_move?(position)
+    !taken?(position) && position.to_i > 0
```
cernanb: Consider the scenario here if `position` that gets passed in is more than 9

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    [6,7,8],
+    [0,3,6],
+    [1,4,7],
+    [2,5,8],
+    [0,4,8],
+    [2,4,6]
+  ]
+
+  def current_player
+    board.cells.select{|cell| !cell.strip.empty?}.count.even? ? player_1 : player_2
```
cernanb: You can probably take advatange of `board#turn_count` here. 

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def winner
+    winning_combo = WIN_COMBINATIONS.detect do |combo|
+      [board.cells[combo[0]], board.cells[combo[1]], board.cells[combo[2]]].uniq.count == 1 && !board.cells[combo[0]].strip.empty?
+    end
+    winning_combo.nil? ? nil : board.cells[winning_combo[0]]
+  end
+
+  def won?
+    !winner.nil?
```
cernanb: Your `won?` method should be the one that returns an array of the winning combo if there is one...nil otherwise. 

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    [0,4,8],
+    [2,4,6]
+  ]
+
+  def current_player
+    board.cells.select{|cell| !cell.strip.empty?}.count.even? ? player_1 : player_2
+  end
+
+  def winner
+    winning_combo = WIN_COMBINATIONS.detect do |combo|
```
cernanb: This method is one that should take advantage of the `won?` method to see if there is a winning combo. If so, it would return the token at either of the 3 positions in the winning combo returned from `won?`

```ruby
if won = won?
  board.cells[won.first]
end
```

danpaulgo:
Hi Cernan,

Thanks so much for the feedback! Just completed all the fixes and pushed to github. Please let me know if everything looks good now.

Best,
Daniel

PR #379, opened 2017-02-07
Project submitted

jmcmanus1:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+class Board
+	attr_accessor :cells
+
+	def initialize()
+		reset!
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+	    @player_1 = player_1
+	    @player_2 = player_2
+	  end
+
+	  def over?
+	    won? || draw?
+	  end
+
+	  def current_player
+	    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #380, opened 2017-02-08
Project submitted

mdjurdj1:


lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+   def turn_count
+      @cells.count {|cell| cell == "X" || cell == "O"}
+    end
+
+    def taken?(input)
+      @cells[input.to_i-1] == "X" || @cells[input.to_i-1] == "O" ? true : false
+    end
+
+    def valid_move?(input)
+      if input.to_i.between?(1,9) && !full? && !taken?(input)
```
cernanb: Same concept as above

`input.to_i.between?(1,9) && !taken?(input)`

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+   def full?
+      @cells.include?(" ") ? false : true
+    end
+
+   def turn_count
+      @cells.count {|cell| cell == "X" || cell == "O"}
+    end
+
+    def taken?(input)
+      @cells[input.to_i-1] == "X" || @cells[input.to_i-1] == "O" ? true : false
```
cernanb: Utilize the implicit return value of the boolean expression

```ruby
def taken?(input)
  @cells[input.to_i-1] == "X" || @cells[input.to_i-1]`
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts ""
+      puts "Player #{@token}, make your move! [1-9]"
+      input = gets.chomp
```
```ruby
def move(board)
  gets.chomp
end
```

PR #382, opened 2017-02-13
Project submitted

tkanen:


lib/board.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
+
+class Board
+attr_accessor :cells, :token, :player
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def full?
+    !self.cells.include? " "
+  end
+
+  def turn_count
+    cells.count { |token| token == "X" || token == "O" }
+  end
+
+  def taken?(user_input)
+    if position(user_input) == " "
```
cernanb: You can use the implicit return of a boolean expression rather than returning true or false explicitly. 
```ruby
def taken?(input)
end
```

lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+  attr_accessor :board, :player_1, :player_2, :index, :user_input
+
+  def initialize(player_1 = Players::Human.new(token = "X"), player_2 = Players::Human.new(token = "O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+  def draw?
+    board.full? && !won?
+  end
+
+  def over?
+    draw? || won? || board.full?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(user_input)
+      puts "Please enter 1-9:"
+      user_input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #394, opened 2017-02-24
Done

achesky:


bin/tictactoe
```diff
@@ -1,3 +1,96 @@
 #!/usr/bin/env ruby
 
-require_relative '../config/environment'
+require "pry"
+
+class Start_game
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def full?
+    @cells.all? {|space| space != " "}
+  end
+
+  def turn_count
+    9 - @cells.count {|cell| cell == " "}
+  end
+
+  def taken?(cell)
+    @cells[cell.to_i-1] != " "? true : false
```
cernanb: Use the implicit return value of the boolean expression instead of returning true or false explicitly. 

`@cells[cell.to_i-1] != " "`

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter the number of the space, 1-9:"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #397, opened 2017-02-25
Done

PhreeMason:


lib/board.rb
```diff
@@ -0,0 +1,58 @@
+require 'pry'
+
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " "," ", " "," ", " "," ", " ", " "]
+    reset!
```
cernanb: I would just call `reset!` here and delegate to that method to create the default empty board. No need to set `@cells` here.

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  def display
+      puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+      puts "-----------"
+      puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+      puts "-----------"
+      puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def reset!
+    @cells = [" ", " "," ", " "," ", " "," ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+      @cells[location] = player.token
+    end
+  end
+
+  def full?
+      @cells.none? { |f| f == " " }
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def won?
+    answer = []
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+      answer.last.all?{ |ele| ele == "X" } || answer.last.all?{ |ele|  ele == "O"}
+  end
+
+  def draw?
+    !won? && @board.full?
+  end
+
+  def winner
+    if won?
```
cernanb: Consider using the following for this method:

```ruby
if won = won?
  board.cells[won.first]
end
```

If `won?` returns true, then and only then will the local variable `won` be assigned, otherwise, the method will just abort with a false return value. But if `won` is assigned, we can use that variable instead of calling `won?` again within the if statement. You won't have to create separate methods to check the winner.

lib/players/human.rb
```diff
@@ -0,0 +1,14 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      puts "#{self.name} Please enter and input"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #398, opened 2017-02-26
Done

sjwakeman:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9," ")
+  end
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
+    if !cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+   @player_1 = player_1
+   @player_2 = player_2
+   @board = board
+   # binding.pry
+   end
+
+   def current_player
+     board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+       @board.taken?(a[0]+1)
+     end
+   end
+
+   def draw?
+     self.board.turn_count == 9 && !self.won?
+   end
+
+   def winner
+     self.won? ? self.board.cells[self.won?[0]]:nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

sjwakeman:
Hey Cernan! Thanks for reviewing the TTT - AI project.

I am curious though perhaps I do not know where to look but I have not been able to find any notification from GitHub. The only notification I have seen was your email to me. Am I missing something?

I guess I have to go back to the https://github.com/sjwakeman/ttt-with-ai-project-v-000 ?

Thanks again for your help with this.

Scott

PR #400, opened 2017-02-27
Done

anniezh12:


lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+
+attr_accessor :cells
+
+  def initialize
+    @cells =  [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+def position(position)
+    @cells[position.to_i-1]
+end
+
+def full?
+@cells.all?{|cell| cell == "X" || cell == "O"}
+end
+
+def turn_count
+  count = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  attr_accessor :board, :player_1, :player_2
+WIN_COMBINATIONS = [ [0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[6,4,2]]
+
+def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+end
+def current_player
+  @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+
+
+ module Players
+class Human < Player
+def move(board)
+  puts "Please Enter a move from 1 to 9"
+  input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #401, opened 2017-02-28
Project submitted

rslim87:


lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+	def full?
+		cells.all? {|spot| spot == "X" || spot == "O"}
+	end
+
+	def turn_count
+		cells.count {|spot| spot == "X" || spot == "O"}
+	end
+
+	def taken?(user_input)
+		if cells[user_input.to_i-1] == "X"
```
cernanb: You should be able to use the implicit return value of the boolean expressions rather than returning true or false explicitly.

`cells[user_input.to_i-1] == "X" || cells[user_input.to_i-1] == "O"`

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+]
+
+  def initialize(p1=Players::Human.new("X"), p2=Players::Human.new("O"), bd=Board.new)
+    @player_1 = p1
+    @player_2 = p2
+    @board = bd
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    win = won?
+    win == nil ? nil : board.cells[win[0]]
```
cernanb: Another way of writing this method is 

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #403, opened 2017-02-28
Done

paaaarv:


lib/board.rb
```diff
@@ -0,0 +1,65 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(number)
+    position =  number.to_i - 1
+    return @cells[position]
+
+  end
+  def full?
+    @cells.each do |position|
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+      if position == " "
+        return false
+      end
+    end
+
+  end
+  def turn_count
+    x = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+    index = position.to_i - 1
+    if @cells[index] == "X" || @cells[index] == "O"
+      return true
+    else
+      return false
+    end
+
+  end
+  def valid_move?(input)
+    if input =~ /[^1-9]/
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+  def current_player
+
```
cernanb: Consider using a ternary operator for this method

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+      return true
+    elsif won?
+      return true
+    else
+      return false
+    end
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |win_combination|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "../player.rb"
+module Players
+
+  class Human < Player
+    def move(board)
+      board.display
+      puts "What move would you like to make?"
+      input = gets
```
```
def move(board)
  gets.chomp
end
```

PR #404, opened 2017-02-28
Done

bentonwong:


bin/tictactoe
```diff
@@ -1,3 +1,30 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+
+  attr_accessor :cells #has an attribute to store the cells of the board
+
+  def initialize #sets the cells of the board to a 9 element array of " "
+    @cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  def index(position) #converts the user's inputted position into its corresponding index within @cells
+    position.to_i - 1
+  end
+
+  def position(position) #takes in user input and returns the value of the board cell'
+    @cells[index(position)]
+  end
+
+  def full? #returns true for a full board; returns false for an in-progress game
+    @cells.each do |cell|
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  def taken?(position) #returns true if the position is X or O
+    if @cells[index(position)] == "X" || @cells[index(position)] == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(position) #returns true for user input between 1-9 that is not taken
+    if (1..9) === position.to_i && @cells[index(position)] == " "
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+  def initialize(player_1=Players::Human.new("X"),player_2=Players::Human.new("O"),board=Board.new)
+    #defaults to two human players, X and O, with an empty board
+    #@"board" contains an instance of a board
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player #returns the correct player, X, for the third move
+      #if sum of x's and o's is even return player_1, else return player_2
```
cernanb: Consider using the `turn_count` method from the board class. 

lib/players/human.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    attr_reader :input, :token
+
+    def initialize(token)
+      @token = token
+    end
+
+    def move(board) #asks the user for input and returns it
+      puts "#{token}: Please select an open position from 1 to 9:"
+      position = gets.strip
```
```
def move(board)
  gets.strip
end
```

lib/game.rb
```diff
@@ -0,0 +1,97 @@
...
+    token_count = board.cells.count{|x| x == "X" || x == "O"}
+    if token_count.even?
+      @player_1
+    else
+      @player_2
+    end
+  end
+
+  def over? #if draw or one of the win combinations, then it is over (e.g. draw occurs when it is full and none of the WIN_COMBINATIONS are met)
+    if won?
```
cernanb: You should be able to just call won? or draw? and use the implicit return values of those methods. 

`won? || draw?`

PR #405, opened 2017-03-01
Tic Tac Toe with Artificial Intelligence

apalski:
Please see README.md for personal comments on the Artificial Intelligence utilised in the project.

### DIFF
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -1,3 +1,24 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+# Runs the game of Tic Tac Toe that is in the /lib folder
+
+puts "Welcome to Tic Tac Toe"
+puts "Do you want a 0, 1 or 2 player game?"
+game = gets.chomp.to_s
+until game == "0" || game == "1" || game == "2"
+  puts "Please enter numbers only - please enter 0, 1 or 2 for the number of players in the game:"
+  game = gets.chomp.to_s
+end
+case game
+  when "0"
+    puts "The tic tac toe demons will now play!"
+    Game.new(Players::Computer.new("X"), player_2 = Players::Computer.new("O")).play
+  when "1"
+    puts "You are playing against a demon!"
+    Game.new(Players::Human.new("X"), player_2 = Players::Computer.new("O")).play
+  when "2"
+    puts "Your game is starting"
+    Game.new.play
+end
--- a/bin/tictactoe
+++ b/bin/tictactoe
@@ -2,23 +2,4 @@
 
 require_relative '../config/environment'
 
-# Runs the game of Tic Tac Toe that is in the /lib folder
-
-puts "Welcome to Tic Tac Toe"
-puts "Do you want a 0, 1 or 2 player game?"
-game = gets.chomp.to_s
-until game == "0" || game == "1" || game == "2"
-  puts "Please enter numbers only - please enter 0, 1 or 2 for the number of players in the game:"
-  game = gets.chomp.to_s
-end
-case game
-  when "0"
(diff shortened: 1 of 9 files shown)


bin/tictactoe
```diff
@@ -1,3 +1,24 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
+
+class Board
+
+	attr_accessor :cells
+
+	def initialize
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+		reset!
```
cernanb: I would just call `reset!` here and delegate to that method to create the default empty board. No need to set the `@cells` instance variable here since `reset!` will do that for you.

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+	end
+
+  # Is the board full - are there any unoccupied positions
+	def full?
+		!@cells.include?(" " || "")
+	end
+
+  # How many turns have been taken
+	def turn_count
+		count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+		if position(input) == "X" || position(input) == "O"
+			true # Position is taken
+		else
+			false # Position is free
+		end
+	end
+
+  # Is the input a valid number between 1 and 9 and is this a valid move
+	def valid_move?(input)
+		if !(1..9).include?(input.to_i) # Valid number between 1 and 9
```
```

File: lib/game.rb [hunk 5]
```diff
@@ -0,0 +1,127 @@
...
+	end
+
+  # Is the game a draw - the board is full and no one has won (no one has gotten one of the winning combinations)
+	def draw?
+		true if board.full? && !won?
+	end
+
+  # Has a player won the game?
+	def won?
+		count_x = 0
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+
+# Human sub-class of Player
+class Players::Human < Player
+
+	# Accepts input from the player for their turn of the game
+	def move(input)
+		input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
apalski: I like the .count method, I need to explore the Ruby classes more when I am coding as I hadn't come across this one.

bin/tictactoe
apalski: Thanks for the feedback Cernan, I probably should have made the time to take my learnings from the gem project back to the ttt project. I have implemented the cli class and moved the game logic to it.

lib/board.rb
apalski: I really didn't get on board with reset! so what you suggested makes more sense to me now.

lib/board.rb
apalski: I definitely need to work on thinking about return values and making use of them.

lib/game.rb
apalski: Again, I need more focus on return values! I needed to refactor #winner to incorporate your suggestion which in turn made my revisit and tighten my logic for this method.

lib/players/human.rb
apalski: I think I will now remember to think harder about return values. Thanks for all the input Cernan, every new thing helps my coding get better!

apalski:
Updated project as per fix comments

PR #406, opened 2017-03-01
Done

jennafrancis:


bin/tictactoe
```diff
@@ -1,3 +1,41 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+  def current_player
+    board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def over? # can probly do a lamda; or maybe just draw/won.
+    draw? || !!won?
+  end
+
+  def won?
+    WIN_COMBINATIONS.detect do |win_combination|
```
cernanb: You can probably refactor this method like so 

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

By default if a winning combination is not detected then the method will return nil.

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+    end
+  end
+
+  def draw?
+    !!won? ? false : @board.full?
+  end
+
+  def winner
+    !!won? ? @board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+
+  class Human < Player
+
+    def move(input)
+      puts "Please enter 1-9:"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  def full?
+    !@cells.include?(" ")
+  end
+
+  def turn_count
+    @cells.count{|token| token == "X" || token == "O"}
+  end
+
+  def taken?(input)
+    position(input) == " " ? false : true
```
cernanb: You can update this method to use the implicit return boolean value. 

```
def taken?(input)
  position(input) != " " 
end
```

PR #407, opened 2017-03-07
Project submitted

crwhitesides:


bin/tictactoe
```diff
@@ -1,3 +1,13 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+          @player_1 = Player::Computer.new("X")
+          play
+        end
+    elsif input.to_i == 2
+      play
+    end
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+  def over?
+    if won? || draw?
+      true
+    else
+      false
+    end
+  end
+
+  def winner
+    if over? && won? != nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+class Player::Human < Player
+  def move(board)
+    puts "What's your move? Select a position on the board (1-9)..."
+    position = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
crwhitesides: Love this about ruby. Much more readable using your suggestions.

PR #408, opened 2017-03-09
Done

BrunoBoehm:


lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+
+	attr_accessor :cells
+
+	def initialize
+		@cells = Array.new(9, " ")
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+	class Human < Player
+
+		def move(board)
+			puts "What's your move big boy?"
+			input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

BrunoBoehm: Hi @cernanb , thanks for the review and the comments.

I've seen all of my tests pass, and it plays well with 0 human players, but when playing against the AI I sometimes get a "stack level too deep (SystemStackError)" where the recursive "turn"  in "if !@board.valid_move?(current_move)" makes an infinite loop. I haven't been able to solve that, it looks like the #move method doesn't get called as it keeps the "current_move" in-memory value or something like this. Have you experimented something similar? 

Have a great Friday, 
Bruno

PR #410, opened 2017-03-10
Done

fastmode:


bin/tictactoe
```diff
@@ -1,3 +1,50 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def start_game
   # Your game logic
  end
end
```

The code in your bin would generally just go in `start_game`.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = Array.new(9, " ")
+  end
+  
+  def initialize
+    reset!
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+## How does that work ?? - arguments? without including modules?
+  def initialize(player_1 = Players::Human.new("X") , player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+  def draw?
+    won? == nil && @board.full? == true
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

bin/tictactoe
fastmode: Great call, I think it looks way more elegant your way.  Also, its about making everything an object so we should have done that in the first place.

lib/board.rb
fastmode: Thank you!

lib/game.rb
fastmode: I like the ternary operator also.  Thank you!

lib/game.rb
fastmode: Thank you, this is something that I think will come with practice.  Not sure I would ever come up with that.  Thank you so much!

PR #411, opened 2017-03-11
Project submitted

villysiu:


bin/tictactoe
```diff
@@ -1,3 +1,34 @@
+
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,42 @@
+class Board
+  attr_accessor :cells
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,42 @@
...
+    !@cells.include? (" ")
+  end
+  def turn_count
+    @cells.collect.count {|x| x == "X" || x == "O"}
+  end
+  def taken?(pos)
+    position(pos) == "X" || position(pos) == "O" ? true : false
+  end
+  def valid_move?(pos)
+     !taken?(pos) && pos.to_i >0 && pos.to_i <10
```
cernanb: Here you can also take advantage of Ruby's `between?` method. 

```
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+    #expect(game.current_player.token).to eq("X")
+    if !@board.full?
+      @board.turn_count.even? ? @player_1 : @player_2
+    end
+  end
+  def over?
+    draw? || won?
+  end
+  def won?
+    WIN_COMBINATIONS.each do |win_combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+  attr_accessor :board
+  class Human < Player
+    # your code here
+    def move(board)
+      input = gets
```
```
def move(board)
  gets
end
```

PR #412, opened 2017-03-11
Done

dalmaboros:


lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def turn_count
+    @cells.count {|e| e != " " }
+  end
+
+  def taken?(input)
+    self.position(input) != " "
+  end
+
+  def valid_move?(input)
+    !self.taken?(input) && input.to_i > 0 && input.to_i < 10
```
cernanb: You could also use the `.between` method here

```ruby
!self.taken?(input) && input.to_i.between?(1,9)
```

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+    board.game = self
+  end
+
+  def current_player
+    if board.turn_count.even?
```
cernanb: You can also use a ternary inside this method
```ruby
board.turn_count.even? ? @player_1 : @player_2
```

PR #413, opened 2017-03-11
Done

satkins16:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def initialize
+    reset!
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  def full?
+    cells.all?{|token| token == "X" || token == "O"}
+  end
+
+  def turn_count
+    @cells.count {|x| x == "X" || x == "O"}
+  end
+
+  def taken?(user_input)
+    if position(user_input) == "X" || position(user_input) == "O"
```
cernanb: Utilize the implicit return of the boolean expressions for this method. 

```ruby
def taken?(user_input)
  position(user_input) == "X" || position(user_input) == "O"
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(user_input)
+    if user_input.to_i.between?(1,9) && taken?(user_input) == false
```
cernanb: You can use the same concept here.

```ruby
def valid_move?(user_input)
end
```

satkins16:
Hey Cernan!

Thanks so much for the help and I updated my lab with all the fixes.

Thanks again and let me know if there's anything else I can change!

PR #414, opened 2017-03-13
Project submitted

FreeBreadsticks:


bin/tictactoe
```diff
@@ -1,3 +1,26 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
+class Board
+  attr_accessor :cells
+
+  #initiazies board as 9 element array with blank spaces
+  def initialize
+    @cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  def full?
+    if @cells.include?(" ")
+      return false
+    end
+    return true
+  end
+
+  #counts how many turns there have been by checking for "X" or "O" in cells
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+        counter += 1
+      end
+    end
+    return counter
+  end
+
+  #determines wether the input given is a valid move on the board by first checking if its a number between 1-9
+  #and then checks if that space is taken
+  def valid_move?(input)
+    if input.to_i.between?(1, 9)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  #returns current player based on the turn count
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  #checks if game has been won
+  def won?
+    #iterate throguh the combination to check if either X or O is the winner
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,27 @@
+require 'pry'
+module Players
+
+  class Human < Player
+
+    def move(board)
+      #retrieves input from human player
+      puts "Enter a number between 1-9:"
+      input = gets.strip
```
cernanb: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression. You also don't have to check if the move is valid here. That should be done in your `turn` method. 
```
def move(board)
  gets.chomp
end
```

PR #416, opened 2017-03-13
Done

TraiLYNNE:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  #class methods
+
+  #instance methods
+  def current_player
+    board.turn_count.even? ? player_1 : player_2
+  end
+
+  def winner
+    board.cells[won?.first] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(input)
+      puts "What is your next move?"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #418, opened 2017-03-18
Project submitted

walwoodr:


bin/tictactoe
```diff
@@ -1,3 +1,15 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def position(input)
+    self.cells[input_to_index(input)]
+  end
+
+  def full?
+    self.cells.none?{|cell| cell == " "}
+  end
+
+  def turn_count
+    self.cells.select{|cell| cell != " "}.size
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def taken?(input)
+    self.position(input) != " "
+  end
+
+  def open?(input)
+    self.position(input) == " "
+  end
+
+  def valid_move?(input)
+    if input.to_i > 0 && input.to_i < 10
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,181 @@
...
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [0,4,8], [2,4,6]]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+  end
+
+  def current_player
+    if self.board.turn_count.even?
```
cernanb: Consider using a ternary for this method

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+require_relative '../player.rb'
+
+module Players
+  class Human < Player
+
+    def move(board, delay = 0)
+      puts "What space would you like to move into, player #{self.token}?"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #419, opened 2017-03-19
Done

itzsaga:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+  def draw?
+    self.board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
```

```ruby
if combo = won?
  board.cells[combo[0]]
end
```

PR #420, opened 2017-03-19
Done

mxdavis:


bin/tictactoe
```diff
@@ -1,3 +1,27 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def full?
+    unless @cells.any? { |i| i == " " }
+      true
+    end
+  end
+
+  def turn_count
+    i = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def taken?(index)
+    i = index.to_i - 1
+    !(@cells[i].nil? || @cells[i] == " ")
+  end
+
+  def valid_move? (index)
+     i = index.to_i
+     !!(i.between?(1, 9) && taken?(i) == false)
```
cernanb: You can avoid double negation by negating the 2nd expression 
`i.between?(1, 9) && !taken?(i)`

PR #422, opened 2017-03-20
Project submitted

etellezp:


lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cat's Game!"
+    end
+  end
+
+  def start
+    puts "Welcome to Tic Tac Toe!"
```
cernanb: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

PR #424, opened 2017-03-21
Project submitted

ryanpmcmahon:


bin/tictactoe
```diff
@@ -1,3 +1,12 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    cells.count{|x| x=="X" or x=="O"}
+  end
+
+  def taken?(input)
+    !(cells[input.to_i - 1] == " ")
+  end
+
+  def valid_move?(input)
+    move = input.to_i
+    move.between?(1,9) && cells[move-1] == " "
```
cernanb: Here you can take advantage of your `taken?` method
`move.between?(1,9) && !taken?(move)`

lib/game.rb
```diff
@@ -0,0 +1,139 @@
...
+      else
+        new(Computer.new("X"), Human.new("O"))
+      end
+    when "2"
+      new(Human.new("X"), Human.new("O"))
+    end
+  end
+
+  def current_player
+    if board.cells.count(player_1.token) > board.cells.count(player_2.token)
```
cernanb: Here you can use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

PR #426, opened 2017-03-22
Done

Jenna424:


lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(string_input)
+    @cells[string_input.to_i - 1]
+  end
+
+  def full?
+    @cells.any? {|cell| cell == " "} ? false : true
```

PR #427, opened 2017-03-24
Done

invaderDev:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " ",]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " ",]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def position(cell)
+    self.cells[cell.to_i - 1]
+  end
+
+  def full?
+    self.cells.none?{|cell| cell == " "}
+  end
+
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+    count
+  end
+
+  def taken?(cell)
+    self.cells[cell.to_i - 1] != " "
+  end
+
+  def valid_move?(cell)
+    !self.taken?(cell) && cell.to_i < 10 && cell.to_i > 0
```
`cell.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+                      [0, 4, 8],
+                      [2, 4, 6]]
+
+  def initialize(player_1 =  Players::Human.new("X"), player_2 =  Players::Human.new("O"), board =  Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
```
cernanb: You can user a ternary in this method to shorten your code

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+  end
+
+  def draw?
+    self.over? && !self.won?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      puts "Pick an empty cell, player #{self.token}:"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #428, opened 2017-03-24
project complete

magpiemac:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize()
+    @cells = cells
```
cernanb: No need to set your instance variable here since calling `reset!` will do that for you.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def display
+      puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+      puts "-----------"
+      puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+      puts "-----------"
+      puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def full?
+     cells.none?{|content| content == " " || content == ""}
```

PR #429, opened 2017-03-24
Done

kkashuda:


lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+   @board.turn_count % 2 == 0 ? @player_1 : @player_2
```
cernanb: You could also use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

lib/players/human.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    def initialize(name)
+      super
+      @name = name
+    end 
+
+    def move(board)
+      puts "Please enter a move 1-9"
+      move = gets.strip
+      return move
```
```
def move(board)
  gets.chomp
end
```

PR #431, opened 2017-03-27
Project submitted

hpaas001:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe"
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    index = input.to_i - 1
+    @cells[index]
+  end
+
+  def full?
+    full_board = @cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    !full_board
+  end
+
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+    [0,4,8],[2,4,6]]
+
+    def initialize(player_1 = Players::Human.new('X'), player_2 = Players::Human.new('O'), board=Board.new)
+    	@player_1 = player_1
+    	@player_2 = player_2
+    	@board = board
+    end
+
+    def current_player
+    	count = self.board.turn_count
```
cernanb: Consider using the `turn_count` method here and a ternary operator to return the appropriate token

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+      	self.board.cells[array[0]] == self.board.cells[array[1]] && self.board.cells[array[1]] == self.board.cells[array[2]] && (self.board.cells[array[0]]=='X' ||self.board.cells[array[0]]=='O')
+    	end
+    end
+
+    def draw?
+    	self.over? && !self.won?
+    end
+
+    def winner
+    	if self.won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+#require_relative '../player.rb'
+
+module Players
+	class Human < Player
+
+		def move(board)
+			user_input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

hpaas001:
Thanks for your feedback, Cernan.

PR #432, opened 2017-03-28
Done

HimaChitalia:


lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(user_input)
+    @cells[user_input.to_i - 1]
+  end
+
+  def full?
+    unless @cells.any? { |cell| cell == " "}
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      input = gets.chomp
+      puts input
```
```
def move(board)
  gets.chomp
end
```

PR #433, opened 2017-03-30
Project submitted

smendoza787:


bin/tictactoe
```diff
@@ -1,3 +1,57 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def reset!
+    self.cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    puts "-----------"
+    puts " #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+  end
+
+  def position(input)
+    self.cells[input.to_i - 1]
+  end
+
+  def full?
+    occupied = 0
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    self.cells.each do |cell|
+      occupied += 1 if cell.include?("X") || cell.include?("O")
+    end
+
+    occupied == 9
+  end
+
+  def turn_count
+    occupied = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,100 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 1 ? self.player_2 : self.player_1
```

lib/game.rb
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  def over?
+    won? || draw? || full?
+  end
+
+  def won?(board = self.board.cells)
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,100 @@
...
+      return false
+    elsif full?(board)
+      return true
+    else
+      return false
+    end
+  end
+
+  def winner(board = self.board.cells)
+    winning_index = won?
```
cernanb: Consider refactoring this method like so:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
```
```
def move(board)
  gets.chomp
end
```

PR #434, opened 2017-03-30
Done

ryanwevans:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    cells[user_input.to_i - 1]
+  end
+
+  def update(location, player)
+    user_input = location.to_i - 1
+    cells[user_input] = player.token
+  end
+
+  def full?
+    cells.include?(" ") == false
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  def turn_count
+    @cells.count {|i| i != " "}
+  end
+
+  def taken?(location)
+    position(location) != " "
+  end
+
+  def valid_move?(location)
+    !taken?(location) && (location.to_i > 0 && location.to_i < 10)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1 if player_1
+    @player_2 = player_2 if player_2
+    @board = board if board
+    self.play
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+    x_count = self.board.cells.count {|cell| cell == "X"}
+    o_count = self.board.cells.count {|cell| cell == "O"}
+    if x_count > o_count
+      return player_2
+    else
+      return player_1
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+    end
+    return false
+  end
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
```
cernanb: You can utilize the `won?` method to determine the winning token. 
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Human < Player
+
+    def move(board=nil)
+      @board = board if board
+      puts "Where would you like to move? Please enter 1-9:"
+      location = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #435, opened 2017-03-31
Project submitted

jingruzhang:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+class Board
+  attr_accessor :cells
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    index = input.to_i - 1
+    @cells[index]
+  end
+
+  def full?
+    !@cells.include?(" ")
```

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    if @board.turn_count.even?
```
cernanb: Consider using a ternary for this method:

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def won?
+    WIN_COMBINATIONS.detect{|combo| @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] && @board.taken?(combo[0]+1)}
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #436, opened 2017-04-02
Done

Jdville03:


lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+    @counter = nil # used to display game number for wargames mode
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+    WIN_COMBINATIONS.detect do |combo|
+      combo.all?{|element| board.cells[element] == "X"} || combo.all?{|element| board.cells[element] == "O"}
+    end
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #440, opened 2017-04-06
Project submitted

abonner1:


bin/tictactoe
```diff
@@ -1,3 +1,96 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells, :choice
+
+  def initialize
+    self.cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def turn_count
+    self.cells.count {|c| c != " "}
+  end
+
+  def taken?(input)
+    self.position(input) != " " ? true : false
+  end
+
+  def valid_move?(input)
+    true if (1..9) === input.to_i && !(self.taken?(input))
```
cernanb: You can refactor to use the implicit return values 

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.current_player = player_1 : self.current_player = player_2
```
cernanb: You can refactor this method like so:

PR #441, opened 2017-04-07
Project submitted

nadinesk:


bin/tictactoe
```diff
@@ -1,3 +1,23 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+	
+
+	
+	def reset! 			
+		cells.clear
+		self.cells = Array.new(9, " ")
+	
+	end
+
+	def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  		puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+	end
+
+	def position(val)
+		cells[val.to_i - 1]
+
+	end
+
+	def full?
+		if cells.all?{|full| full == "X" || full == "O"}
```
cernanb: Using `all?` will return either a true or false value by default so no need to use an if else condition. 

`cells.all?{|full| full == "X" || full == "O"}`

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+			true
+		else
+			false
+		end
+
+	end
+
+	def valid_move?(index)
+		
+		if (index.to_i > 0 && index.to_i < 10) && !taken?(index)
```
cernanb: Same concept above here 
`index.between?(1,9) && !taken?(index)`

lib/game.rb
```diff
@@ -0,0 +1,129 @@
...
+		board.cells.each do |turn|
+			if turn == "X" || turn == "O"
+				counter +=1
+			end
+		end
+		counter		
+	end
+
+	def current_player		
+		if self.turn_count.even? == TRUE
```
cernanb: You can use a ternary operator here 

`self.turn_count.even? ? : player_1 : player_2`

lib/game.rb
```diff
@@ -0,0 +1,129 @@
...
+	end
+
+	def draw?
+  		
+  		board.full? && !won? 
+
+	end	
+
+	def winner
+
```
cernanb: You can take advantage of your `won?` method here
```ruby
if won = won?
  board.cells[won.first]
end
```

PR #442, opened 2017-04-09
Project submitted

bryanhou1:


lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+  def won?
+    WIN_COMBINATIONS.each do |win|
+      if win.all? { |ind| cells[ind] =="X" } || win.all? { |ind|  cells[ind] =="O" }
+        return win
+      end
+    end
+    false
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def position(index)
+    self.cells[index.to_i-1]
+  end
+
+  def position_on_board?(index)
+    index.to_i.between?(1,9)
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "What's your next move?"
+      input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #443, opened 2017-04-09
Done

blinded93:


lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+    self.cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def position(move)
+    self.cells[move.to_i - 1]
+  end
+
+  def full?
+    !self.cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,114 @@
...
+    WIN_COMBINATIONS.detect do |win|
+      self.board.cells[win[0]] == self.board.cells[win[1]] && self.board.cells[win[1]] == self.board.cells[win[2]] && (self.board.cells[win[0]] == "X" || self.board.cells[win[0]] == "O")
+    end
+  end
+
+  def draw?
+    self.board.full? && !self.won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #444, opened 2017-04-09
Done

lukeymac:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
+game.start
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+  attr_accessor :cells
+
+  def initialize(cells = nil)
```

PR #446, opened 2017-04-11
Done

zeniavilla:


bin/tictactoe
```diff
@@ -1,3 +1,27 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+
+    attr_accessor :cells, :turn_count
+
+    def initialize
+        self.reset!
+    end
+
+    def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def position(input)
+        self.cells[input.to_i - 1]
+    end
+
+    def full?
+        self.cells.all? {|cell| cell == "X" || cell == "O"}
+    end
+
+    def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    def current_player
+        self.board.turn_count.even? ? self.player_1 : self.player_2
+    end
+
+    def over?
+        self.draw? || self.won?
+    end
+
+    def won?
+        WIN_COMBINATIONS.any? {|array| array.all? {|cell| self.board.cells[cell] == "X"}} ||WIN_COMBINATIONS.any? {|array| array.all? {|cell| self.board.cells[cell] == "O"}}
```
cernanb: Consider updating this method to return the winning combo if there is winner. 

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    def draw?
+        self.board.full? && !self.won?
+    end
+
+    def winner
```
cernanb: You can take advantage of the `won?` method here:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+    class Human < Player
+        
+        def move(board)
+            puts "Where would you like to move? "
+            input = gets.strip
```
```
def move(board)
  gets.chomp
end
``

PR #447, opened 2017-04-11
Done

worldismiao:


lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def position(n)
+    self.cells[n.to_i-1]
+  end
+
+  def full?
+    !self.cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    9 - self.cells.count(" ")
+  end
+
+  def taken?(n)
+    position(n) != " "
+  end
+
+  def valid_move?(n)
+    n.to_i <=9 && n.to_i >=1 && !taken?(n)
```
`n.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def turn_count
+    self.board.cells.count{|x|x != " "}
+  end
+
+  def current_player
```
cernanb: Consider using a ternary for this method:

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def over?
+    if won? || draw?|| full?
+      true
+    else
+      false
+    end
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/player.rb
```diff
@@ -0,0 +1,41 @@
...
+  attr_reader :token
+  def initialize(token)
+  @token = token
+  end
+end
+
+module Players
+  class Human < Player
+    def move(board)
+      input = gets.strip
```
cernanb: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #448, opened 2017-04-12
Project submitted

r0tiart:


bin/tictactoe
```diff
@@ -1,3 +1,48 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
+class Board
+  attr_accessor :board, :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    @cells =[" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  end
+
+  def to_integer(input)
+    integer = input.to_i
+    integer -= 1
+    integer
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+    !self.cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  def taken?(position)
+    if self.position(position) == " "
+      false
+    else
+      true
+    end
+  end
+
+  def valid_move?(move)
+    if !taken?(move) && self.to_integer(move).between?(0,8)
```
cernanb: You can user the implicit return value here:

`!taken?(move) && self.to_integer(move).between?(0,8)`

lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+    [3,4,5],
+    [6,7,8],
+    [0,3,6],
+    [1,4,7],
+    [2,5,8],
+    [0,4,8],
+    [2,4,6]
+  ]
+
+  def current_player
```
cernanb: You can use a ternary operator here:

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "../player.rb"
+module Players
+  class Human < Player
+    def move(board)
+      @board = board
+      @input =  gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #450, opened 2017-04-14
Done

WilliamBarela:


WilliamBarela:
Hi. I submitted too soon. I should be finished with this by April 16, 2017. I will write another comment to let you know that it is ready for review. Thank you!

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def turn_count
+    self.cells.count{|cell| cell == "O" || cell == "X"}
+  end
+
+  def taken?(cell)
+    (self.position(cell) == "X" || self.position(cell) == "O") ? true : false
+  end
+
+  def valid_move?(cell)
+    (self.taken?(cell) == true || self.position(cell) == nil) ? false : true
```
cernanb: You can use negation to simplify the expression:

```
input.to_i.between?(0,8) && !taken(cell)
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def turn_count
+    @board.cells.count{|cell| cell == "O" || cell == "X"}
+  end
+
+  def current_player
+    (self.turn_count.even? == true) ? player_1 : player_2
```
cernanb: You can use the implicit return rather than check for boolean equality.
`self.turn_count.even? ? player_1 : player_2`

WilliamBarela:
Hi Cernan! 

These are great suggestions! I will make the updates as suggested.

All of the best,
William

PR #451, opened 2017-04-15
Project submitted

laamia:


lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def turn_count
+    @cells.count{|position| position != " "}
+  end
+
+  def taken?(input)
+    @cells[input.to_i - 1] != " "
+  end
+
+  def valid_move?(input)
```
cernanb: You can update this method to utilize two different helper methods. 

```
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+  attr_accessor :board, :player_1, :player_2
+  WIN_COMBINATIONS = [[0,1,2],[3,4,5],[6,7,8],[0,3,6],[1,4,7],[2,5,8],[0,4,8],[2,4,6]]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
```
cernanb: You can use a ternary operator here:

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+    else
+      player_2
+    end
+  end
+
+  def over?
+    won? || board.full?
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+    end
+    return false # if you make it through the entire loop with no luck, it's false
+  end
+
+  def draw?
+    !won? && over?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/player.rb
```diff
@@ -0,0 +1,28 @@
+class Player
+  attr_reader :token
+
+  def initialize(token)
+    @token = token
+  end
+
+  def move(board)
+    puts "Where do you want to move? Please choose a number 1-9."
+    input = gets.strip
```
cernanb: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #453, opened 2017-04-18
Done

asaki444:


lib/Board.rb
```diff
@@ -0,0 +1,51 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+  def initialize
```
cernanb: I would just call `self.reset!` here and delegate to that method to create the default empty board.

lib/Board.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts "------------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+    puts " "
+  end
+
+  def position(pos)
+    cells[index(pos)]
+  end
+
+  def full?
```
cernanb: Use the most expressive and efficient iterator for a problem. `Board#full?` is asking if all elements are not empty strings. You can use logic that is symmetrical to those semantics with the [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) iterator.

lib/Board.rb
```diff
@@ -0,0 +1,51 @@
...
+    cells.detect{|cell| cell == " "}? false : true
+  end
+
+  def turn_count
```
cernanb: Use the highest level and most expressive iterator possible to solve a problem. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
end
```

lib/Game.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def draw?
+       board.full? && !won?
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
```
cernanb: Instead of calling `Game#won?` twice, consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

NakyungLee:
Niki Lee == NakyungLee == github username lol

PR #456, opened 2017-04-20
Project submitted

professionalxmaz:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+Game.begin
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.none? {|i| i == " "} ? true : false
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    [3, 4, 5],#mid
+    [6, 7, 8],#bot
+    [0, 3, 6],#left vert
+    [1, 4, 7],#mid vert
+    [2, 5, 8],#right vert
+    [0, 4, 8],#slash1
+    [2, 4, 6]#slash2
+  ]
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+      @board.cells[combo[1]] == @board.cells[combo[2]] &&
+      @board.taken?(combo[0] + 1)
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+  
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #458, opened 2017-04-20
Done

spensation:


bin/tictactoe
```diff
@@ -1,3 +1,31 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
+require 'pry'
+class Board
+attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
```
cernanb: This is the method I would create the array with empty spaces. 
`@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]`

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(user_input)
+    index = (user_input.to_i) - 1
+    return "#{@cells[index]}"
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+    @cells.none? {|index| index == " "}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+  def taken?(user_input)
+    index = (user_input.to_i) - 1
+    if @cells[index] == " " || @cells[index] == "" || @cells[index] == nil
+      return false
+    else
+      return true
+    end
+  end
+
+  def valid_move?(user_input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+  [2,4,6], #left diagonal
+]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  def draw?
+    if won? || !full?
+      return false
+    else
+      return true
+    end
+  end
+
+  def winner
```
cernanb: You can take advantage of your `won?` method to simplify your code. 
```ruby
if won = won?
  board.cells[won.first]
end
```

PR #459, opened 2017-04-20
Project submitted

kpiipari:


lib/board.rb
```diff
@@ -0,0 +1,51 @@
+require 'pry'
+class Board
+
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def taken?(input)
+    cell = position(input)
+    cell == "X" || cell == "O" ? true : false
+  end
+
+  def valid_move?(input)
+    i = input.to_i
+    i <= 9 && i > 0 && !taken?(input)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,93 @@
...
+      combo_chars = combination.map { |i| board.cells[i] }
+      combo_chars.all? { |c| c == "X" } || combo_chars.all? { |c| c == "O" }
+    end
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,14 @@
+module Players
+  class Human < Player
+
+    def initialize(token)
+      super(token)
+    end
+
+    def move(board)
```
```
def move(board)
  gets.chomp
end
```

PR #460, opened 2017-04-21
Done

alexisadorn:


bin/tictactoe
```diff
@@ -1,3 +1,77 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+  def position(position) #Takes in user input & returns the value of the board cell
+    index = input_to_index(position)
+    cells[index]
+  end
+
+  def full?
+    cells.all? {|index| index == "X" || index == "O"}
+  end
+
+  def turn_count #Returns the amount of turns
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def current_player
+    board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def over?
+    won? || draw? ? true : false
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+                           (position_1 == "O" && position_2 == "O" && position_3 == "O"))
+    }
+    return false
+  end
+
+  def draw?
+    !won? && board.full? ? true : false
+  end
+
+  def winner
```
cernanb: You can simplify this method by using an if statement assignment. 
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+module Players
+  class Human < Player
+    def move(board)
+      user_position = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

bin/tictactoe
alexisadorn: Thanks Cernan! I thought that it looked bad in the bin file, but for some reason creating a new class didn't occur to me. I've updated the files!

PR #462, opened 2017-04-22
Done

Apex-Code:


lib/game.rb
```diff
@@ -0,0 +1,90 @@
...
+	@player_1 = player_1
+
+	@player_2 = player_2
+
+	@board = board
+
+	end
+
+  def current_player
+
```

PR #465, opened 2017-04-24
Done

lucketta:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
+    @cells.clear
```
cernanb: You can remove `@cells.clear` since the line below will reset the @cells array to empty strings.

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def position(index)
+    @cells[valid_index(index)]
+  end
+
+  def full?
+    @cells.all? {|cell| cell != " "}
+    end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def valid_index(index)
+    if index.is_a? String
+      index = index.to_i - 1
+    else
+      index
+    end
+  end
+
+  def valid_move?(index)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [0,4,8], [6,4,2]]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,67 @@
...
+    def over?
+      if (@board.full? || self.won? || self.draw?)
+      true
+      else
+        false
+      end
+
+    end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
```
```
def move(board)
  gets.chomp
end
```

PR #466, opened 2017-04-25
Done

jessawid:


bin/tictactoe
```diff
@@ -1,3 +1,60 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  attr_accessor :cells
+
+  @cells = []
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+      index = input
+    end
+    @cells[index]
+  end
+
+  def full?
+    @cells.all? {|i| i == "X" || i == "O"}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+class Players
+  class Human < Player
+
+    def move(board)
```
```
def move(board)
  gets.strip
end
```

PR #467, opened 2017-04-26
Project submitted

andrewjford:


bin/tictactoe
```diff
@@ -1,3 +1,54 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def taken?(location)
+    self.position(location) != " "
+  end
+
+  def valid_move?(location)
+    !self.taken?(location) && location.to_i.between?(1,9)
+  end
+
+  def full?
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+  ]
+
+  def initialize(p1=Players::Human.new("X"),p2=Players::Human.new("O"),
+      board=Board.new)
+    @player_1 = p1
+    @player_2 = p2
+    @board = board
+  end
+
+  def current_player
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
```
```
def move(board)
  board.display
  print "Enter move (1-9):"
  gets.chomp
end
```

PR #474, opened 2017-04-30
Done

jhyatt1017:


lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+        turn
+      end
+      if draw?
+        puts "Cat\'s Game!"
+      else
+        puts "Congratulations #{winner}!"
+      end
+    end
+
+    def self.start
```
cernanb: Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  def update(user_input, player)
+    cells[user_input.to_i-1]=player.token if !taken?(user_input) && valid_move?(user_input)
+    display
+  end
+
+  def full?
+    !cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+    cells[user_input.to_i-1]
+  end
+
+  def taken?(user_input)
+    position(user_input) != " "
+  end
+
+  def valid_move?(user_input)
+    user_input_to_i = user_input.to_i
+    user_input <= "9" && user_input >= "1" && !taken?(user_input)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+  ]
+
+  attr_accessor :board, :player_1, :player_2
+    def initialize (player_1 = Players::Human.new("X"), player_2=Players::Human.new("O"), board= Board.new)
+      @board = board
+      @player_1 = player_1
+      @player_2 = player_2
+    end
+
+    def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,146 @@
...
+        win_combo.all?{|index| board.cells[index]=="X"} ||
+        win_combo.all?{|index| board.cells[index]=="O"}
+      end
+    end
+
+    def draw?
+      board.full? && !won?
+    end
+
+    def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+#####inherits from Player
+#####move
+#asks the user for input and returns it
+
+module Players
+  class Human < Player
+    def move (board)
+      puts "#{token} is your turn. Please enter a number 1-9:"
+      user_input=gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #475, opened 2017-05-05
Done

crawfordmikyle:


bin/tictactoe
```diff
@@ -1,3 +1,40 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+  def to_index(input)
+    input.to_i - 1
+  end
+
+  def position(input) # converts user position input into index value and return the cell of that position
+    @cells[input.to_i-1]
+  end
+
+  def full?
+    if @cells.include?(" ")
```
cernanb: You can utilize the implicit return value for this method. 

```
def full?
  !@cells.include?(" ")
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+      false
+    else
+      true
+    end
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def taken?(cell)
+    if @cells[to_index(cell)] != " "
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(cell)
```
cernanb: Another way to write this method can be as follows: 
```
```

lib/game.rb
```diff
@@ -0,0 +1,88 @@
...
+    [2,4,6]
+  ]
+
+  def initialize(player_1=Players::Human.new("X"),player_2=Players::Human.new("O"),board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,88 @@
...
+    @count = board.turn_count
+    if @count % 2 == 0
+      return player_1
+    else
+      return player_2
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+  end
+    return false
+  end
+
+  def draw?
+    board.full? == true && won? == false
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #476, opened 2017-05-07
Done

debjohnson33:


bin/tictactoe
```diff
@@ -1,3 +1,43 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
+class Board
+
+attr_accessor :cells
+
+  #sets array w/ 9 slots & '' on each slot
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+      position = position.to_i
+      @cells[position-1]
+  end
+    # returns true for full board
+  def full?
+      !@cells.include?(" ")
+  end
+
+    #counts how many slots are occupied
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+    #checks to see if a cell is occupied w/ X or O returns true||false
+  def taken?(num)
+    num = num.to_i
+    @cells[num-1]=="X" || @cells[num-1]=="O"
+  end
+
+    #same as above to validate the move taken
+  def valid_move?(num)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+        @board.cells[index[1]] == @board.cells[index[2]] && @board.taken?(index[2]+1)
+      end
+    end
+    # returns boolean if its a draw
+    def draw?
+      won?.nil? && board.full?
+    end
+
+    # returns X || O depending on the winner
+    def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+	class Human < Player
+		def move(input)
+			puts "Where do you want to move?"
+			input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #478, opened 2017-05-08
Project submitted

nicholasbair:


bin/tictactoe
```diff
@@ -1,3 +1,16 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    return player_1 if board.turn_count.even?
```
cernanb: You can use a ternary to implicitly return a value `@board.turn_count.even? ? @player_1 : @player_2`.

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+    WIN_COMBINATIONS.detect do |c|
+      b[c[0]] == b[c[1]] && b[c[1]] == b[c[2]] && b[c[0]] != " "
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #481, opened 2017-05-09
Project submitted

dropheaven:


bin/tictactoe
```diff
@@ -1,3 +1,83 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    cells.none? {|c| c == " "}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    moves_made = @board.cells.count("X") + @board.cells.count("O")
+    moves_made.even? ? @player_1 : @player_2
```
cernanb: You can utilize the board turn_count method here
```ruby
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def over?
+    board.full? || won?
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    false
+  end
+
+  def draw?
+    @board.full? && won? == false
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #483, opened 2017-05-09
Project submitted

Eclecticdynasty:


bin/tictactoe
```diff
@@ -1,3 +1,21 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def full?
+    cells.all? {|full| full == "X" || full == "O"}
+  end
+
+  def turn_count
+    cells.count{|cell| cell == "X" || cell == "O"}
+  end
+
+  def taken?(input)
```
cernanb: Utilize the implicit return value rather than explicitly returning a value. 

`cells[input.to_i - 1] == "X" || cells[input.to_i - 1] == "O"`

PR #484, opened 2017-05-11
Project submitted

dschlauderaff:


lib/board.rb
```diff
@@ -0,0 +1,64 @@
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    self.cells = [' ',' ',' ',' ',' ',' ',' ',' ',' ',]
```

lib/board.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def board_pos(input)
+    self.cells[input]
+  end
+
+  def full?
+    !self.cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,159 @@
...
+  ]
+
+  # Defaults to a game with 2 human players and a new board
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.board = board
+    self.player_1 = player_1
+    self.player_2 = player_2
+  end
+
+  def current_player
```
cernanb: You can take advantage of the turn_count method from the Board class here. 

lib/game.rb
```diff
@@ -0,0 +1,159 @@
...
+      position_1 == position_2 && position_2 == position_3 && position_1 != " "
+    end
+  end
+
+  def draw?
+    board.full? && won? == nil
+  end
+
+  # Return winning player or nil if no one has won.
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+
```
```
def move(board)
  gets.chomp
end
```

PR #485, opened 2017-05-12
Done

MikeMirzak:


bin/tictactoe
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require 'pry'
+
+
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(num)
+    num = num.to_i
+    cells[num-1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+    value = true
+    cells.each do |cell|
+      if cell == " "
+        value = false
+      end
+    end
+    return value
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def taken?(num)
+    value = false
+    if position(num) == "X" || position(num) == "O"
+      value = true
+    end
+    return value
+  end
+
+  def valid_move?(num)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,98 @@
...
+    [0,1,2],[0,3,6],[0,4,8],[1,4,7],[2,5,8],[2,4,6],[3,4,5],[6,7,8]
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,98 @@
...
+      self.board.cells[combo[1]] == self.board.cells[combo[2]] &&
+      self.board.taken?(combo[0]+1)
+    end
+  end
+
+  def draw?
+    over? && !won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "enter move"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #487, opened 2017-05-13
Done

squillen:


bin/tictactoe
```diff
@@ -1,3 +1,74 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+# ttt-with-ai-project-v-000
+
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+  class Human < Player
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #489, opened 2017-05-14
Solution

Crittenbach1:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #490, opened 2017-05-15
Done

stephenharrington04:


bin/tictactoe
```diff
@@ -1,3 +1,75 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(user_input)
+    cells[user_input.to_i - 1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def taken?(user_input)
+    if position(user_input) == "X" || position(user_input) == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(user_input)
```
cernanb: You can utilize the implicit return value in this method

```
```

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def player_1=(new_player)
+    @player_1 = new_player
+  end
+
+  def player_2=(new_player)
+    @player_2 = new_player
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+  end
+
+  def full?
+      @board.cells.none? do |value|
+      value == " "
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,106 @@
...
+      position_3 = @board.cells[win_index_3]
+
+      if position_1 == "X" && position_2 == "X" && position_3 == "X" || position_1 == "O" && position_2 == "O" && position_3 == "O"
+        return win_combination
+      end
+      end
+    return false
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #491, opened 2017-05-16
completed

leog7one:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+new_game = Game.new
```

lib/game.rb
```diff
@@ -0,0 +1,155 @@
...
+	    start
+	end
+
+
+
+
+
+
+#Game #current_player returns the correct player, X, for the third move
+	def current_player
```

PR #492, opened 2017-05-16
Done

CurlsandCode:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(user_input)
+    cells[user_input.to_i-1]
+  end
+  
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def turn_count
+    cells.count { |turn| turn !=" "}
+    end
+
+  def taken?(position)
+  cells[position.to_i-1] == "X" || cells[position.to_i-1] == "O"
+  end
+
+   def valid_move?(index)
+  index.to_i.between?(1,9) && !taken?(index) ? true:false
```
cernanb: You can use the implicit return value of the expression here

`index.to_i.between?(1,9) && !taken?(index)`

lib/players/human.rb
```diff
@@ -0,0 +1,7 @@
+ class Players::Human < Player
+    def move(board)
+      puts "Select a cell (1-9)"
+      user_input= gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #495, opened 2017-05-17
Done

szeidman:


bin/tictactoe
```diff
@@ -1,3 +1,49 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def position(input)
+    # takes in user input and returns the value of the board cell
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    # true for full board, false if game in progress
+    !(@cells.include?(nil || " "))
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+  attr_accessor :player_1, :player_2, :board
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+      @board.cells[win_combination[0]] == @board.cells[win_combination[1]] && @board.cells[win_combination[0]] == @board.cells[win_combination[2]] && (@board.cells[win_combination[0]] == "X" || @board.cells[win_combination[0]] == "O")
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #496, opened 2017-05-18
Done

gnappo1:


lib/board.rb
```diff
@@ -0,0 +1,51 @@
+require "pry"
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2 #oppure usa .odd?/.even?
```

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+    #vado nelle combo vincenti e cerco se almeno una di quelle (fatta di indexes) ha, nei rispettivi posti indicati dagli indexes, tutti token uguali.
+    WIN_COMBINATIONS.find do |win_combination|
+      win_combination.all?{|index| board.cells[index] == "X"} ||
+      win_combination.all?{|index| board.cells[index] == "O"}
+    end
+  end
+
+  def winner
+    #accedo direttamente all'array vincente usando un qualsiasi index (0..2) e chiamando won?[]. Perchè? Perchè won? restituisce un array di index vincenti!
+    won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #497, opened 2017-05-18
Done

curlywallst:


bin/tictactoe
```diff
@@ -1,3 +1,38 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+
+  def update(move, player)
+    @cells[index(move)] = player.token
+# in here I call player.token
+  end
+
+  def full?
+    @cells.all?{|i| i == "X" || i == "O"} ? true : false
```
cernanb: You can use the implicit return value of this expression, so the entire method can be:

`@cells.all?{|i| i == "X" || i == "O"}`

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def turn_count
+    turn_count = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+  class Human < Player
+
+
+    def move(board)
+      puts "Please enter 1-9:"
+      move_made = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #498, opened 2017-05-19
Master

CLackrone:


lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+
+	def initialize
+		self.reset!
+	end
+
+	def cells
+		@cells
+	end
+
+	def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+		puts " #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+	end
+
+	#Returns value at that position
+	def position(user_input)
+		index = (user_input.to_i-1)
+		self.cells[index]
+	end
+
+	def full?
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+	def taken?(user_input)
+		space = self.position(user_input)
+		if space == "X" || space == "O"
+			true
+		else
+			false
+		end
+	end
+
+	def valid_move?(user_input)
```
cernanb: You can use the implicit return value of the expressions in this method; so the entire method can be
`!self.taken?(user_input) && user_input.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def draw?
+  	board.full? && !won?
+  end
+
+  def over?
+  	draw? || won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+class Human < Player
+	def move(input)
+		puts "Where would you like to move?"
+		input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #499, opened 2017-05-19
Done

iamtraviscole:


bin/tictactoe
```diff
@@ -1,3 +1,46 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def turn_count
+    @cells.count {|token| token == "X" || token == "O"}
+  end
+
+  def taken?(input)
+    position(input) == "X" || position(input) == "O"
+  end
+
+  def valid_move?(input)
+    input.to_i.between?(1,9) && taken?(input) == false
```
cernanb: You can use the negation operation to avoid using an equality to `false`

`input.to_i.between?(1,9) && !taken?(input)`

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+    [6,4,2]
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+      @board.cells[array[1]] == @board.cells[array[2]] &&
+      @board.taken?(array[0]+1)
+      end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter 1-9"
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #500, opened 2017-05-20
Done

sammay82:


bin/tictactoe
```diff
@@ -1,3 +1,12 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+ attr_accessor :cells
+ @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+   def initialize
+ 	   reset!
+   end
+ 
+ def reset!
+  @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+ def position(input)
+ 	 cells[input.to_i - 1]
+  end
+
+  def update(input, player)
+    cells[input.to_i - 1] = player.token
+  end
+ 
+  def full?
+    if cells.all? {|value| value == "X" || value == "O"}
```
cernanb: Because the expression will evaluate to a boolean, you can use the implicit return value; so the entire method can be 
`cells.all? {|value| value == "X" || value == "O"}`

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+ ]
+ 
+ def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+ 	@player_1 = player_1 
+ 	@player_2 = player_2 
+ 	@board = board 
+ end
+
+  def current_player 
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

PR #501, opened 2017-05-20
Project submitted

chipsaboy:


lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+  class Human < Player
+    def move(board)
+      puts "Please enter a number between 1-9"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #502, opened 2017-05-21
Done

hyoyou:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i-1]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
```
cernanb: There are a couple ways to solve this that don't involve comparing the expression from the include? method to a boolean value. 

The first is 
```ruby
!@cells.include?(" ")
```

The next involves calling the .all? method on @cells.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def turn_count
+    9 - @cells.count(" ")
+  end
+
+  def taken?(input)
+    position(input) == " " ? false : true
```
cernanb: Same concept her as above. 

`position(input) != " "`

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please make a move (1-9):"
+      input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #503, opened 2017-05-22
Project submitted

rocketNeck:


lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def initialize(player_1 = Player::Human.new("X"), player_2 = Player::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+  end
+
+  ###### state #####
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

PR #505, opened 2017-05-26
Done

rosiemin:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize()
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      puts "Cat's Game!"
+    end
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #506, opened 2017-05-26
Project submitted

ramell86:


lib/board.rb
```diff
@@ -0,0 +1,94 @@
+require 'pry'
+
+class Board
+    attr_accessor :cells, :token
+    
+    def initialize
+        
+        @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    end
+ 
+    def reset!
+        cells.clear
+        self.cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,94 @@
...
+             if @cells[index] == " " || @cells[index] == "" || @cells[index] == nil
+            false
+             elsif @cells[index]== "X" || "O"
+          true
+          
+      end
+     end
+
+     def valid_move?(index_number)
+
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+    def current_player
+        board.turn_count.even? ? player_1 : player_2
+     end
+      
+     def over?
+        won? || draw?
+    end
+
+    def won?
+            WIN_COMBINATIONS.each do |win|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,18 @@
...
+module Players
+    class Human < Player
+
+        
+        def move(board)
+            
+            puts "Please choose a space: 1-9"
+           
+            board.display
+            input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #508, opened 2017-05-30
Done

karinnainiguez:


lib/board.rb
```diff
@@ -0,0 +1,65 @@
+
+
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def turn_count
+    turns = @cells.count("X") + @cells.count("O")
+    turns
+  end
+
+  def taken?(input)
+    index = input.to_i - 1
+    if @cells[index] == "X" || @cells[index] == "O"
```
cernanb: You can utilize the return values of the expressions rather than using a conditional. 
def taken?
  index = input.to_i - 1
  @cells[index] == "X" || @cells[index] == "O"
end
```

File: lib/board.rb [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(input)
```
cernanb: You can use the taken? method as a helper for this method
```
```

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  def player_1
+    @player_1
+  end
+
+  def player_2
+    @player_2
+  end
+
+  def turn_count
```
cernanb: You don't need to define this method here since you have it defined in another class.

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+    turn=0
+    board.cells.each do |index|
+      if index=="X"|| index=="O"
+        turn+=1
+      end
+    end
+    turn
+  end
+
+  def current_player
```
cernanb: You can use a single line ternary inside this method. 

lib/game.rb
```diff
@@ -0,0 +1,119 @@
...
+      on_board_3 = board.cells[index_3]
+      (on_board_1 == "X" && on_board_2 == "X" && on_board_3 == "X") || (on_board_1 == "O" && on_board_2 == "O" && on_board_3 == "O")
+    }
+  end
+
+  def draw?
+    self.over? && !(self.won?)
+  end
+
+  def winner
```
cernanb: You can use the return value of won? here. 

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,16 @@
+require_relative "../player.rb"
+
+class Players
+
+  class Human < Player
+
+    def move(board)
+      puts "Where would you like to place your move player #{self.token}? (a number between 1 and 9)"
+      puts ""
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #510, opened 2017-06-01
Done

vtsowell:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    input.to_i - 1
+  end
+
+  def update(index, player)
+    fixed_index = input_to_index(index)
+    !taken?(index) ? @cells[fixed_index] = player.token : nil
+  end
+
+  def full?
+    !@cells.detect{|cell| (cell.nil? || cell == " " || cell == "")}
```

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+        @board.display
+        @board.update(current_move, player)
+        puts "#{player.token} moved #{current_move}"
+        @board.display
+        puts "\n\n"
+      end
+    end
+
+    def current_player
+        @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+      @board = board
+      @player_1 = player_1
+      @player_2 = player_2
+    end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+  def over?
+    won? || draw?
+  end
+
+  def draw?
+    !won? && @board.full?
+  end
+
+  def winner
+    won? ? @board.position((won?[0] + 1).to_s) : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #513, opened 2017-06-03
Full Working Ai

tehcoyote:


bin/tictactoe
```diff
@@ -1,3 +1,40 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+
+class Board
+  attr_accessor :cells
+    def initialize(cells = [" ", " "," "," "," "," "," "," "," "])
+      @cells = cells
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def reset!
+      @cells = [" ", " "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    def position(normalized_index)
+      @cells[normalized_index.to_i - 1]
+    end
+    def taken?(normalized_index)
+      cell = position(normalized_index)
+      return false if(cell != "X" && cell != "O")
+      return true
+    end
+
+    def valid_move?(normalized_index)
```
```
normalized_index.between?(0,8) && !taken(normalized_index)
```

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+    if @board.valid_move?(u_input)
+      move(u_input, current_player)
+      @board.display
+    else
+      turn
+    end
+  end
+
+  # Define your play method below
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+    if @board.turn_count % 2 == 0
+      return @player_1
+    else
+      return @player_2
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  def over?
+    if(won? || draw?)
+      return true
+    end
+    return false
+  end
+
+  def winner
+    win_combo = won?
```
cernanb: Consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #514, opened 2017-06-04
Project submitted

laramontana:


bin/tictactoe
```diff
@@ -1,3 +1,31 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+  attr_accessor :cells
+
+  @cells = []
+
+  def reset!
+    self.cells = Array.new(9, " ")
+  end
+
+  def initialize
+    self.cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    self.cells.count("X") + self.cells.count("O")
+  end
+
+  def taken?(position) #check board position
+    self.cells[position.to_i-1] == "X" || #board position -1 because of test using range 1-9 (user input numbers)
+    self.cells[position.to_i-1] == "O"
+  end
+
+  def valid_move?(position)
+    !taken?(position) && position.to_i >0 && position.to_i <=9
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+    @player_2 = player_2
+    @board = board
+    @board.display
+  end
+
+  def current_player
+     board.turn_count.odd? ? player_2 : player_1
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+  def draw?
+    @board.full? && !won? ? true : false
+  end
+
+  def over?
+     (won? || draw?) ? true : false
+     # IF board is not full, game is in progress (FALSE), ELSE, game is over (TRUE)
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+  class Human < Player
+
+#The human player must implement a #move method that takes in a board argument and allows a human player to enter a move via the CLI.
+#The method should return the value the user enters.
+# Even though the method accepts a board argument, it does not need to use it.
+
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #516, opened 2017-06-06
Done

samsonyuwono:


bin/tictactoe
```diff
@@ -1,3 +1,78 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+    cells[user_input.to_i-1]
+  end
+
+  def full?
+  @cells.all? do |index|
+    index == "X" || index == "O"
+    end
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+     end
+     false
+   end
+
+  def draw?
+    @board.full? && !won?
+   end
+
+   def winner
+     win_array = won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+  def move(board)
+    puts "Please input your move"
+    move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

cernanb (commented):
Hi Samson and Sabrina,

Great job on your TTT - AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can contact me by email: cernan@flatironschool.com or on slack: (@cernanb).

Regards,

Cernan 
cc @sabrina-jones

PR #517, opened 2017-06-06
Project submitted

DoriansGrave:


bin/tictactoe
```diff
@@ -1,3 +1,22 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+  attr_accessor :board, :player_1, :player_2, :team
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [2,4,6], [0,4,8]]
+
+   def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+     @board = board
+     @player_1 = player_1
+     @player_2 = player_2
+   end
+
+   def current_player
```

PR #531, opened 2017-06-23
Project submitted

jelocodes:


bin/tictactoe
```diff
@@ -1,3 +1,101 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+@win_count = {
+	"X" => 0,
+	"O" => 0
+}
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board
+	attr_accessor :cells, :index
+
+	def initialize
+	    @cells = []
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	def input_to_index(input)
+		@index = input.to_i - 1
+	end
+
+	def position(input)
+		@cells[input_to_index(input)]
+	end
+
+	def full?
+		!@cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+	end
+
+	def turn_count
+		@cells.collect{|i| i if i != " " }.compact.size + 1
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+    	[6, 7, 8],
+    	[0, 4, 8],
+    	[2, 4, 6],
+    	[0, 3, 6],
+    	[1, 4, 7],
+    	[2, 5, 8]
+  	]
+
+  	def current_player
+  		if @board.turn_count.even?
```
cernanb: You can use a one line ternary for this method. 

PR #564, opened 2017-07-19
Done

lpassamano:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def reset!
+    self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  def update(position, player)
+    index = self.position_to_index(position)
+    self.cells[index] = player.token
+  end
+
+  def full?
+    self.cells.none? {|cell| cell == " "}
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+  def current_player
+    self.board.turn_count.even? ? self.player_1 : self.player_2
+  end
+
+### GAME END METHODS ###
+  def over?
+    true if self.board.full? || self.won? != false || draw?
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,170 @@
...
+      self.class.draws= (1)
+      true
+    elsif self.won? == false && self.board.full? == false
+      false
+    elsif self.won? != false
+      false
+    end
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

cernanb (commented):
Hi Leigh,

Great job on your TTT - AI project. I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can contact me by email: cernan@flatironschool.com or on slack: (@cernanb).

Regards,

Cernan

PR #565, opened 2017-07-20
working ttt game with ai

nerdybug:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

PR #566, opened 2017-07-20
Computer.rb

byojelly:
Hi guys, 
I believe all is good with this computer.rb file.

Take a look then we can merge into master.

-Alex


byojelly:
So, im not trying to merge 13 commits, just the last one with Finished commit.rb (still trying to undersatnd github workflow)

PR #568, opened 2017-07-20
Done

soumyaveer:


bin/tictactoe
```diff
@@ -1,3 +1,8 @@
 #!/usr/bin/env ruby
```

lib/board.rb
```diff
@@ -0,0 +1,99 @@
...
+      Board::ALL_TOKENS.include?(token)
+    end
+  end
+
+  def update(position, player)
+    self.cells[position_to_index(position)] = player.token
+  end
+
+  def valid_move?(position)
+    !self.taken?(position) && Board::ALL_POSITIONS.include?(position.to_i)
```
cernanb: You could also take advantage of the `between?` method here
`!self.taken?(position) && position.to_i.between?(1, 9)`

bin/tictactoe
soumyaveer: Hi, Cernan (@cernanb) . I made some changes in the code, as per your guidance. Please let me know in case any further improvements might be needed.
Thank you.

PR #603, opened 2017-08-16
Project submitted

tonymastrorio:


bin/tictactoe
```diff
@@ -1,3 +1,29 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+        @cells[index]
+    end
+
+    def full?
+        @cells.all? do |slot|
+            slot == "X" || slot == "O"
+        end
+    end
+
+    def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    if board.turn_count.odd?
```
cernanb: You can also use a ternary operator here 
`board.turn_count.odd? ? @player_2 : @player_1`

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+        board.cells[win_array[0]] != " " && board.cells[win_array[0]] == board.cells[win_array[1]] && board.cells[win_array[1]] == board.cells[win_array[2]]
+    end
+  end
+  
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if win_array = won?
```
cernanb: You can just include the if condition. By default if that evaluates to false then the method will return `nil`.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+    class Human < Player
+        def move(board)
+            board.display
+            puts "Choose an open position on the board"
+            input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #605, opened 2017-08-19
Done

ddeleon267:


lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+          position_1 == position_2 && position_1 == position_3 && self.board.taken?(win_index_1+1)
+    end
+  end
+
+  def draw?
+    self.board.full? && !won?
+  end
+
+  def winner
+    board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #606, opened 2017-08-20
Done

Sidhe-is-me:


bin/tictactoe
```diff
@@ -1,3 +1,44 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+class Board
+  attr_accessor :cells, :available_spaces
+
+  def reset!
+    @cells= [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+  def position(input)
+    index = input.to_i - 1
+    @cells[index]
+  end
+
+  def full?
+    @cells.none?{|i| (i != "X" && i != "O")}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+      board.update(input,self.current_player)
+      board.display
+    else
+      puts "Please try again"
+      self.turn
+    end
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #607, opened 2017-08-22
Done

RMandelbaum:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+require 'pry'
+class Board
+
+attr_accessor :cells, :board, :token
+#
+#
+ def reset!
+   @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+def position(users_input)
+   index = users_input.to_i - 1
+   @cells[index]
+end
+
+def full?
+  @cells.none?{|i| i == " "}
+end
+
+def turn_count
```

```ruby
def turn_count
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "../player.rb"
+module Players
+
+class Human < Player
+
+  def move(board)
+    puts "Please make a move"
+    user_input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #619, opened 2017-08-28
Done

kylekinnear:


lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def won?
+    Game::WIN_COMBINATIONS.detect do |c|
+      c.all? {|i| board.cells[i] == player_1.token} || c.all? {|i| board.cells[i] == player_2.token}
+    end
+  end
+
+  def winner
+    board.cells[won?[0]] unless won? == nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

If `won?` returns a truthy value, then and only then will the local variable `won` be assigned, otherwise, the method will just abort with a false return value. But if `won` is assigned, we can use that variable instead of calling `won?` again within the if statement.

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #648, opened 2017-09-14
Done

lasttruth:


bin/tictactoe
```diff
@@ -1,3 +1,23 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+while !user_input.to_i.between?(1,3)
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
+class Board
+
+  attr_accessor :cells
+
+def initialize (cells = nil)
+  @cells = cells || Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+def input_to_index(input)
+  index = input.to_i - 1
+end
+
+def full?
+   self.cells.none? {|cell| cell.lstrip.empty?}
+end
+
+def turn_count
+  self.cells.select{|c| !c.lstrip.empty?}.count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+def taken?(input)
+  if self.position(input) == " "
+         false
+       else
+         true
+       end
+end
+
+def valid_move?(input)
+  !taken?(input) && (1..9).member?(input.to_i)
```
`position.between?(1,9)`

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+end
+
+def update(input, player)
+   self.cells[input.to_i - 1] = player.token
+end
+
+def reset!
+     self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
weezwo: Looks like a redefinition of your `reset!` method above... Prefer the method used in the other version (`self.cells = Array.new(9, " ")`). It's more concise and will prevent errors from repeatedly typing `" "`.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+     [2,4,6],
+   ]
+   def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    self.player_1 = player_1
+    self.player_2 = player_2
+    self.board = board
+   end
+
+   def current_player
+     if self.board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+   def won?
+     WIN_COMBINATIONS.detect do |combo|
+         @board.cells[combo[0]] == @board.cells[combo[1]] &&
+         @board.cells[combo[1]] == @board.cells[combo[2]] &&
+         @board.taken?(combo[0]+1)
+       end
+   end
+
+   def winner
+     won? ? @board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+module Players
+  class Human < Player
+
+    def initialize(token)
+      super
+    end
+
+    def move(board)
+      move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
```diff
@@ -0,0 +1,58 @@
...
+def full?
+   self.cells.none? {|cell| cell.lstrip.empty?}
+end
+
+def turn_count
+  self.cells.select{|c| !c.lstrip.empty?}.count
+end
+
+def taken?(input)
+  if self.position(input) == " "
```

PR #688, opened 2017-10-04
Done

vlaunyc:


bin/tictactoe
```diff
@@ -1,3 +1,22 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to TicTacToe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+  def full?
+    cells.all?{|c| c == "X" || c == "O"}
+  end
+
+  def turn_count  #returns the amount of used/played cells
+    cells.count{ |c| c == "X" || c == "O"}  #counts when X or O is on board
+  end
+
+  def taken?(user_input)
+    position(user_input) == " " ? false : true   #position(user_input) == "X" || position(user_input)== "O"
```

PR #705, opened 2017-10-10
Done

snickers495:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeCLI.new
```
jilustrisimo: Awesome job!

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+class Board
+  attr_accessor :cells
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+    self.display
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
```
jilustrisimo: You can abstract this into 1 line `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+    new_input = input.to_i - 1
+    @cells[new_input]
+  end
+
+  def full?
+    if !@cells.any? {|cell| cell == " "}
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+      return @player_1
+    else
+      return @player_2
+    end
+  end
+
+  def board
```
jilustrisimo: Take advantage of `attr_accessor ` macro for `:board, :player_1, :player_2`

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def player_1
+    @player_1
+  end
+
+  def player_2
+    @player_2
+  end
+
+  def over?
```
jilustrisimo: Make use of your other methods
```
def over?
  won? || draw?
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    if board.full? == true
+      true
+    else
+      false
+    end
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+
+
+  def draw?
+    if self.over? && self.won? == false
+      return true
+    end
+  end
+
+  def winner
+    if self.won? != false
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #709, opened 2017-10-11
Done

SallyZ12:


bin/tictactoe
```diff
@@ -1,3 +1,49 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+
+  class Human < Player
+
+    def move (board)
+      puts "Please Enter 1 - 9:"
+      board = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def initialize (player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #722, opened 2017-10-18
Project submitted

breadoliveoilsalt:


bin/tictactoe
```diff
@@ -1,3 +1,100 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
+class Board
+
+	attr_accessor :cells, :win_combinations
+
+	def initialize
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+	end
+
+
+	def reset!
+		@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+    	puts "-----------"
+    	puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+	end
+
+	def position(num)
+		cells[num.to_i-1]
+	end
+
+	def full?
+		!cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,49 @@
...
+	end
+
+	def turn_count
+		cells.select { |pos| pos == "X" || pos == "O" }.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def turn_count
+		board.cells.count { |square| square != " " }
+  end
+
+  def current_player
+		turn_count.even? ? player_1 : player_2
+  end
+
+	def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end

File: lib/game.rb [hunk 7]
```diff
@@ -0,0 +1,69 @@
...
+	def draw?
+		full? && !won?
+	end
+
+  def over?
+  	 won? || draw?
+ 	end
+
+	def winner
+		won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

scraps.rb
```diff
@@ -0,0 +1,51 @@
+This is a scraps file.  Does this work ok?
```
jilustrisimo: Not sure what this file does, but if it just a file for notes, might be best to add to the `.gitignore` file so it doesn't get added for code reviews and merges.

lib/game.rb
breadoliveoilsalt: Feel free to give a shout if anything I say here sounds off.  The comment here took me a while to figure out.  At first, I though this would result in an "unassigned variable" error on account of `won`.  But then I realized the comment is NOT saying `if won == won?`. That would produce the error.  Instead, `if won = won?` does two things:  1) it checks if `won?` returns something truthy.  2) If so, the return value of `won?` is assigned to `won`.  That's pretty slick.

breadoliveoilsalt:
Thanks for the comments!  Changes were made in a branch called refactoring and then merged into the master branch.

jilustrisimo:
Awesome job working in another branch then merging the code in!

PR #723, opened 2017-10-19
Project submitted

bshap93:


bin/tictactoe
```diff
@@ -1,3 +1,42 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+puts "Enter number of Players:"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+
+  attr_accessor :cells
+  def initialize(cells = nil)
+    self.cells = cells || Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,84 @@
...
+class Game
+  attr_accessor :board, :player_1, :player_2
+  WIN_COMBINATIONS = [[0,1,2], [3,4,5], [6,7,8], [0,3,6], [1,4,7], [2,5,8], [0,4,8], [6,4,2]]
+
+  def initialize(player_1=Player::Human.new("X"), player_2=Player::Human.new("O"), board=Board.new)
+    @player_1 = player_1; @player_2 = player_2; @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+class Human < Player
+
+  def move(game, board)
+    puts "Enter a number between 1 and 9"
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end

jilustrisimo (commented):
Hi Benjamin,
Great job on your TTT w/ AI project. I did a code review on Github and left some comments for you. You should have received some notifications from Github. Please notify me when you have completed the fixes or if you have any questions by email: janthony.ilustrisimo@wework.com or on slack: (@jilustrisimo).
Regards,
J.

PR #724, opened 2017-10-19
Done

RonaldVilorio:


bin/tictactoe
```diff
@@ -1,3 +1,88 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeController.new.call
```
jilustrisimo: Great job keeping your `bin` file small, but delete commented out code before submitting work.

lib/board.rb
```diff
@@ -0,0 +1,60 @@
+require 'pry'
+class Board
+  # Is this saving
+
+  attr_accessor :cells
+
+  def initialize(cells = Array.new(9," "))
+    @cells = cells
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  end
+
+  def reset!
+    cells.clear
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "------------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "------------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(user_input)
+    user_input = user_input.to_i
```
jilustrisimo: You can abstract this to 1 line `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  def full?
+    cells.all?{|cell| cell != " "}
+  end
+
+  def turn_count
+    cells.count{|token| token == "X" || token == "O"}
+  end
+
+  def taken?(user_input)
+    user_input = user_input.to_i
```
jilustrisimo: Make use of `Board#position`:
```
def taken?(input)
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+    cells[user_input-1] == "X" || cells[user_input-1] == "O"
+  end
+
+  def valid_move?(user_input)
+    user_input = user_input.to_i
+    !full? && !taken?(user_input) if user_input >= 1 && user_input < 10
```
jilustrisimo: You can simplify checking the valid integer range, and make it more semantic, by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+    @board = board if board
+  end
+
+  def current_player
+    #  if board.turn_count.even?
+    #    player_1
+    #  else
+    #    player_2
+    #  end
+     board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+    # checks each board cell index by placing in the index of the combination and comparing that to "X" or "O"
+    # On the first loop will check the index 0,1,2 of board.cells and then on the next iteration check the next combination
+    # until a match is found
+
+  end
+  def draw?
+    board.full? && won? == nil
+  end
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/player.rb
```diff
@@ -0,0 +1,10 @@
+class Player
+  def initialize(token)
+    @token = token
+  end
+
+  def token
```
jilustrisimo: You can use the `attr_accessor` macro

RonaldVilorio:
Hey thank you for taking the time to review my code. I have made most of
the fixes, but I have some questions before I resubmit.

1.Why use self.reset! inside the initialize method if when a new board is
created it will automatically take in the default cells which is set to an
a new array with 9 empty strings?

2.@board.turn_count.even? ? @player_1 : @player_2, could you clarify how to
read this syntax?

3. position.between?(1,9) is not working an throwing me an argument error
in the valid_move? method. Why is position not taking any arguments when
using the dot operator?

lib/board.rb
jilustrisimo: You'll soon learn about a programming concept called DRY (Don't repeat yourself).  In this case, recreating an empty board in 2 methods right next to each other might not seem like a big deal. However, with larger applications and when you are working in a team, you don't want to be using the same or similar code to accomplish the same thing in multiple places.  If all those methods need the same code for the app to function correctly and you or someone changes it in just in the places they remember that app will break.  It is best practice to define some functionality by itself then call it where you need it in your app.  If the app needs to change you only need to change the code in one place and everything will continue to function.

lib/game.rb
jilustrisimo: `even?` basically does the same as the modulo `i % 2 == 0`.  If the `turn_count` is `0,2,4...` then it's `player_1`'s turn and vice versa.  `.odd?` does the opposite.  It's to show you how many different ways you can accomplish the same thing in Ruby and it's less likely to cause a bug due to a typo.

jilustrisimo:
Hi Ronald!
I left additional comments under the parts you had questions for.  In the future, you can always comment under previous comments if they are related.  Let me know if you have any more questions!

lib/board.rb
RonaldVilorio: Hey I tried this code:
user_input.between?(1,9) 
but I couldn't pass for the input that were taken so I added an if statement
user_input.between?(1,9) if !taken?(user_input)
Is this alright or should I find another way?

lib/board.rb
RonaldVilorio: Alright so it's like saying "Don't put all your eggs in one basket", and one method does only one thing. I think I kind of understand better but I will learn more about it later.

lib/game.rb
RonaldVilorio: After @board.turn_count.even? the ? will check player 1 and player 2 for if it's even. Could I add a player 3 right after player 2 to check all 3 players?
@player_1 : @player_2 : @player_3

jilustrisimo: Think of a ternary expression the same as an `if... else...` statement. You and try and string multiple expressions (results) together but it'll only return 1 of 2 sets of data based on the condition being evaluated.

RonaldVilorio: Alright, I have submitted all the changes. Thank you for help.

PR #725, opened 2017-10-22
Done

dakaras:


bin/tictactoe
```diff
@@ -1,3 +1,19 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
+class Board
+  attr_accessor :cells, :display, :position, :turn_count, :update
+  def initialize
+    @cells = Array.new(9, " ")      #has an attribute to store the cells of the board
```

jilustrisimo: Remove unneeded comments before submitting work for review.

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(board_cell)                #takes in user input and returns the value of the board cell
+      board_cell = board_cell.to_i-1
+      @cells[board_cell]
```
jilustrisimo: You can abstract this to 1 line: `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def full?    #returns true for a full board, returns false for an in-progress game
```

This is also more efficient in that the loop will stop after the iteration for which a cell is not empty.  Also, try not to explicitly return a boolean from an if statement if not needed.

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def taken?(board_cell)  #returns true if the position is X or O
+    if position(board_cell)  == "X" || position(board_cell)  == "O"
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(board_cell)         #returns true for user input between 1-9 that is not taken
```
jilustrisimo: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```

File: lib/board.rb [hunk 6]
```diff
@@ -0,0 +1,66 @@
...
+    if board_cell.to_i.between?(1, 9) && taken?(board_cell) == false
+      true
+    else
+      false
+    end
+  end
+
+  def update(board_cell, player)      #updates the cells in the board with the player token according to the input
+    if valid_move?(board_cell) == true
```
jilustrisimo: You do not need to compare your conditional to a boolean with an if statement.  If the value doesn't return `nil` (which equals false) it will run the block.

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def winner
+    #returns X when X won
+    !won? ? nil : board.cells[won?.first]
+    #returns O when O won
+    #returns nil when no winner
+  end
+
+  def current_player #returns the correct player, X, for the third move
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+    [2,5,8],
+    [6,4,2],
+    [0,4,8]
+  ]
+
+  def over? #returns true for a draw or a won game
+    won? || draw?
+  end
+
+  def won?  #returns false for a draw
```
jilustrisimo: Good job using `.detect` but here you can abstract this to return the winning combination or `nil` if none is found.

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def draw?
+    if board.full? && !won?
+      true  #returns true for a draw, returns false for a won game or an in-progress game
+    end
+  end
+
+  def winner
+    #returns X when X won
+    !won? ? nil : board.cells[won?.first]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Human < Player  #inherits from Player
+
+    def move(board)
+      puts "Player '#{self.token}', please enter a valid move" #asks the user for input and returns it
+      your_move = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

dakaras: Hi J, thank you for looking over my project. I really appreciate the feedback. I saved the recommended changes you suggested and pushed them to Github. I want to caution that my code won't work completely as we discussed in Slack. Please let me know if there's anything I should change.

PR #727, opened 2017-10-23
Project submitted

scorrea83:


bin/tictactoe
```diff
@@ -1,3 +1,15 @@
...
+puts "   |   |   "
+puts "-----------"
+puts "   |   |   "
+puts "-----------"
+puts "   |   |   "
+
+game = Game.new
+puts "What kind of game would you like to play?"
+game.start
+game.end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  def position(user_input)
+    cells[user_input.to_i - 1]
+  end
+
+  def full?
+    cells.none? { |position| position == " "}
+  end
+
+  def turn_count
+    cells.select { |position| position != " "}.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,125 @@
...
+    self.board = board
+  end
+
+  def reset! #when called will reset Game attributes to initial state.
+    self.player_1 = Players::Human.new("X")
+    self.player_2 = Players::Human.new("O")
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,125 @@
...
+        end
+    end
+  end
+
+  def draw?
+    self.board.full? && !won?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,125 @@
...
+             # calls #play method to begin game play.
+    game_type = nil
+    puts "For a computer vs computer game type '0-player'"
+    puts "For a human vs computer game type '1-player'"
+    puts "For a human vs human game type '2-player'"
+    game_type = gets.strip
+
+    case game_type
+    when "0-player"
+      binding.pry
```
jilustrisimo: Remove breakpoints before submitting PRs for review.

lib/game.rb
scorrea83: cool, completely forgot about those two methods!, thanks!

lib/game.rb
scorrea83: sorry, forgot about those!

lib/board.rb
scorrea83: I had no idea .count took arguments! I guess I learned it a while back when only needing to return a number of elements in an array and never went back to look at it in the Docs! so useful!

PR #728, opened 2017-10-23
Done

Gevanstron:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Hello there, sunshine! Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
+
+  def initialize
+    self.reset!
```
jilustrisimo: Good job calling `reset!`

lib/board.rb
```diff
@@ -0,0 +1,59 @@
+require 'pry'
+
+class Board
+
+  attr_accessor :cells
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(user_input)
+    index = user_input.to_i - 1
+    cells[index]
```
jilustrisimo: You can refactor this to 1 line `cells[user_input.to_i - 1]`

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def full?
+    self.cells.none? {|cell| cell == " "}
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+    counter
+  end
+
+  def taken?(user_input)
+    if position(user_input) == "X" || position(user_input) == "O" then true else false
+    end
+  end
+
+  def valid_move?(user_input)
+    if user_input.to_i.between?(1,9) && !taken?(user_input) then true else false
```
jilustrisimo: You do not need an if statement here. `user_input.to_i.between?(1,9) && !taken?(user_input)` by itself will return a boolean depending on how the conditions evaluate.

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    if board.turn_count.even?
```
jilustrisimo: Good job using `.even?`.  You could use a ternary expression instead of an if statement and just wanting 1 of 2 expressions return based on a condition `board.turn_count.even? ? @player_1 : @player_2`

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+      player_1
+    else
+      player_2
+    end
+  end
+
+  def over?
+    if won?
```
jilustrisimo: You can refactor to `won? || draw?` to keep the code cleaner.

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+      win.all? {|w| board.cells[w] == "X"} || win.all? {|w| board.cells[w] == "O"}
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #729, opened 2017-10-23
Done

lsivak:


bin/tictactoe
```diff
@@ -1,3 +1,63 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+    reset!
+  end
+
+  def reset!
+    cells.clear
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]}  "
+    puts "-----------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]}  "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]}  "
+  end
+
+  def full?
+    cells.all? do |cell|
```
jilustrisimo: No need to use an if statement to return a boolean, the same could be accomplished with:
```
cells.all?{|token| token == "X" || token == "O"}
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+      if cell == "X" || cell == "O"
+        true
+      elsif cells.include?("")
+        false
+      end
+    end
+  end
+
+  def input_to_index(input)
+    #binding.pry
```
jilustrisimo: Remove commented out code before submitting PRs.

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+    input = input.to_i - 1
+  end
+
+  def position(input)
+    input = input_to_index(input)
```
jilustrisimo: You can abstract to `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+    cells[input]
+  end
+
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+      if cell != " " && count <= 9
+        count += 1
+      end
+    end
+    count
+  end
+
+  def valid_move?(input)
+
+    if input.to_i.between?(1,9) && position(input) == " "
```
jilustrisimo: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value. No need for an if statement.
```
```

lib/board.rb
```diff
@@ -0,0 +1,76 @@
...
+      true
+    end
+  end
+
+  def taken?(input)
+    if position(input) == "X" || position(input) == "O"
```
jilustrisimo: You can abstract to `!(position(input) == " " || position(input) == "")` or `position(input) == "X" || position(input) == "O"`

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+
+  end
+
+  def current_player
+
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+      player_2
+    end
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def draw?
+    if @board.full? && !won?
```
jilustrisimo: `@board.full? && !won?` will return a boolean by itself.

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def won?
+    WIN_COMBINATIONS.detect do |win|
+      @board.cells[win[0]] == "X" && @board.cells[win[1]] == "X" && @board.cells[win[2]] == "X"  && @board.taken?(win[0]+1)||
+      @board.cells[win[0]] == "O" && @board.cells[win[1]] == "O" && @board.cells[win[2]] == "O" && @board.taken?(win[0]+1)
+    end
+  end
+
+  def winner
+    won? ? @board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+    Players::Human
+    def move(board)
+      puts "Please enter a number from 1..9"
+      move = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lsivak:
Hi J.

I made the changes you suggested below. A few things though....

1. I don't see any messages on github (but maybe I'm not looking in the
correct place).
2. I tried to change my valid_move? to... input.to_i.between?(1,9) &&
!taken(input)... as you suggested, but it breaks the whole game, so I left
as is.
3. In define winner, I used, if won? instead of assigning won = won? Why do
I want to use won twice in the if statement instead of won? I didn't think
it mattered, but maybe it does.

Thanks so much for all of your suggestions!

Lara

PR #732, opened 2017-10-25
Done

zolkocarbon:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
+Game.start
```
jilustrisimo: Awesome job with keeping your bin file small!

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+
+class Board
+
+  attr_accessor :cells
+
+  DIM = 3
+  SIZE = DIM*DIM
+
+  def initialize
+    self.reset!
```
jilustrisimo: Awesome job calling `reset!` in this method.

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def reset!
+    @cells = %w(\ ) * SIZE
```
jilustrisimo: 👍  Interesting take on creating the array.

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    index = input.to_i-1
+    cells[index]
```
jilustrisimo: you can refactor this to 1 line `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+      element == "X" || element == "O"
+    end
+  end
+
+  def turn_count
+    @cells.count{|token| token == "X" || token == "O"}
+  end
+
+  def current_player
+    turn_count % 2 == 0 ? "X" : "O"
```
jilustrisimo: Use [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F)

Also, this method is technically game logic and not board logic.  Any particular reason you placed it here?

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def taken?(input)
+    #binding.pry
```
jilustrisimo: Delete commented out code, especially breakpoints, before submitting your code for review.

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+    if @cells[index] == " " || @cells[index] == "" || @cells[index] == nil
+      false
+    else
+      true
+    end
+  end
+
+  def valid_move?(input)
+    index = input.to_i-1
+    index.between?(0, 8) && !taken?(input)
```
jilustrisimo: You could just do `input.to_i.between?(1,9) && !taken?(input)` instead of saving it to a variable.

lib/game.rb
```diff
@@ -0,0 +1,137 @@
...
+    @player_2 = player_2
+    @board = board
+  end
+
+  def turn_count
+    board.cells.count{|token| token == "X" || token == "O"}
+  end
+
+  def current_player
+    turn_count % 2 == 0 ? player_1 : player_2
```
jilustrisimo: Please check comment in board.rb.

lib/game.rb
```diff
@@ -0,0 +1,137 @@
...
+    #if even, return player_1
+  end
+
+  def over?
+    #game is over if board is full or someone won
+    if draw? || won?
```
jilustrisimo: You don't need an if statement here.  `draw? || won?` will return true if either evaluate to `true`

lib/game.rb
```diff
@@ -0,0 +1,137 @@
...
+  def draw?
+    !won? && full?
+  end
+
+  def full?
+    board.cells.all? {|p| p == "X" || p == "O"}
+  end
+
+  def won?
+    WIN_COMBINATIONS.each do |win_set|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+      if board_value_1 == "O" && board_value_2 == "O" && board_value_3 == "O"
+        return win_set
+      end
+    end
+    return false
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cat's Game!"
+    end
+  end
+
+  def self.start
+    puts "Welcome to Tic Tac Toe!"
```
jilustrisimo: Prefer to encapsulate CLI logic into a class.

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board = nil)
+      puts "Player #{self.token}. Please enter 1-9:"
+      user_input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #733, opened 2017-10-25
Done

ghost:


lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def position cell_num
+       #need to convert to integer and correct for humans thinking 1-9, computer 0-8
+       idx = cell_num.to_i - 1
+       cells[idx]
+    end
+
+    def full?
+        !cells.detect{|pos| pos == " "}
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
...
+        cells.count {|pos| pos != " "}
+    end
+
+    def taken? pos
+        idx = pos.to_i - 1
+        cells[idx] != " "
+    end
+
+    def valid_move? pos
+        chosen = pos.to_i
```
jilustrisimo: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
```

lib/board.rb
```diff
@@ -0,0 +1,59 @@
+class Board
+    attr_accessor :cells
+
+    def initialize
+        self.cells = Array.new
```

lib/game.rb
```diff
@@ -0,0 +1,159 @@
...
+    ]
+
+    def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+        @player_1 = player_1
+        @player_2 = player_2
+        @board = board
+    end
+
+    def current_player
+        board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,159 @@
...
+
+        board.display
+        if won?
+            puts "Congratulations #{winner}!"
+        else
+            puts "Cat's Game!"
+        end
+    end
+
+    def start
```
jilustrisimo: Again, encapsulate all CLI logic into this own file.

lib/players/human.rb
```diff
@@ -0,0 +1,22 @@
+module Players
+    class Human < Player
+        def move board
```
jilustrisimo: Separation of Concerns dictates that CLI logic like how to display board should be in its own CLI class.  All this should do is ask for the input, then take that input and apply it to your game logic.

jilustrisimo (commented):
Hi Lawrence and Yomaira,
Great job on your TTT w/ AI project.  You both took used interesting ways to get this app working! I did a code review on Github and left some comments for you. You should have received some notifications from Github. I also suggest you take a look at the solution branch together (now that you have finished) see a look at another way to solve these problems.  It doesn't completely use the most effective code but it will enlighten you both on other ways to tackle these problems.  Please notify me when you have completed the fixes or if you have any questions by email: janthony.ilustrisimo@wework.com or on slack: (@jilustrisimo).
Regards,
J.

PR #734, opened 2017-10-26
Done

marsbhoward:


lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

PR #735, opened 2017-10-26
Done

Julee117:


lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+    if won?
+      board.display
+      puts "Congratulations #{winner}!"
+    else
+      board.display
+      puts "Cat's Game!"
+    end
+  end
+
+  def self.start
```
jilustrisimo: Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```
jilustrisimo: Good job keeping the bin file short!

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```
jilustrisimo: You can just call `reset!` here since you are already initializing a new array there.

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+  def current_player
+    board.turn_count.even? ? player_1 : player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def won?
+    WIN_COMBINATIONS.any? do |arr|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+  end
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def winner
+    return board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Where would you like to make your move(1-9)?"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #737, opened 2017-10-27
Done

mconway890:


lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def full?
+    cells.all? do |cell|
+      cell == "X" || cell == "O"
+    end
+  end
+
+  def turn_count
+    turns = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def input_to_index(input)
+    input.to_i - 1
+  end
+
+  def position(input)
+    input = input_to_index(input)
+    cells[input]
```
jilustrisimo: You can abstract to `cells[input.to_i-1]`

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+    cells.each do |cell|
+      if cell == "X" || cell == "O"
+        turns += 1
+      end
+    end
+    turns
+  end
+
+  def taken?(input)
+    if position(input) == "X" || position(input) == "O"
```
jilustrisimo: You can just check to see if the cell is not an empty string with `Board#position`

```
def taken?(input)
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def valid_move?(input)
+    if !taken?(input) && input.to_i.between?(1,9)
```
jilustrisimo: No need for an if statement. `!taken?(input) && input.to_i.between?(1,9)` wil return a boolean by itself.

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+       end
+     end
+  end
+
+  def draw?
+    !won? && @board.full?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

mconway890:
Hey there!

Went over all your suggestions/changes. Believe I fixed them all. Thanks
for your input and help!

-Michele Conway

PR #739, opened 2017-10-29
Done

kpeck71:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+TicTacToeController.new.call
```
jilustrisimo: Awesome job with keeping a small bin!

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells, :player
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  #   index = input.to_i - 1
+  # end
+
+  def taken?(input)
+    index = input.to_i - 1
+    !(self.cells[index] == " ")
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def won?
+     WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/board.rb
kpeck71: Ah yes that is so much better

PR #800, opened 2017-12-07
Done

dbfarms:


lib/board.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def position(input)
+    input = input.to_i - 1
+    #binding.pry
+    return "#{@cells[input]}"
+  end
+
+  def full?
+    @cells.detect {|space| space == " "} ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def turn_count
+    turn_count = Array.new
+    turn_count = @cells.find_all {|space| space == "X" || space == "O"}
+    return turn_count.size
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def taken?(position)
+    @cells[(position.to_i - 1)] == " " ? false : true
+  end
+
+  def valid_move?(input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def player_1
+    @player_1
+  end
+
+  def player_2
+    @player_2
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+    WIN_COMBINATIONS.each do |combo|
+      if @board.cells[combo[0]] == @board.cells[combo[1]]  &&  @board.cells[combo[1]] == @board.cells[combo[2]]
+        return false
+      end
+    end
+    return true
+=end
+  end
+
+  def winner
```
cernanb: You can use the `won?` method here 
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "Enter position to play: "
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #801, opened 2017-12-07
Done

mietone:


lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.none? { |x| x == " "}
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  def turn_count
+    9 - @cells.count(" ")
+  end
+
+  def taken?(input)
+    position(input) != " "
+  end
+
+  def valid_move?(input)
+    position(input) == " " && input.to_i.between?(1,9)
```
cernanb: You can take advantage of the `taken?` method here 
```
```
`

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Player_#{token}: Make your move 1-9"
```
```
def move(board)
  gets.chomp
end
```

PR #804, opened 2017-12-11
Done

nourski:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+attr_accessor :cells
+  def initialize
+    @cells = []
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    reset!
+  end
+
+  def reset!
+    @cells.clear
```
cernanb: You don't need to do both .clear and create a new array in this method.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  def position(position)
+    @cells[position.to_i - 1]
+  end
+
+  def full?
+    @cells.include?(" " || "") == false
+  end
+
+  def turn_count
+    @cells.select {|i| i == "X" || i == "O"}.length
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def taken?(position)
+    position(position) == "X" || position(position) == "O"
+  end
+
+  def valid_move?(position)
+    position.to_i < 10 && position.to_i > 0 && taken?(position) == false
```
cernanb: You can use the `.between?` method and the negation operation in this method here. 

`position.to_i.between?(1, 9) && !taken?(position)`

bin/tictactoe
```diff
@@ -1,3 +1,40 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def winner
+    won? ? @board.cells[won?[0]] : nil
+  end
+
+  def won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #986, opened 2018-03-15
Done

denisepen:


bin/tictactoe
```diff
@@ -1,3 +1,18 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    reset!
```
cernanb: Great use of your reset method here!

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
```
cernanb: You could refactor this method to use the implicit return value. 

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+  @board = board
+  @player_1 = player_1
+  @player_2 = player_2
+end
+
+def board
+  @board
+end
+
+def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+def draw?
+  @board.full? && !won?
+end
+
+def over?
+  won? || draw?
+end
+
+def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter your selection (1-9)"
+      input =gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #991, opened 2018-03-20
Done

MarsRoamer:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+
+  def initialize
+      reset!
+    end
+
+  def self.cells
+    @cells
```
kkterai: With `attr_accessor :cells`, this method is not needed, and it looks like it is not actually used in your code. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def position(user_input)
+    @cells[user_input.to_i - 1]
+  end
+
+  def full?
+    @cells.none? {|n| n == ' '}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

MarsRoamer:
Thanks, Kenlyn! I appreciate it!  I'll go through and do some refactoring.

PR #992, opened 2018-03-20
Done

z00kie:


bin/tictactoe
```diff
@@ -1,3 +1,53 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

The code in your bin would generally just go in a `call` method.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = []
+    reset!
+  end
+
+  def cells
```
kkterai: With `attr_accessor :cells`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+      board.cells[combo[0]] == board.cells[combo[1]] &&
+      board.cells[combo[1]] == board.cells[combo[2]]
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"), board=Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```
kkterai: Consider using [`Integer#even?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-even-3F) or [`Integer#odd?`](http://ruby-doc.org/core-1.8.7/Integer.html#method-i-odd-3F) for greater expressiveness.

kkterai:
Hi Emmy,

Great job on your Tic Tac Toe with AI Project! I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can contact me by email: kenlyn@flatironschool.com or on slack: @kkterai.

Regards,

Kenlyn Terai

PR #993, opened 2018-03-20
Done

esugano:


bin/tictactoe
```diff
@@ -1,3 +1,31 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells, :the_player
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+  
+  def full?
+    @cells.include?(" ") ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+  
+  def turn_count
+    turns = 0
```

```ruby
def turn_count
end
```

lib/board.rb
esugano: good idea! thanks

PR #994, opened 2018-03-22
Project submitted

jpkim921:


bin/tictactoe
```diff
@@ -1,3 +1,90 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,85 @@
+require 'pry'
+class Board
+  attr_accessor :cells, :free_corner
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  # position method already has #idx built in so dont need to call #idx again.
+  def position(input)
+    cells[self.idx(input)]
+  end
+
+
+  def full?
+    cells.all?{|cell| cell != " "}
```
cernanb: Excellent use of the `all?` method here.

lib/board.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def turn_count
+    cells.select{|cell| cell != " "}.count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def taken?(input)
+    position(input) == " " ? false : true
+  end
+
+  def valid_move?(input)
+    if (0..8).include?(self.idx(input)) && self.taken?(input) == false
```
cernanb: You should be able to use the implicit return values rather than check for equality. 

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+      @board = board
+      # binding.pry
+    end
+
+    # current_player either returns player 1 or player 2
+    def current_player
+      self.board.turn_count.odd? ? self.player_2 : self.player_1
+    end
+
+    def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+module Players
+
+  class Human < Player
+    attr_accessor :input
+
+    def move(board)
+      puts "Enter move:"
+      @input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #996, opened 2018-03-26
Done

AspenJames:


lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "-----------"
+    puts " #{self.cells[6]} | #{self.cells[7]} | #{self.cells[8]} "
+  end
+
+  def position(input)
+    self.cells[input.to_i - 1]
+  end
+
+  def full?
+    self.cells.include?(" ") ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def turn_count
+    self.cells.select{|c| c == "X" || c == "O"}.count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def taken?(input)
+    (self.cells[input.to_i - 1] == "X" || self.cells[input.to_i - 1] == "O") ? true : false
+  end
+
+  def valid_move?(input)
+    ((1..9).to_a.include?(input.to_i) && !taken?(input.to_i)) ? true : false
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"),
+    player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def won?
+    won = false
+    cells = self.board.cells
+    WIN_COMBINATIONS.each do |arr|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  def draw?
+    if self.board.full? && !self.won?
+      true
+    else
+      false
+    end
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      self.board.cells[won?[0]]
+    else
+      nil
```
weezwo: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
  @board.cells[win_combo[0]]
end
```

weezwo (commented):
Nice job!! Left a few comments for you. There's no need to submit this project again -- these are just a few tips for the future.

AspenJames:
Thank you for the feedback!
I'll take a deeper dive into this soon!

Aspen

*Aspen James*
*They/Them/Theirs*
(206)290-2148

PR #997, opened 2018-03-27
Done

cloudycua:


bin/tictactoe
```diff
@@ -1,3 +1,134 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,57 @@
...
+    WIN_COMBINATIONS.detect do |combo|
+      board.cells[combo[0]] == board.cells[combo[1]] && board.cells[combo[1]] == board.cells[combo[2]] && board.cells[combo[0]] != " "
+    end
+  end
+
+  def draw?
+    !board.cells.any? {|c| c == " "} && !self.won?
+  end
+
+  def winner
```
kkterai: Nice job using only one call to `Game#won?` once!

PR #998, opened 2018-03-27
Done

arye-dov-eidelman:


lib/CLI.rb
```diff
@@ -0,0 +1,34 @@
+class CLI
+  def initialize
```
weezwo: Typically this method would be called something like `call` and your bin would call on `CLI.new.call`.

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(user_input)
+    @cells[user_input.to_i-1]
+  end
+
+  def full?
+    !@cells.include?(' ')
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def turn_count
+    @cells.find_all{ |cell| cell != ' '}.length
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def taken?(user_input)
+    position(user_input) != ' '
+  end
+
+  def valid_move?(user_input)
+    !taken?(user_input) && (1..9).include?(user_input.to_i)
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+    # over? && !won?
+  end
+
+  def winner
+    won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1000, opened 2018-03-28
Done

Dee-Race:


bin/tictactoe
```diff
@@ -1,3 +1,37 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
+    cells.all?{ |character| character == "X" || character == "O"}
```
weezwo: Nice job using the `all?` iterator. You could make this even more DRY by checking that all spaces are not empty (rather than whether they contain an X or O) — something like this:

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

PR #1002, opened 2018-03-29
Done

zep1994:


bin/tictactoe
```diff
@@ -1,3 +1,9 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def turn
+    player = current_player
+    current_move = player.move(@board)
+    if !@board.valid_move?(current_move)
+      turn
+    else
+      puts "Turn: #{@board.turn_count+1}\n"
```
weezwo: I am concerned by the striking similarity between your code and the code in the solution branch.

lib/game.rb
```diff
@@ -0,0 +1,70 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

zep1994:
Hello,

I did not copy and paste from the solution. I remember this one being
challenging and leading me all over the internet. I looked over a couple
blog posts about the initialization and changing the spec file, but I went
through this method after method. I did not code the methods in that order,
I changed the order to an order someone had talked about in a blog/post,
but I only passed this test when I changed that spec file and did not think
the order was the issue to begin with, but to make sure I ran the test and
left them in the order. I cannot remember exactly, but I spent about two
hours making changes to that code and I only added files, because with all
of the information, it was a lot and so I went with what other previous
labs had done. Much of the logic was similar to the previous labs and
readmes. There were things I needed help on, but I did not take someone
else's work. I spent hours going over this lab. I was in a study group
before I did this lab on this subject and knew I would need all the help I
could get. I did my work though. I did not go copy and paste to be done. I
was up all that night working on this stuff and was deeply engaged.
Flatiron means the world to me and I would not just go copy and paste to
get done. I am trying to start a company, not just get a job like most and
I am aware that I need to engaged in the material. I have already emailed
my coach to tell her I am going to slow down to do what Avi said in one of
his videos, "take your time". I take notes on much of what I hear and read
and I remember this one really pushing me, but I did not take someone else
code. I stayed up two nights in a row learning about OO and taking notes. I
slept for two days after. That is my logic that I learned and have learned.
I still have questions, because I have a lot, but I did not skip any
portion of this lab. I am sorry for the long email, but this is deeply
concerning, because I love this school more than anything and want to make
the most out of it. Not copy and paste.

Thomas Matlock

bin/tictactoe
zep1994: Makes sense, this reminded me of first working in CLI, when a lab talked about using the bin folder for any code dealing with the actual running of the program and since it "calls" it, use the bin folder.

lib/game.rb
zep1994: I should have thought of that. I was looking at OO TTT Lab from Procedural Ruby so I was thinking modulus, because it gave me such a headache at first, but that should've been my first sign  .even? or .odd? was a better refactor.

PR #1003, opened 2018-03-29
Project submitted

icvlahovic:


bin/tictactoe
```diff
@@ -1,3 +1,148 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,148 @@
...
+def coin_flip
```
weezwo: `coin_flip` is a nice touch. :)

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    cells.count("X") + cells.count("O") == 9
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+      board.taken?(win_combination[0] + 1)
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def over?
+    (won? || draw?) ? true : false
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def winner
+    won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      puts "Please enter 1-9 to make your move:"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,122 @@
...
+        board.cells[win_combination[1]] == token &&
+        board.cells[win_combination[2]] == " ") ||
+        (board.cells[win_combination[0]] == token &&
+        board.cells[win_combination[1]] == " " &&
+        board.cells[win_combination[2]] == token) ||
+        (board.cells[win_combination[0]] == " " &&
+        board.cells[win_combination[1]] == token &&
+        board.cells[win_combination[2]] == token)
+      end
+      winning_cell = three_in_a_row.find { |cell| board.cells[cell] == " " } if three_in_a_row != nil
```
weezwo: You do not need to compare to nil -- rely on the implicit return value of the expression. For example:

```
winning_cell = three_in_a_row.find { |cell| board.cells[cell] == " " } if three_in_a_row
```

PR #1004, opened 2018-03-29
Done

MisterSenor:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,68 @@
+require 'pry'
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def position(input)
+    board_position = input.to_i - 1
+    @cells[board_position]
+  end
+
+  def full?
+    @cells.none? do |check|
+      if check == " "
+        true
```
weezwo: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression:

`cells.none? {|cell| cell == " "}`

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+    end
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+      if @cells[index] == " " || @cells[index] == "" || @cells[index] == nil
+        false
+      elsif @cells[index] == "X" || @cells[index] == "O"
+        true
+      end
+    end
+
+    def valid_move?(input)
+      index = input.to_i - 1
+      if taken?(input) == false && index.between?(0,8)
```
Reverse (or negate) the boolean with the `!` (bang) operator.

```
 def valid_move?(input)
  index = input.to_i - 1
  !taken?(input)  && index.between?(0,8)
end
```

You may also want to consider creating an `input_to_index` method to take care of the conversion.

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def initialize(player_1 = nil, player_2 = nil, board = nil)
+    @player_1 = player_1 || Players::Human.new("X")
+    @player_2 = player_2 || Players::Human.new("O")
+    @board = board || Board.new
+    #binding.pry
+  end
+
+  def current_player
+    if @board.turn_count.odd?
```
weezwo: This definitely works, but it's also a great place to use a ternary operator!

`@board.turn_count.odd? ? @player_2 : @player_1`
Or:

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+  def over?
+    if draw?
+      true
+    elsif won?
+      true
+    end
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+      @player_2
+    else
+      @player_1
+    end
+  end
+
+
+  def won?
+    WIN_COMBINATIONS.find do |combo|
+     if @board.cells[combo[0]] == "X" && @board.cells[combo[1]] == "X" && @board.cells[combo[2]] == "X"
```
weezwo: Rather than repeating this logic for X and O, try something like this:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+
+module Players
+
+  class Human < Player
+
+    def move(input)
+      input = gets.chomp
```
```
def move(board)
  gets.chomp
end
```

PR #1006, opened 2018-03-30
Done

Kiveo:


bin/tictactoe
```diff
@@ -1,3 +1,75 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Hello and welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+class Board
+  attr_accessor :cells
+
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    @cells[spot.to_i-1] == "X" || @cells[spot.to_i-1] == "O"
+  end
+
+  def valid_move?(spot)
+    !taken?(spot) && spot.to_i.between?(1,9)
+  end
+
+  def update(spot, current_player)
+    if valid_move?(spot)
+      if turn_count%2 == 0
```
weezwo: Is this conditional necessary? Is there any difference in which code is executed?

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def position(user_input)
+    @cells[user_input.to_i-1]
+  end
+
+  def full?
+    @cells.all? {|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count  #how many x's and O's, will be the turns done.
+    @cells.count("X") + @cells.count("O")
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    board.turn_count%2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def over?
+    board.full? #|| won?
+  end
+
+  def won?
+    win_combo = false #default if there are no win_combos
+    # board.cells[x] compares to win_combos , and if win_combos[z] all == "x", then return the combo
+    WIN_COMBINATIONS.each do |numarray|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+      if @board.cells[numarray[0]] =="X" && @board.cells[numarray[1]] == "X" && @board.cells[numarray[2]] == "X" || @board.cells[numarray[0]] == "O" && @board.cells[numarray[1]] == "O" && @board.cells[numarray[2]] == "O"
+        win_combo = numarray
+      end
+    end
+    win_combo #default false or win_combo if found
+  end
+
+  def draw?
+    won? == false && @board.full? == true
```
weezwo: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def winner
+    winner = nil
+    WIN_COMBINATIONS.each do |numarray|
```
weezwo: Use your `won?` method here, rather than repeating this logic.

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+    def move(input)
```
```
def move(board)
  gets.chomp
end
```

PR #1007, opened 2018-04-02
Done

Naudria:


lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9," ")
```
kkterai: Since this utilizes the same code as the `reset!` method, prefer to call the `reset!` method here.

PR #1008, opened 2018-04-03
Done

mnimmz:


bin/tictactoe
```diff
@@ -1,3 +1,61 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

PR #1010, opened 2018-04-03
Done

kristinreddington:


bin/tictactoe
```diff
@@ -1,3 +1,32 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
+class Board
+  attr_accessor :cells, :turn_count
+  #@cells = []
+def initialize
+  @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+end
+
+def cells
+  @cells
+end
+
+def reset!
+  self.cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
weezwo: Prefer `Array.new(9, " ")` as used above -- It's more concise and will prevent errors from repeatedly typing `" "`.

lib/board.rb
```diff
@@ -0,0 +1,74 @@
+class Board
+  attr_accessor :cells, :turn_count
+  #@cells = []
+def initialize
+  @cells = Array.new(9, " ")
+end
+
+def cells
+  @cells
```
weezwo: With `attr_accessor :cells`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+#end
+
+def position(input)
+  position = input.to_i - 1
+  @cells[position]
+end
+
+  def full?
+    @cells.none? do |space|
+      if space == " "
```
weezwo: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. This method's content could be `cells.none? {|cell| cell == " "}`

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+        true
+      else
+        false
+      end #ends if/else
+  end #ends block code
+end #ends full? method
+
+def turn_count
+  @cells.count {|c| c == "X" || c == "O"}
```
weezwo: Nicely done!!

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+end
+
+def taken?(move)
+  if @cells[move.to_i - 1] == "X" || @cells[move.to_i - 1] == "O"
+    true
```
weezwo: This is another chance to avoid returning boolean literals.

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  else
+    false
+  end
+end
+
+  def valid_move?(move)
+    move_taken = move.to_i
+    if move_taken.between?(1, 9) && taken?(move_taken) == false
+     true
```
weezwo: See above. :)

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+]
+
+def initialize(player_1 = Players::Human.new("X") , player_2 = Players::Human.new("O"), board = Board.new)
+  @player_1 = player_1
+  @player_2 = player_2
+  @board = board
+end
+
+def current_player
+ return @player_1 if board.turn_count.even?
```
weezwo: This might be a good place to use a ternary operator. `@board.turn_count.even? ? @player_1 : @player_2` Avoid the return keyword!

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+  def draw?
+  @board.full? && !won?
+  end
+
+  def over?
+    @board.full? || won? || draw?
+  end
+
+  def winner
+    if WIN_COMBINATIONS.any? {|array| array.all? {|cell| self.board.cells[cell] == "X"}}
```
weezwo: Consider using an if statement assignment. For example:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+          @board.cells[win_combination[1]] == @board.cells[win_combination[2]] &&
+          @board.taken?(win_combination[0]+1)}
+  end
+
+  def draw?
+  @board.full? && !won?
+  end
+
+  def over?
+    @board.full? || won? || draw?
```
weezwo: Do we need to call on both `@board.full?` and `draw?` here? Which would be better?

PR #1011, opened 2018-04-03
Done

ruuuh20:


lib/board.rb
```diff
@@ -0,0 +1,3 @@
+class Board
```
kkterai: Good start! Keep going and let us know when complete!

PR #1014, opened 2018-04-05
Done

AAM77:


lib/board.rb
```diff
@@ -0,0 +1,69 @@
+class Board
+
+  attr_accessor :cells
+
+
+  def initialize
+    self.cells ||= Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  ## determines the value in the input position
+  def position(user_input)
+    cells[user_input.to_i - 1]
+  end # position
+
+
+  ## determines if the board is full
+  def full?
+    true if cells.all?{|pos| pos == "X" || pos == "O"}
```
weezwo: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression. This method's content could read: `cells.all?{|pos| pos == "X" || pos == "O"}`

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  end # full?
+
+
+  ## determines what turn it is now
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  ## determines if the spot has been taken
+  def taken?(user_input)
+    position(user_input) == "X" || position(user_input) == "O" ? true : false
+  end # taken?
+
+
+  ## determines if the entered input is of the specified parameters
+  def valid_move?(user_input)
+    index = user_input.to_i - 1
+    true if (0 <= index && index <= 8) && taken?(user_input) == false
```
weezwo: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  def turn_count
+    count = 0
+    cells.each {|pos| count += 1 if pos == "X" || pos == "O"}
+    count
+  end # turn_count
+
+
+  ## determines if the spot has been taken
+  def taken?(user_input)
+    position(user_input) == "X" || position(user_input) == "O" ? true : false
```

lib/game.rb
```diff
@@ -0,0 +1,113 @@
...
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end # initialize
+
+
+  ## determines whose turn it is
+  def current_player
+    if board.turn_count.even?
```
weezwo: This looks good, but would also be the perfect opportunity to use a ternary expression! `@board.turn_count.even? ? @player_1 : @player_2`

AAM77:
Thank you for all of the great comments and suggestions!
I look forward to going back and changing the code to see how it works out. :)

PR #1016, opened 2018-04-06
Done

zep1994:


zep1994:
Not sure if I am finished? maybe, I should speak with a coach about the objectives for this. I know we are supposed to commit on Git for the Data Gem, but what about this and student scraper and music cli?

PR #1018, opened 2018-04-10
Done

micholi:


lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+  def update(input, player)
+    cells[input.to_i-1] = player.token
+  end
+
+  def full?
+    cells.all?{|position| position == "X" || position == "O"}
+  end
+
+  def turn_count
+    cells.count("X") + cells.count("O")
```
weezwo: This can also be accomplished with one call to `#count`. See below:
```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+        end
+    end
+  end
+
+  def draw?
+    board.full? && !won?
+  end
+
+  def winner
+    if won? && board.cells[won?[0]] == "X"
```

```ruby
if won = won?
  board.cells[won.first]
end
```

micholi: Thanks so much for your feedback and tips on how to improve my code! This is super helpful!

PR #1020, opened 2018-04-10
Done

DGMarkel:


lib/game_logic.rb
```diff
@@ -0,0 +1,89 @@
+class TicTacToe
```
weezwo: Nice job encapsulating this logic within a CLI class!! You'll usually want to give this file the same name as the class (except in snake case) and will likely want to name the class itself to indicate its role as the CLI.

lib/game_logic.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def initialize
+    welcome
```
weezwo: A more typical pattern here would be to simply call `TicTacToe.new.welcome` in the bin file.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board
+attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  def position(user_input)
+    @cells[user_input.to_i - 1]
+  end
+
+  def full?
+    @cells.all? {|cell| cell != " "}
+  end
+
+  def turn_count
+    counter = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+    counter
+  end
+
+  def taken?(user_input)
+    self.position(user_input) == "X" || self.position(user_input) == "O"
+  end
+
+  def valid_move?(user_input)
+    !taken?(user_input) && user_input.to_i > 0 && user_input.to_i <= 9
```
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    if board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+  def won?
+    WIN_COMBINATIONS.detect do |win_combination|
+      board.cells[win_combination[0]] == board.cells[win_combination[1]] &&
+      board.cells[win_combination[0]] == board.cells[win_combination[2]] &&
+      board.cells[win_combination[0]] != " "
+    end
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      input = gets.strip
+      input
```
```
def move(board)
  gets.chomp
end
```

PR #1022, opened 2018-04-11
Done

Awilmerding1:


lib/game.rb
```diff
@@ -0,0 +1,206 @@
+require "pry"
+
+class Game
```
kkterai: We would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```
kkterai: Short length of code is ideal for the bin file. However, prefer to encapsulate CLI logic into a class (it is currently in Game.rb). For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} \n\n"
+  end
+  
+  def reset! 
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
+require "pry"
+class Board
+  attr_accessor :cells
+  attr_reader :display
+  
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+  def reset! 
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+  
+  def position(user_input)
+    self.cells[user_input.to_i-1]
+  end
+  
+  def full?
+    if self.cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+      false 
+    else
+      true
+    end
+  end
+  
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+  def taken?(input)
+    if self.cells[input.to_i-1] != " "
+      true 
+    else 
+      false 
+    end
+  end
+  
+  def valid_move?(input)
+    if self.taken?(input.to_i) == false && input.to_i.between?(1,9)
```
kkterai: Good job using the #between? method!

lib/game.rb
```diff
@@ -0,0 +1,206 @@
...
+  def current_player
+    if self.board.turn_count.even? == true 
+      self.player_1
+    else 
+      self.player_2
+    end
+  end
+  
+  def won?
+    WIN_COMBINATIONS.any? do |array|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

PR #1023, opened 2018-04-12
Project submitted

vanetoro:


lib/board.rb
```diff
@@ -0,0 +1,67 @@
+  class Board
+    attr_accessor :cells, :token
+  
+  def initialize
+    @cells = [' ', ' ', ' ', ' ', ' ', ' ',' ', ' ',' ']
+  end
+  
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  
+  def full?
+    if @cells.find { |i| i == ' '}
+      false
+    else
+      true
+    end 
+  end  
+  
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+    turn.odd? ?  player_2 :  player_1
+  end 
+  
+  def over?
+    won? || draw?
+  end  
+  
+  def won?
+     answer = WIN_COMBINATIONS.find do |combo|
```
kkterai: Nice job using #find!

PR #1024, opened 2018-04-12
Project submitted

treverandersen:


lib/board.rb
```diff
@@ -0,0 +1,41 @@
...
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} \n-----------\n #{cells[3]} | #{cells[4]} | #{cells[5]} \n-----------\n #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
+    cells.all?{|i| i == "X" || i == "O"}
```
kkterai: Good job using the most expressive and efficient iterator for this problem, #all?

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+      @player_2 = Players::Computer.new("O")
+    when "2"
+      @player_1 = Players::Human.new("X")
+      @player_2 = Players::Human.new("O")
+    end
+  end
+
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    tokens = combo.map {|cell| @board.cells[cell]}
+    tokens.uniq.size == 1 && tokens.none? { |token| token == " " }
+   end
+
+   if matches.size > 1
+     @board.cells[matches[0][0]] == player_1.token ? matches[0] : matches[1]
+   else
+     matches[0]
+   end
+ end
```
kkterai: Here, another option is to use [`Enumerable#detect`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect) to return the winning combination or `nil` if none is found.

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,52 @@
...
+    	if winning_square
+    		square = winning_square
+    	elsif board.cells[4] == " "
+    		square = "5"
+    	else
+    		until board.valid_move?(square)
+      			square = (1..9).to_a.sample
+      		end
+      	end
+      	square.to_s
```
kkterai: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression. For example:
```
def move(board)
  gets.chomp
end
```

PR #1025, opened 2018-04-12
Done

TrevorSeitz:


lib/board.rb
```diff
@@ -0,0 +1,63 @@
+class Board
+  attr_accessor :cells, :turn_count, :WIN_COMBINATIONS
+  def initialize
+    @cells = cells
+    reset!
+  end
+
+  def cells
```
kkterai: With `attr_accessor :cells`, this method is not needed. `attr_accessor` is a macro that automatically adds both the setter method and the getter method to your class for the selected attribute. In this case, both
```
def cells
  @cells
end
```
and
```
def cells=(cells)
  @cells = cells
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    @cells.none? do |space|
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+      if space == " "
```
kkterai: Instead, rely on the implicit boolean return value of the expression.

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+        true
+      else
+        false
+      end #ends if/else
+    end #ends block code
+  end #ends full? method
+
+  def turn_count
+    @cells.count {|c| c == "X" || c == "O"}
```
kkterai: Good job using #count!

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def taken?(move)
+    @cells[move.to_i - 1] == "X" || @cells[move.to_i - 1] == "O"
+  end
+
+  def valid_move?(move)
+    move_taken = move.to_i
```
```
move.to_i.between?(1,9) && !taken(input)
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+      board = Board.new
+      new_game = Game.new(player_1, player_2, board)
+      play
+    end
+  end
+end
+end
+
+def current_player
+  board.turn_count.even? ? @player_1 : @player_2
```
kkterai: Good job using the most semantic method, #even?

PR #1026, opened 2018-04-12
Done

jesshasquirks:


bin/tictactoe
```diff
@@ -1,3 +1,53 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def reset!
+    cells.clear
+    self.cells = Array.new(9, " ")
```
kkterai: You can refactor this to `@cells = Array.new(9, " ")`

lib/board.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def position(user_input)
+    cells[user_input.to_i - 1]
+  end
+
+  def full?
+    cells.none? { |cell| cell == " "}
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,99 @@
...
+    [6,7,8], # Bottom row
+    [0,3,6],
+    [1,4,7],
+    [2,5,8],
+    [0,4,8],
+    [2,4,6]
+  ]
+
+  def current_player
+    if board.turn_count.even?
```
kkterai: Nice job using `even?`. You can also refactor to use a ternary operator: `@board.turn_count.even? ? @player_1 : @player_2`

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+    def move(token)
```
```
def move(board)
  gets.chomp
end
```

PR #1027, opened 2018-04-13
Done

Perchitti:


bin/tictactoe
```diff
@@ -1,3 +1,23 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+require 'pry'
+
+class Board
+
+
+attr_accessor :cells
+
+
+def initialize
+  @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+def won?
+    WIN_COMBINATIONS.detect do |combo|
+      @board.cells[combo[0]] == @board.cells[combo[1]] &&
+      @board.cells[combo[0]] == @board.cells[combo[2]] &&
+      @board.taken?(combo[0] + 1)
+    end
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+
+def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+  @board = board
+  @player_1 = player_1
+  @player_2 = player_2
+end
+
+
+def won?
+    WIN_COMBINATIONS.detect do |combo|
```
kkterai: Nice job using `#detect` here!

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+  class Human < Player
+
+def move(input)
+  puts "Please enter 1-9:"
+  gets.strip
```
kkterai: Nice job using implicit returns!

Perchitti:
Thank you Kenlyn

PR #1028, opened 2018-04-13
Done

gitluky:


bin/tictactoe
```diff
@@ -1,3 +1,54 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def display
+    puts " #{cells[0]} | #{cells[1]} | #{cells[2]} "
+    puts "-----------------------------------------"
+    puts " #{cells[3]} | #{cells[4]} | #{cells[5]} "
+    puts "-----------------------------------------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def initialize(player_1=Players::Human.new("X"), player_2=Players::Human.new("O"),board=Board.new)
+    self.board = board
+    self.player_1 = player_1
+    self.player_2 = player_2
+  end
+
+  def current_player
+    number_of_turns = self.board.cells.select {|turns| turns != " "}.count
+    if number_of_turns.even?
```
kkterai: Nice job using the most semantic method possible here, `#even?`

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+      player_1
+    else
+      player_2
+    end
+  end
+
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    elsif possible_winning_combination.count > 1
+      possible_winning_combination.select do |combo|
+        board.cells.min_by {|token| board.cells.count(token)} == board.cells[combo[0]]
+      end.flatten
+    else
+      possible_winning_combination.flatten
+    end
+  end
+
+  def winner
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,97 @@
...
+    #if you can win, return the winning spot, if you are not and Your opponent is going to win, return that spot so you can block them.
+    def win_or_block(board, token)
+      if token == self.token
+        winning_combinations = find_available_combinations(board, self.token).select {|combo| (combo & find_moves_made(board, self.token)).count == 2 }
+        winning_combination = winning_combinations.sample
+        (winning_combination - find_moves_made(board, self.token)).first
+      else
+        winning_combinations = find_available_combinations(board, self.opponent.token).select {|combo| (combo & find_moves_made(board, self.opponent.token)).count == 2 }
+        winning_combination = winning_combinations.sample
+        (winning_combination - find_moves_made(board, self.opponent.token)).first
```
kkterai: Good job using implicit returns here!

PR #1029, opened 2018-04-13
Done

Cl3arglass:


bin/tictactoe
```diff
@@ -1,3 +1,13 @@
...
+
+
+
+# puts "Welcome to Tic Tac Toe!"
+# puts "How many human players?"
+# puts "0, 1, or 2?"
+# game = Game.new(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+# game.play
+
+Game.new.start
```
kkterai: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def reset!
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+   puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+   puts "-----------"
+   puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(user_input)
+    @cells[user_input.to_i - 1]
+  end
+
+  def full?
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    !@cells.include?(" ")
+  end
+
+  def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
...
+    self.cells.reject {|cell| cell == " "}.size
+  end
+
+  def taken?(user_input)
+    # position(user_input) == "X" || position(user_input) == "O"
+    position(user_input) != " "
+  end
+
+  def valid_move?(user_input)
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    # @board.cells.collect{|space| space == "X"}.size.odd? || @board.cells == @board.reset! ? @player_1 : @player_2
+    @board.turn_count.even? ? @player_1 : @player_2
```
kkterai: Good job using the most semantic method possible here, `#even?`

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+  # def won?
+  #   # Why is X the winner?
+  #   WIN_COMBINATIONS.any? do |combo|
+  #     if @board.taken?(combo[0]) && @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]]
+  #       return combo
+  #     end
+  #   end
+  # end
+  def won?
+
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+    position_2 = @board.cells[win_index_2] # load the value of the board at win_index_2
+    position_3 = @board.cells[win_index_3] # load the value of the board at win_index_3
+
+    if position_1 == "X" && position_2 == "X" && position_3 == "X"
+      return win_combination # return the win_combination indexes that won.
+    elsif position_1 == "O" && position_2 == "O" && position_3 == "O"
+      return win_combination
+      end
+   end
+   false
```
kkterai: Be sure to use the implicit returns of your methods.

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  def draw?
+    @board.full? && !self.won?
+  end
+
+  def winner
+    if self.won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1031, opened 2018-04-15
Done

urnotjessie:


bin/tictactoe
```diff
@@ -1,3 +1,11 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+user_choice = "y"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+      self.turn
+    end
+    if self.draw?
+      puts "Cat's Game!"
+    elsif self.won?
+      puts "Congratulations #{self.winner}!"
+    end
+  end
+
+  def start
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+require 'pry'
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    user_input_int = user_input.to_i
+    self.cells[user_input_int - 1]
+  end
+
+  def full?
+    self.cells.all?{|cell| cell == "X" || cell == "O"}
+  end
+
+  def turn_count
+    self.cells.select {|cell| cell == "X" || cell == "O"}.size
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def taken?(position)
+    self.position(position) == "X" || self.position(position) == "O"
+  end
+
+  def valid_move?(position)
+    if position.to_i.between?(1, 9)
```
weezwo: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+      !self.taken?(position)
+    else
+      false
```

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
```

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def over?
+    (self.draw? || self.won?) ? true : false
+  end
+
+  def won?
```
weezwo: You could simplify this method quite a bit! It could read:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+  def draw?
+    if self.won? || !self.board.full?
+      false
+    else
+      true
+    end
+  end
+
+  def winner
+    if self.won? != nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    self.board.turn_count % 2 == 0 ? self.player_1 : self.player_2
+  end
+
+  def over?
+    (self.draw? || self.won?) ? true : false
```

PR #1032, opened 2018-04-16
Done

ecssiah:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
+game.start
```

lib/game.rb
```diff
@@ -0,0 +1,125 @@
...
+    player_2 = Players::Human.new("O"),
+    board = Board.new
+  )
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,125 @@
...
+      player_1.token == "X" ? player_2 : player_1
+    end
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def won?
+    WIN_COMBINATIONS.find do |combo|
```
weezwo: Novel solution!! Here's another way:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,18 @@
+module Players
+
+class Human < Player
+
+  def move(board)
+    print "#{@token} choose a position: "
+    input = gets.strip
+
+    if board.valid_move?(input)
```
weezwo: Ideally, this logic would be encapsulated within the board class, and Player::Human.move would only handle the gathering of input.

PR #1033, opened 2018-04-16
Done

jojojoleeleelee:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+  def start
```
weezwo: You'll want some logic here that will kick off a game. :)

Prefer to encapsulate CLI logic into a class. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def full?
+    cells.detect {|c| c == " " || c == ""} ? false : true
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  end
+
+  def turn_count
+    cell_count = cells.select {|c| c == "X" || c == "O"}
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,53 @@
...
+    if cells[input.to_i-1] == " "
+      return false
+    else
+      return true
+    end
+  end
+
+  def valid_move?(input)
+    if input.to_i.between?(1,9) && !taken?(input)
+      return true
```

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+      (@board.cells[combo[2]] == "X" || @board.cells[combo[2]] == "O")
+    end
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    if won? != nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb

jojojoleeleelee: thanks so much for the feedback! They are tremendously helpful! Will regurgitate and implement your comments. Just wondering, is it possible to have more experts such as yourself look over my code?

PR #1034, opened 2018-04-17
Done

ajquintero:


lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    cells[input.to_i - 1]
+  end
+
+  def full?
+    if !(cells.include?(" "))
+      true
```
kkterai: Be sure to use the implicit returns of your methods when you can, rather than specifically writing `true`.

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  def turn_count
+    cells.count{|t| t == "X" || t == "O"}
```
kkterai: Good job using the most semantic method possible here, `#count`

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+    [2,4,6]
+  ]
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+  def draw?
+    @board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,118 @@
...
+      @board.cells[won?.first]
+    end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+
+  class Human < Player
+    def move(input)
+      puts "Please enter your move"
+      gets.strip
```
kkterai: Good job using the implicit return here!

PR #1035, opened 2018-04-17
Done

ritchey0713:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+new_game = Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  def position(input) 
+    cells[input.to_i-1]
+  end 
+  
+  def update(input,player)
+   cells[input.to_i-1] = player.token
+  end 
+  
+ def full?
+    if cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+      return false
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+  ]
+  
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board=Board.new)
+     @player_1 = player_1
+     @player_2 = player_2
+     @board = board
+  end 
+  
+  def current_player 
+   @board.turn_count % 2 ==0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,107 @@
...
+    WIN_COMBINATIONS.detect do |winning_combo|
+      @board.cells[winning_combo[0]] == @board.cells[winning_combo[1]] &&
+      @board.cells[winning_combo[0]] == @board.cells[winning_combo[2]] &&
+     (@board.cells[winning_combo[2]] == "X" || @board.cells[winning_combo[2]] == "O")
+    end 
+  end 
+    
+    
+  def winner
+    if won? != nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1036, opened 2018-04-17
Done

BarbaraPruz:


bin/tictactoe
```diff
@@ -1,3 +1,43 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+def get_number_players
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9,' ')
```

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+  def won?
+    # f!@*ed up test: this code will detect the 1st win on the board
+    # but test gives board with 2 wins.  the first X win is the
+    # expected result.   Why would/How can we have a board with 2 wins?
+    # WIN_COMBINATIONS.detect do | combo |
+    #  @board.cells[combo[0]]!=" " &&
+    #  @board.cells[combo[0]]==@board.cells[combo[1]]&&
+    #  @board.cells[combo[0]]==@board.cells[combo[2]]
+    #end
+    winners = WIN_COMBINATIONS.select do | combo |
```
weezwo: You might want to use [`Enumerable#detect`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-detect)  rather than `#select` here.

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+    winners.find { | win | @board.cells[win[0]]==@player_1.token} || winners.find { | win | @board.cells[win[0]]==@player_2.token}
+  end
+
+  def draw?
+    @board.full? && !won?
+  end
+
+  def winner
+    win = won?
+    win ? @board.cells[win[0]] : nil
```
weezwo: Prefer not to return nil from a ternary statement, instead relying on the truthiness of a statement.

```ruby
if won = won?
  board.cells[won.first]
end
```

BarbaraPruz:
Hi

Thanks for the feedback.  I appreciate it!

Barbara

PR #1037, opened 2018-04-18
Done

seanfebo87:


lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+  
+  def full?
+    @cells.all? do |space|
+      space != " "
+    end
+  end
+  
+  def turn_count 
+    count = 0
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+    turn
+    end
+      if won?
+        puts "Congratulations #{winner}!"
+      elsif draw?
+        puts "Cat's Game!"
+      end
+  end
+    
+  def start
```
weezwo: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+  def draw?
+    !won? && @board.full?
+  end
+  
+  def over?
+     won? || draw?
+  end
+  
+  def winner 
+    if won? == nil
```
weezwo: Avoid equality comparisons with boolean literals/nil, relying instead on the implicit return value of the boolean expression.

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+      nil
+    else
+      @board.cells[won?[0]]
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players 
+  class Human < Player 
+    
+    def move(board)
+      puts "Please enter 1-9"
+     input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1042, opened 2018-04-23
Done

xD4NNYx:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
```

lib/players/computer.rb
```diff
@@ -0,0 +1,42 @@
...
+          if cmb.select{|i| board.position(i+1) == token}.size == 2 && cmb.any?{|i| board.position(i+1) == " "}
+            move = cmb.select{|i| !board.taken?(i+1)}.first.to_i.+(1).to_s
+
+          # If you can't play any winning moves, play a move to block the opponent from winning.
+          elsif cmb.select{|i| board.position(i+1) != " " && board.position(i+1) != token}.size == 2 && cmb.any?{|i| board.position(i+1) == " "}
+            move = cmb.select{|i| !board.taken?(i+1)}.first.to_i.+(1).to_s
+          end
+        end
+
+        # If none of the WIN_COMBINATIONS patterns have two squares taken by the same token and a third empty square, play into the first open square you find, first checking corners and then checking sides.
```
howardbdev: Instead of `move == nil`, consider using Ruby's built-in methods such as `#nil?`, in this case `if move.nil?`

PR #1043, opened 2018-04-24
Done

ivillicana:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board
+  attr_accessor :cells
+
+
+  def initialize
+    self.reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move (board)
+      puts "Please enter where you would like to move (1-9):"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1044, opened 2018-04-24
Done

SaraGinsburg:


lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_accessor :cells
+  attr_reader   :player
+  # def cells
+  #   []
+  # end
+
+  def initialize
+    @cells = Array.new(9," ")
+  end
+
```
howardbdev: I would just call `self.reset!` here (in `#initialize`) and delegate to that method to create the default empty board.

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  def position(pos)
+    @cells[pos.to_i - 1]
+  end
+
+
+  def full?
+    @cells.include?(" ") ? false : true
+  end
+
```
howardbdev: For this and several other methods, consider using Ruby's implicit return rather than a ternary.  In this example, `@cells.include?(" ")` will return true or false, so it alone would be sufficient

lib/board.rb
```diff
@@ -0,0 +1,55 @@
...
+  def taken?(pos)
+    position(pos) == " " ? false : true
+  end
+
+  def valid_move?(pos)
+    taken?(pos) || pos.to_i < 1 || pos.to_i > 9  ? false : true
+  end
+
```
howardbdev: consider simplifying this conditional using Ruby's `#between` method, as well as using implicit return for the boolean:

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+   end
+
+   def over?
+     draw? || won?
+   end
+
+   def winner
+     won? ? @board.cells[won?[0]] : nil
+   end
+
```
howardbdev: Consider that the `#won?` method iterates through a collection, would it be more efficient to assign the return value to a variable, instead of calling `#won?` twice?  It would look like this:
```
winner = won?
winner ? @board.cells[winner[0]] : nil
```

PR #1045, opened 2018-04-25
Done

Pamela454:


bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+  game1.play
+
+elsif response == '2'
+  game2 = Game.new
+  puts "Who should go first and be player X?"
+    game2.play
+
+else
+  puts "Option not available."
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+  #returns turn count based on how many positions in cell array are filled
+  def turn_count(cells = @cells)
+  turn = 0
+    cells.each do |taken|
+      if taken == "X" || taken == "O"
+    turn += 1
+    end
+    end
+    turn
+  end
```
howardbdev: While this method works, consider refactoring to use the highest level and most expressive iterator possible to solve a problem. [`count`](http://ruby-doc.org/core-2.2.0/Array.html#method-i-count) on the cells is perfect and the rest of this method is not needed. Entire body could read:

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  #takes in index and returns true if value is X or O
+  def taken?(cells = @cells, location)
+    if @cells[location.to_i - 1] == "X" || @cells[location.to_i - 1] == "O"
+      return true
+    else
+      return false
+    end
+  end
```
howardbdev: Could we simplify this down to one line using Ruby's implicit return?  Also, since we have access to the board, do we need to pass `cells`?  It should work the same way with just your condition:
```
def taken?(location)
  @cells[location.to_i - 1] == "X" || @cells[location.to_i - 1] == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  #state remembered interactions
+  def current_player(board= @board)
+    if @board.turn_count % 2 == 0
+      player_1
+    else
+      player_2
+    end
+  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+
+  class Human < Player
+    def move(board)
+      return gets
+    end
+  end
```
howardbdev: Do we need the `return` keyword here?  Might be a good idea to use `gets.chomp`, too.  Is there a message to the user, something like "Where do you want to go?"

PR #1046, opened 2018-04-25
Done

Effreeman:


bin/tictactoe
```diff
@@ -1,3 +1,20 @@
...
+
+play_again = ""
+until play_again.downcase == "n" || play_again.downcase == "no"
+  puts "Would you like to play again?"
+  play_again = gets.strip
+  play_again.downcase == "y" || play_again.downcase == "yes"
+  game = Game.new
+  until game.won?
+    game.play
+  end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
+class Board
+  attr_accessor :cells
+
+  def initialize()
+    reset!
+  end
```
howardbdev: Nice use of the `#reset!` method you've already built to keep your code DRY! 👍

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,43 @@
...
+    cells[input.to_i-1] = player.token
+  end
+
+  def position(input)
+    cells[input.to_i-1]
+  end
+
+  def taken?(input)
+    !(position(input) == " " || position(input) == "")
+  end
```
howardbdev: We can simplify this method by distributing the negation. This might be a little confusing, but `!(a || b)` <=> `!a && !b` (this property is one of [DeMorgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)).

PR #1047, opened 2018-04-25
Done

katiebugsings:


lib/board.rb
```diff
@@ -0,0 +1,53 @@
+class Board
```
weezwo: It looks like you might be missing a bin file to tie all of the code together and execute it.

lib/game.rb
```diff
@@ -0,0 +1,87 @@
...
+  def draw?
+    self.board.full? && !won?
+  end
+
+  def over?
+    self.won? || self.draw?
+  end
+
+  def winner
+    if self.won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1048, opened 2018-04-25
Done

BrettSWeisberg:


bin/tictactoe
```diff
@@ -1,3 +1,26 @@
...
+else
+puts "invalid entry - Please start again."
+end
+elsif num_players.to_i == 2
+puts "only 2"
+game = Game.new(Players::Computer.new("X"),Players::Human.new("O"))
+game.play
+elsif num_player.to_i == 0
+puts "0 players"
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

Also, pay attention to your indenting.  Ruby convention is a two-space indent after any class declaration, method, or block.  You can check out this great [Ruby style guide](https://github.com/bbatsov/ruby-style-guide)!

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+  attr_accessor :cells
+
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
```
howardbdev: Consider using the `#reset!` method you already built inside `#initialize` to keep your code a little more DRY.  Also, by convention, list `#initialize` as the first method, right at the top under any `attr_accessors` and class variable declarations.  That way anyone looking at your code can immediately see what happens on instantiation of the class, without having to search for it.

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+    cells[input.to_i - 1]
+    end
+
+    def full?
+    if  cells.count("X") + cells.count("O") == 9
+           true
+        else
+          false
+        end
+      end
```
howardbdev: Instead of breaking into an `if-else`, we could just use Ruby's implicit return and have just the conditional here:
```
def full?
  cells.count("X") + cells.count("O") == 9
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def turn_count
+      return cells.count("X") + cells.count("O")
+    end
```
howardbdev: no need for the `return` keyword :)

lib/board.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def valid_move?(input)
+    #  taken?(input) == false && input.to_i < 10
+      input.to_i.between?(1,9) && !taken?(input)
+      end
+
+      def update(input, player)
+        cells[input.to_i - 1] = player.token
+      end
```
howardbdev: Again, watch your spacing and indentation ;)

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+  def initialize(player_1 = Players::Human.new("X"),player_2 = Players::Human.new("O"),board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
+
+  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+       puts "Please enter your selection."
+       input = gets.strip
+       return input
+    end
```
howardbdev: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #1049, opened 2018-04-26
Done

cpdundon:


bin/tictactoe
```diff
@@ -1,3 +1,45 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic-Tac-Toe with AI!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

The code in your bin would generally just go in `call`.Prefer t

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+class Board
+  attr_accessor :cells
+
+  def initialize
+    self.reset!
+
+  end
+
+  def reset!
+    self.cells = [" "] * 9
```
weezwo: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements.

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  def position(location)
+    c = self.cells
+
+    c[location.to_i - 1]
+  end
+
+  def turn_count
+    c = self.cells
+
+    c.select{|e| e.strip != ""}.size
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def full?
+    count = self.turn_count
```
weezwo: We can also get this value without calling #turn_count by using the [`#all?`](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-all-3F) iterator.

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def taken?(location)
+    c = self.cells
+
+    c[location.to_i - 1].strip != ""
+  end
+
+  def valid_move?(location)
+    possible = location.strip.scan(/^[1-9]$/).length > 0
```
`position.between?(1,9)`

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    return nil if !possible
```
weezwo: Avoid both returning the literal `nil` and using the return keyword. Rely on ruby's implicit return values.

lib/board.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    idx = location.strip.to_i - 1
+    c = self.cells
+    c[idx].strip == ""
```
weezwo: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value. The body of this method could read:
```
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+    self.board = board
+    self.player_1 = p1
+    self.player_2 = p2
+  end
+
+  def current_player
+    b = self.board
+    t = b.turn_count
+
+    t % 2 == 0 ? player_1 : player_2
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+  def draw?
+    self.board.turn_count > 8 && !won?
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
+    w = won?
```
weezwo: Here's a way to write this method that avoids an explicit return:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    return nil if !!! w
```
weezwo: One `!` will suffice here -- you should never need to use three.

lib/players/human.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Players
+  class Human < Player
+
+    def initialize(token)
+      super
+    end
+
+    def move(board)
+      board.display
```
```
def move(board)
  gets.chomp
end
```

PR #1050, opened 2018-04-26
Project submitted

YanezDiego:


bin/tictactoe
```diff
@@ -1,3 +1,46 @@
...
+  puts "Would you like to play again? Y/N"
+    input = gets.strip
+    if input == 'y'
+      start
+    else
+      puts "Good game Goodbye"
+    end
+  end
+
+start
```
howardbdev: Prefer to encapsulate CLI logic into a class, like what you did for your CLI portfolio project. For example, the only code in your bin executable should be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  attr_accessor :cells
+
+    def reset!
+      @cells = Array.new(9, " ")
+    end
+
+    def initialize
+      reset!
+    end
+
```
howardbdev: Good use of your `#reset!` method to keep your code DRY.  Conventionally, the `#initialize` method should be the first method listed in a class, so that anyone collaborating or looking at your code can immediately see what happens upon instantiation.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    def turn_count
+      turns = 0
+      @cells.each do |token|
+        if token == "X" || token =="O"
+          turns += 1
+        end
+      end
+      turns
+    end
```

```ruby
def turn_count
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      puts puts "Please select from 1-9"
+      input = gets.strip
+    end
```
howardbdev: Here, we could just write `gets.strip` as the last line.  No need for the `input` variable.

PR #1051, opened 2018-04-27
Done

katleiahramos:


bin/tictactoe
```diff
@@ -1,3 +1,45 @@
...
+    start
+  when 'n'
+    puts "Thanks for playing! Goodbye!"
+  else
+    new_game
+  end
+end
+
+
+start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
howardbdev: Just call `#reset!` inside `#initialize` since it's the same code :)

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+    !@cells.include?(" ")
+  end
+
+  def turn_count
+    counter = 0
+    self.cells.each do |space|
+        counter += 1 if space == "X" || space == "O"
+    end
+    counter
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def taken?(position)
+    index  = input_to_index(position)
```
howardbdev: We don't really have to create a new variable `index`, instead we can just use `input_to_index(position` in-line.

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+    WIN_COMBINATIONS.find {|combo| @board.taken?("#{combo[0]+1}") && @board.cells[combo[0]] == @board.cells[combo[1]] && @board.cells[combo[1]] == @board.cells[combo[2]] }
+  end
+
+  def draw?
+    if @board.full? && !(won?)
+      true
+    else
+      false
+    end
+  end
```
howardbdev: Avoid using `if`/`else` statements to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. 

`@board.full? && !won?`

This applies to `#over?` as well.

spec/04_game_spec.rb
```diff
@@ -110,7 +110,7 @@
       game = Game.new
       game.board.cells = ["X", "O", "X",
                           "O", "O", "X",
-                          "O", "O", "X"]
+                          "O", "X", "X"]
```
howardbdev: Good catch!

PR #1052, opened 2018-04-27
Done

camsay:


bin/tictactoe
```diff
@@ -1,3 +1,56 @@
...
+  until count == 100
+    Game.new(player_1 = Players::Computer.new("X"), player_2 = Players::Computer.new("O")).play
+    count += 1
+  end
+  Game.win_count
+end
+
+
+
+start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def self.win_count
+    x =  @@wins.select {|w| w == "X"}.count
+    o =  @@wins.select {|w| w == "O"}.count
+    puts
+    puts "X won #{x} times."
+    puts "O won #{o} times."
+    puts "#{100 - (x + o)} draws."
+  end
```
howardbdev: I like the extra step of recording wins!  Is the draws accurate?  We could add a different string each time there's a draw, like "draw", after line 68.

PR #1053, opened 2018-04-28
Done

netzay:


lib/board.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+  
+  def taken?(index)
+    if position(index) == "X" || position(index) == "O"
+      return true
+    end
+    if position(index) == " " || position(index) == ""
+      return false
+    end
+  end
```
howardbdev: We can use Ruby's implicit return to simplify this down to just one line:
```
def taken?(index)
  position(index) == "X" || position(index) == "O"
end
```

This also applies to other methods, including `#valid_move?`, `#current_player`, `#draw`, and `#over`.

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+    @player_2 = player_2
+  end
+  
+  def current_player
+    if @board.turn_count % 2 == 0 
+      return @player_1
+    else
+      return @player_2
+    end
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+      position_1 = @board.cells[win_1]
+      position_2 = @board.cells[win_2]
+      position_3 = @board.cells[win_3]
+
+    if position_1 == "X" && position_2 == "X" && position_3 == "X" || position_1 =="O" && position_2 == "O" && position_3 == "O"
+      return win_combo
+    end
+  end
+  false
+end
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+    if !won? == true && full? == true
+      return true
+    end
+    if won? == false || full? == false
+      return false
+    end
+    if won? == true 
+      return false
+    end
+  end
```
howardbdev: See comment for `#taken?`

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+  
+  def over?
+    if draw? || won?
+      return true
+    else
+      return false
+    end
+  end
```
howardbdev: See comment for `#taken?`

lib/game.rb
```diff
@@ -0,0 +1,105 @@
...
+    if won? && winner == "X"
+      puts "Congratulations X!"
+    end
+    if won? && winner == "O" 
+      puts "Congratulations O!"
+    end
+    if draw? == true
+      puts "Cat's Game!"
+    end
+  end
```
howardbdev: It's probably worth putting these together into `if..elsif...else` .. better not to break up `if-else` statements if there is only one possible thing you want to happen.

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+require "pry"
+module Players
+  attr_accessor :board, :game, :player
+  class Human < Player
+    
+    def move(board)
+      puts "Enter move:"
+      user_input = gets.chomp
+      user_input
+    end
```
howardbdev: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

netzay:
Hi Howard,

I re-scheduled my project review for Wed June 6th at 1:30 but for some
reason the app still has the red "schedule your review session" banner
up...just wanted to make sure you see my appointment :)
Thank you!

~Ally

netzay:
I just rescheduled again for the 31st I didn't realize how far away the 6th
was :) so I'll cancel the one on the 6th!
Thank you!

~Ally

PR #1055, opened 2018-04-29
Done

Arbeker01:


bin/tictactoe
```diff
@@ -1,3 +1,25 @@
 #!/usr/bin/env ruby
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

bin/tictactoe
```diff
@@ -1,3 +1,25 @@
...
+new_play = ""
+until new_play.downcase == "n" || new_play == "no"
+  puts "Would you like to play again?"
+  new_play = gets.strip
+  new_play.downcase == "y" || new_play.downcase == "yes"
+  game = Game.new
+  until game.won?
+    game.play
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,45 @@
+class Board
```
kkterai: Great job using the most expressive methods available (e.g. `count`, `between`, `all`)!

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter 1-9:"
+      gets.strip
+    end
```
kkterai: Good job using utilizing the implicit return value here!

PR #1056, opened 2018-05-01
Project submitted

nickbharucha:


lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def over?
+    draw? || won? || board.full?
+  end
+
+  def winner
+    if !won?
+      nil
+    elsif board.cells[won?.first] == player_1.token
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      input = gets
+      input
```
kkterai: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression. For example:
```
def move(board)
  gets.chomp
end
```

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+      position_1 = board.cells[combination[0]]
+      position_2 = board.cells[combination[1]]
+      position_3 = board.cells[combination[2]]
+      if position_1 == player_1.token && position_2 == player_1.token && position_3 == player_1.token
+        return combination
+      elsif position_1 == player_2.token && position_2 == player_2.token && position_3 == player_2.token
+        return combination
+      end
+    end
+    false
```
kkterai: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.

PR #1057, opened 2018-05-01
Done

dmcgrann:


bin/tictactoe
```diff
@@ -1,3 +1,42 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+class Players
+  class Human < Player
+    attr_accessor :board
+
+    def move(board)
+      gets.strip
+    end
```
howardbdev: Might be good to prompt the player, such as "Please enter your move:"

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+    board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    board.cells[won?[0]] if won?
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

dmcgrann:
Hi Howard,

Thanks for your suggestions. Looking back on this recently, I did think
that it was strange to have so much code in the bin, but I thought maybe
this was just specific to the project. So that is good to know. Likewise
with #winner, I understand why the conditional makes more sense. I have
updated everything, and again, I appreciate the feedback.

Best,
David

PR #1058, opened 2018-05-02
Done

Evbw:


lib/game.rb
```diff
@@ -0,0 +1,63 @@
+class Game
+  attr_accessor :board, :player_1, :player_2
+
```
howardbdev: Since we have an `attr_accessor` for `board`, we ought to use the getter (or setter) methods to access `@board`, rather than accessing it directly, as you've done below (other than in `#initialize`)

Evbw: Would it be best to define individual setter and getter methods? I'm afraid I don't see the difference

PR #1059, opened 2018-05-02
Done

FionaDL:


bin/tictactoe
```diff
@@ -1,3 +1,9 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require 'pry'
+
+
+  puts "Welcome to Tic-tac-toe!"
+  Game.game_assign
+  Game.play_again?
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
```
howardbdev: Why not just use the `#reset!` method you built in `#initialize` ?  Using methods like that rather than repeating code helps keep our code DRY

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  def position(x)
+   input = cells[x.to_i - 1]
+    input
+  end
+
+  def full?
+    if cells.none? {|cell| cell == " "}
+      true
+    end
+  end
```
howardbdev: Here, we can use Ruby's implicit return rather than an `if` statement.  This could be simplified to:
``` ruby
def full?
  cells.none? {|cell| cell == " "}
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+  cells.count{|token| token == "X" || token == "O"}
+end
+
+  def taken?(input)
+    if cells[input.to_i - 1] == "X" || cells[input.to_i - 1] == "O"
+      true
+    else
+      false
+    end
+  end
```
howardbdev: Here, as well as for `#valid_move?` and `Game#draw?`, again we can use implicit return for booleans rather than `if`/`if-else` statements

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      "Please enter 1-9: "
+      user_input = gets.strip
+      user_input
+    end
```
howardbdev: We don't really need to create a `user_input` variable here.  Since we're returning the user's input, we could just leave the last line of the method as `gets.strip`, as shown below.  Also, don't forget to `puts` the prompt ;)
``` ruby
def move(board)
  puts "Please enter 1-9: "
  gets.strip
end
```

PR #1060, opened 2018-05-03
Done

mjboroski:


lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+
+  class Human < Player
+
+    def move(board)
+      input=gets.chomp
+      input if board.valid_move?(input)
+    end
```
howardbdev: The if statement here seems redundant.  You're already checking for valid input in your `Game#turn` method, so here all we need to do is ask the user to input, and return the response.  This method can get called again if the input is invalid:
``` ruby 
def move(board)
  puts "Make your move!"
  gets.chomp
end
```

RUBRIC.md
```diff
@@ -1,6 +1,6 @@
...
-  - [ ] 0 player mode works
-  - [ ] 1 player mode works
-  - [ ] 2 player mode works
-  - [ ] Game is playable via CLI
-  - [ ] AI is able to play with some strategy
+  - [x] All tests pass
+  - [x] 0 player mode works
+  - [x] 1 player mode works
+  - [x] 2 player mode works
+  - [x] Game is playable via CLI
```
howardbdev: Is this game playable?  How would someone start to play?  See main comment note...

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(index)
+    @cells[index.to_i - 1]
+  end
+
+  def full?
+    @cells.include?(" ") ? false : true
+  end
```
howardbdev: Avoid using the ternary operator to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. 

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def turn_count
+    counter=0
+    @cells.each do |cell|
+      if cell=="X"||cell=="O"
+        counter+=1
+      end
+    end
+    counter
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def taken?(index)
+    @cells[index.to_i - 1]=="X" || @cells[index.to_i - 1]=="O"
+  end
+
+  def valid_move?(input)
+    index=input.to_i
+    index>=1 && index<=9 && taken?(index) == false
+  end
```
howardbdev: Use 'and' operator (`&&`) to combine multiple logical expressions and utilize their implicit return value.
```
```

PR #1061, opened 2018-05-03
Done

taylorado:


lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+      cells[(input.to_i)-1]
+    end
+
+    def turn_count
+      cells.count{|x| x == "X" || x =="O"}
+    end
+
+    def full?
+      self.turn_count == 9
+    end
```
howardbdev: This works!  Could also use `cells.all? {|c| c != " "}`

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def taken?(input)
+      (position(input) == "X" || position(input) == "O")
+    end
```
howardbdev: Don't need the parens :)

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+          end
+  end
+
+  def current_player
+    @board.turn_count.even? ? @player_1 : @player_2
+  end
+
+  def opponent
+    @board.turn_count.odd? ? @player_1 : @player_2
+  end
```
howardbdev: Do we need this method?

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+     won? || draw?
+   end
+
+   def winner
+     if !won? || draw?
+       nil
+     else
+       @board.cells[won?[0]]
+     end
+   end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      puts "Please enter 1-9:"
+      return gets.strip
+    end
```
howardbdev: no need to use the `return` keyword, Ruby convention is to avoid it unless it's explicitly needed

spec/05_computer_player_spec.rb
```diff
@@ -10,7 +10,8 @@
       computer = Players::Computer.new("X")
       board = Board.new
 
-      valid_moves = ["1", "2", "3", "4", "5", "6", "7", "8", "9"]
+      valid_moves = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+      #valid_moves = ["1", "2", "3", "4", "5", "6", "7", "8", "9"]
```
howardbdev: Instead of changing the spec file, could we adjust our code to pass the test?  It's no big deal here, but if you're working on a dev team it's best to leave the specs alone (but of course you can [and should] ask questions about them if you feel theres a better way!)

taylorado: Hi Howard:

Thanks for all the feedback!  I hesitated to change the spec file, and would not have done so within a team environment without some discussion.  Essentially, I had the whole thing working save this one error and I couldn't quite figure out why - I made this edit just to confirm my suspicion about the error.  Could certainly could have gone back and changed the code, but is there any reason the tests do not allow for the valid_moves array to be comprised of either all strings or all integers?

Thanks Again

PR #1062, opened 2018-05-03
Done

scrptktty:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+GameController.play
```
howardbdev: Good use of a separate controller class!

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def turn_count
+    turn_number = 0
+    @cells.each do |cell|
+      if cell == "X" || cell == "O"
+        turn_number += 1
+      end
+    end
+    turn_number
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,61 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def input_to_index(user_input)
+    user_input.to_i - 1
+  end
+
+  def position(user_input)
+    user_index = input_to_index(user_input)
```
howardbdev: Is there a way we can make this adjustment one time so that we don't have to call `#input_to_index` at the beginning of every method?

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+    WIN_COMBINATIONS.detect do |win_combo|
+      position1 = @board.cells[win_combo[0]]
+      position2 = @board.cells[win_combo[1]]
+      position3 = @board.cells[win_combo[2]]
+
+      if position1 != " " && position1 == position2 && position2 == position3
+        return win_combo
+      end
+    end
+  end
```
howardbdev: We aren't really using `#detect` to its full potential here.. Instead of an `if` clause with an explicit `return` command, let `#detect` implicitly return either the winning combo or `nil`:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,89 @@
...
+  def full?
+    !@board.cells.include?(" ")
+  end
+
+  def draw?
+    !won? && full?
+  end
+
+  def over?
+    won? || full? || draw?
```
howardbdev: This is a bit redundant - we only need two of these - `won? || draw?` would suffice.

spec/05_computer_player_spec.rb
```diff
@@ -17,4 +17,28 @@
...
+      board = Board.new
+      board.update("1", computer)
+      board.update("2", computer)
+
+      computer_move = computer.guess(board)
+
+      expect(computer_move).to eq(2)
+
+    end
+  end
```
howardbdev: Nice job writing your own tests!!

PR #1063, opened 2018-05-04
Done

vbustabad:


lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+  attr_accessor :cells
+
+  def reset!
+    @cells.clear
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
howardbdev: having `@cells.clear` is a bit redundant -- the second line reassigns `@cells` to the array you want, so it doesn't matter what its previous value was.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
howardbdev: Why not just use the `#reset!` method you've already built?  Using methods instead of repeating the same code helps keep our projects DRY.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+  def turn_count
+    count = 0
+    @cells.each do |token|
+      puts "The number of turns that has been played is #{count}."
+      if token == "X" || token == "O"
+        count += 1
+      end
+    end
+    count
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def taken?(input)
+    if @cells[input.to_i - 1] == "X" || @cells[input.to_i - 1] == "O"
+      return true
+    else
+      return false
+    end
+  end
```
howardbdev: Avoid using the ternary operator or `if`/`else` to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression.  This comment also applies to `#valid_move?` and `#winner`. 
``` ruby 
def taken?(input)
  @cells[input.to_i - 1] == "X" || @cells[input.to_i - 1] == "O"
end
```

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
+  end
```

PR #1064, opened 2018-05-05
Done

jimjim698:


bin/tictactoe
```diff
@@ -1,3 +1,44 @@
...
 
 require_relative '../config/environment'
+
+
+#!/usr/bin/env ruby
+
+require_relative '../config/environment'
+require 'pry'
+
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
+require 'pry'
+
+class Board
+attr_accessor :cells
+
+def initialize
+  @cells = Array.new(9," ")
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+  puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+end
+
+
+def position(input)
+  input = input.to_i-1
+  @cells[input]
+end
+
+def full?
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+  if @cells.include?(" ")
+    false
+  else
+    true
+  end
+end
+
+
+def turn_count
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+    counter
+  end
+
+  def taken?(input)
+    input = input.to_i - 1
+    @cells[input] == "X" || @cells[input] == "O"
+  end
+
+def valid_move?(input)
```
cernanb: You can use the implicit return of the boolean expressions instead of checking for equality to false. 

```
```

lib/game.rb
```diff
@@ -0,0 +1,85 @@
...
+attr_accessor :board, :player_1, :player_2
+
+def initialize(player_1 = Players::Human.new("X"), player_2= Players::Human.new("O"), board= Board.new)
+@player_1 = player_1
+@player_2 = player_2
+@board = board
+end
+
+
+def current_player
```
cernanb: You can use a ternary expression for this method
```
def current_player
  @board.turn_count.even? ? @player_1 : @player_2`
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+module Players
+
+class Human < Player
+
+  def move(board)
```
```
def move(board)
  gets.chomp
end
```

PR #1065, opened 2018-05-06
Done

Jirles:


bin/tictactoe
```diff
@@ -1,3 +1,13 @@
...
+
+puts "Welcome to TicTacToe!"
+puts "Would you to like to play in Standard Mode or Wargames Mode?"
+puts "Select standard or wargames."
+input = gets.chomp
+if input == "wargames"
+  Game.war 
+else
+  Game.start 
+end
```
howardbdev: Great bin file, short length is ideal. Also, love the flair with the wargames option!  Nice touch!  We would, however, prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,61 @@
+class Board
+
+  attr_accessor :cells
+  #data and logic of the tictactoe board, stores state of board
+  
+  def initialize
+    @cells = reset!
+  end
```
howardbdev: We could simplify to just `reset!`, as the assignment is made within that method.

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+  def won?
+    winner = nil
+    WIN_COMBINATIONS.each do |combo|
+      converted_combo = combo.collect{ |i| board.cells[i] }
+      if converted_combo.all?{|cell| cell == "X"} || converted_combo.all?{|cell| cell == "O"}
+        winner = combo 
+      end
+    end
+    winner 
+  end
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+    won? || draw?
+  end
+
+  # checks to see who is the winner, returns X, O, or nil
+  # if there is a winner, 
+  # use won? (returns the winning combo from WIN_COMBINATIONS) to index into board cells
+  # else nil 
+  def winner
+    won? ? board.cells[won?[0]] : nil 
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

spec/05_computer_player_spec.rb
```diff
@@ -17,4 +20,186 @@
...
+  describe "#opponent_token" do 
+    it "assigns the opponent's token according to the computer's token" do 
+      
+      expect(computer.opponent_token).to eq("O")
+      
+      computer2 = Players::Computer.new("O")
+      
+      expect(computer2.opponent_token).to eq("X")
+    end
+  end
```
howardbdev: Writing your own tests?  Love it!

PR #1071, opened 2018-05-10
Project submitted

yimjay:


bin/tictactoe
```diff
@@ -1,3 +1,11 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
+game.welcome
```

lib/game.rb
```diff
@@ -0,0 +1,108 @@
...
+    @board.full? && !won?
+  end
+  
+  def over?
+    draw? || won?
+  end
+  
+  def winner
+    won? ?  @board.cells[won?.first] : nil
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1072, opened 2018-05-10
Done

helenndoo:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+game = Game.new
+game.welcome
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? player_1 : player_2
+  end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Computer < Player
+
+    attr_accessor :board
+
+    def move(board)
+      "1"
+    end
```
howardbdev: Doesn't seem like the computer can do much here... how could we at least build some logic so the computer can play the game -- even if it makes bad choices for moves?

PR #1073, opened 2018-05-12
Done

ktamoguis:


bin/tictactoe
```diff
@@ -1,3 +1,12 @@
...
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
+
+#newgame = Game.new
+newgame = Game.new(Players::Computer.new("X"),Players::Computer.new("O"))
+
+newgame.board.display
+
+newgame.play
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells
+
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
howardbdev: Let's just use the `#reset` method since we've already written the code for it :)

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def reset!
+    @cells.clear
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```
howardbdev: This is redundant - setting the array to nine empty strings will reset it from whatever it was, no need to clear it first.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def position(index)
+    @cells[index.to_i - 1]
+  end
+
+  def full?
+    @cells.all? do |position|
+      position == "X" || position == "O"
+    end
+  end
```
howardbdev: When using an iterator with only one line, convention is to use curly braces.  This is just style convention, and of course will not affect performance at all:

``` ruby
@cells.all? { |position| position == "X" || position == "O" }
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def turn_count
+    @cells.count("X") + @cells.count("O")
+  end
```
howardbdev: `#count` can take a block - which means we can simplify this a bit:

`cells.count { |cell| cell == "X" || cell == "O" }`

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def taken?(index)
+    @cells[index.to_i - 1] == "X" || @cells[index.to_i - 1] == "O"
+  end
+
+  def valid_move?(index)
+    index_new = index.to_i - 1
+    index_new.between?(0,8) && !taken?(index)
+  end
```
howardbdev: No real need for the new variable here.. how about just `index.to_i.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,103 @@
...
+    @board = board
+  end
+
+  def current_player
+    if @board.turn_count % 2 == 0
+      @player_1
+    else
+      @player_2
+    end
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      if @position_1 == "O" && @position_2 == "O" && @position_3 == "O"
+        return win_combination
+      else
+        false
+      end
+    end
+    false
+  end
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def draw?
+    !won? && @board.full?
+  end
+
+  def over?
+    won? || draw? || @board.full?
+  end
```
howardbdev: This is redundant.  Just `won?  || draw?` will do.

lib/game.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def winner
+      if won? == false
+        nil
+      else
+        @position_1
+      end
+  end
```
howardbdev: Interesting use of an instance variable in @position_1... This can still be simplified to be a bit more reliable, though:

```ruby
if won = won?
  board.cells[won.first]
end
```
This will implicitly return `nil` if there is no winner.

lib/tictactoe.rb
```diff
@@ -0,0 +1,152 @@
+class  TicTacToe
```
howardbdev: Is there a reason there are two classes like this?

ktamoguis:
Hi Howard,

Saw your email but didn't get any notifications from Github.
Ok I'll take a look at my code considering your inputs.

Best,
Karl

PR #1076, opened 2018-05-13
Done

Ekelly140:


bin/tictactoe
```diff
@@ -1,3 +1,38 @@
...
+    wargame = Game.new(Players::Computer.new("X"), Players::Computer.new("O"), Board.new)
+    wargame.play
+    if wargame.won?
+      count += 1
+    end
+  end
+  puts "The game was won #{count} times!"
+else
+  puts "Please enter 0-2"
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def full?
+    !@cells.include?(" ")
+  end
+
+  def turn_count
+    turn = 0
+    @cells.each {|cell| turn += 1 if cell == "X" || cell == "O"}
+    turn
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def taken?(space)
+    input = space.to_i - 1
+    return true if cells[input] != " "
+    return false if cells[input] == " "
+  end
```
howardbdev: Use Ruby's implicit return when returning booleans instead of `if` or `if`/`else` statements.  In this case, 
``` ruby
def taken?(space)
    !(position(space) == " " || position(space) == "")
end
```
where `#position` is a method that takes care of translating the index for us..

lib/board.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def valid_move?(space)
+    input = space.to_i
+
+    if input > 0 && input < 10 && taken?(input) == false
+      true
+    else
+      false
+    end
+  end
```
howardbdev: Similar to last comment, avoid using ternary operator or `if`/`else` to return booleans.  Also, consider using `#between` ...

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+    WIN_COMBINATIONS.find do |winning_combo|
+      place_1 = board.cells[winning_combo[0]]
+      place_2 = board.cells[winning_combo[1]]
+      place_3 = board.cells[winning_combo[2]]
+
+      if place_1 != " " && place_1 == place_2 && place_2 == place_3
+        return winning_combo
+      end
+    end
+  end
```
howardbdev: We aren't using `#detect` to its full potential here... Instead:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def draw?
+    !won? && @board.full?
+  end
```
howardbdev: Better to use getter (and setter) methods that we've built instead of accessing instance variables directly.  In this case, use `board.full?`

PR #1113, opened 2018-06-02
Done

tsyaeger:


lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def valid_move?(index)
+    index.to_i.between?(1,9) && !taken?(index)
+  end
+  
+  def full?
+    ! (@cells.any?{|cell| cell == " " })
+  end
+  
+  def reset!
+    @cells = [" "] * 9
```
weezwo: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements.

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+
+
+class Board
+  
+  attr_accessor :cells
+  
+  def initialize
+    @cells = [" "] * 9
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+    loc = position(location)
+    loc == "X" || loc == "O"
+  end
+  
+  def valid_move?(index)
+    index.to_i.between?(1,9) && !taken?(index)
+  end
+  
+  def full?
+    ! (@cells.any?{|cell| cell == " " })
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  
+  def reset!
+    @cells = [" "] * 9
+  end
+  
+  def turn_count
+    moves = @cells.select{|i| i == "X" || i == "O"}
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+  
+  
+  def current_player
+    count = @board.turn_count
+    count.even? ? @player_1 : @player_2
+  end
+    
+    
+  def won?
+    WIN_COMBINATIONS.each do |combo|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+  end
+  
+  
+  def over?
+    won? || draw? ? true : false
+  end
+  
+  
+  def winner
+
```
weezwo: You could make this method more concise by utilizing the winning combo returned by `won?`:

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+      end
+    end
+    false
+  end
+  
+  
+  def draw?
+    if (won? == false && @board.full? == true)
+      puts "Cat\'s Game!"
+      true
```
weezwo: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+    else
+      false
+    end
+  end
+  
+  
+  def over?
+    won? || draw? ? true : false
```

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+      if xs > os
+        puts "Congratulations X!"
+        winner = "X"
+      else
+        puts "Congratulations O!"
+        winner = "O"
+      end
+      return winner
+    end
+    return nil
```
weezwo: You do not need to explicitly return `nil` from an `if` statement; an `if` will return `nil` if the condition fails.
```
  @board.cells[win_combo[0]]
end
```

lib/game.rb
```diff
@@ -0,0 +1,164 @@
...
+    else
+      turn
+    end
+  end
+    
+  
+  def play
+    @board.display
+
+    until over? == true
```
weezwo: Avoid equality comparisons with boolean literals, relying instead on the implicit return value of the boolean expression.

tsyaeger:
Hi Luisa. Thanks for this. I'll definitely refactor.

PR #1119, opened 2018-06-05
Done

ConorHamilton19:


bin/tictactoe
```diff
@@ -1,3 +1,49 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
+class Board 
+  attr_accessor :cells
+  
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+  
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,44 @@
...
+  def position(input)
+    @cells[input.to_i-1]
+  end
+  
+  def full?
+    @cells.none?{|c| c == " "}
+  end
+  
+  def turn_count
+   board = @cells.select{|a| a == "X" || a == "O"}
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,73 @@
...
+    def draw?
+      self.board.full? && !won?
+    end 
+    
+    def over?
+     draw? || won?
+    end 
+    
+    def winner
+      if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

You'll also want to avoid directly returning `nil` -- rely instead on the implicit return value of the expression. :)

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+class Players
+class Human < Player
+  
+  def move(board)
+    input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1120, opened 2018-06-06
Project submitted

Hector-Ariceaga:


lib/game.rb
```diff
@@ -0,0 +1,129 @@
...
+  end
+    
+  def play 
+    turn until over?
+    board.display
+    puts winner ? "Congratulations #{winner}!" : "Cat's Game!"
+  end
+  
+
+  def start
```
weezwo: Great bin file, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+require 'pry'
+class Board
+  attr_accessor :cells
+  
+  def reset!
+    self.cells = Array.new(9, " ")
+  end
+  
+  def initialize
+    self.cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    input = num.to_i - 1 
+    self.cells[input]
+  end
+  
+  def full?
+    self.cells.all? {|c| c != " " && c != ""}
+  end 
+  
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    self.cells.each {|c| if c != " " && c != ""; count +=1 end}
+    count
+  end 
+  
+  def taken?(num)
+    self.position(num) != " " && self.position(num) != ""
+  end 
+  
+  def valid_move?(num)
+    num.to_i <= 9 && num.to_i >= 1 && !taken?(num) ? true : false
```

lib/game.rb
```diff
@@ -0,0 +1,129 @@
...
+  def draw?
+    board.full? && !won?
+  end 
+  
+  def over?
+    won? || draw?
+  end
+  
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,8 @@
+module Players
+  class Human < Player
+    def move(board)
+      puts "Please enter a number (1-9):"
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1121, opened 2018-06-06
Done

daleterrell:


lib/board.rb
```diff
@@ -0,0 +1,38 @@
+class Board
+  attr_accessor :cells
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```
weezwo: Another way to instantiate an Array with `i` values is `Array.new(9, " ")` (see [Array.new](http://ruby-doc.org/core-2.2.0/Array.html#method-c-new)), where `9` is the number of elements in the array and `" "` is the default value for the elements. It's more concise and will prevent errors from repeatedly typing `" "`.Another

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

`@board.turn_count.even? ? @player_1 : @player_2`Use [`In

lib/player/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  
+class Human < Player
+  attr_accessor :board
+  
+  def move(board)
+    input=gets.strip
```
weezwo: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #1127, opened 2018-06-09
Done

enciie:


bin/tictactoe
```diff
@@ -1,3 +1,39 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,50 @@
+class Board #represents the data and logic of a Tic-tac-toe game board
+  require 'pry'
+  attr_accessor :cells
+  #property cells that stores the data of the state of the board in an array
+  @cells = []
+
+  def initialize
+    @cells = Array.new(9, " ") #or you can call reset! method
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def winner # returns token of winner
+    if winning_combo = won?
+      @winner = @board.cells[winning_combo.first]
+    end
+  end
+
+  def turn
+    puts "Please enter 1-9:"
```

lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
+module Players
+  class Human < Player #human inherits from player class
+#The human player must implement a #move method that takes in a board argument and allows a human player to enter a move via the CLI.
+#The method should return the value the user enters.
+# Even though the method accepts a board argument, it does not need to use it.
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1128, opened 2018-06-09
Done

Sillhouette:


lib/game.rb
```diff
@@ -0,0 +1,167 @@
+require 'pry'
+##
+# => Game handles the functionality of the Tic-Tac-Toe game
```
cernanb: Great bin file above, short length is ideal. However, we would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+  ##
+  def initialize
+    reset!;
+  end
+
+  ##
+  # => Creates a fresh grid for a board
+  ##
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,77 @@
...
+  ##
+  def taken?(input)
+   !(position(input) == " " || position(input) == "" )
+  end
+
+  ##
+  # => This checks whether or not a given move is valid
+  ##
+  def valid_move?(input)
+    input.to_i >= 1 && input.to_i <= 9 && !taken?(input);
```
cernanb: You can use ruby's `between?` method here
```
```

lib/game.rb
```diff
@@ -0,0 +1,167 @@
...
+      else
+        puts "Invalid input, please select a valid option.";
+        define_players;
+      end
+  end #end #define_players
+
+  ##
+  # => This figures out who's turn it is
+  ##
+  def current_player
```

lib/game.rb
```diff
@@ -0,0 +1,167 @@
...
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2;
+  end
+
+  ##
+  # => This checks if somebody has won the game or not
+  ##
+  def won?
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,167 @@
...
+  # => This checks if the game has finished in a win or a draw
+  ##
+  def over?
+    won? || draw?;
+  end
+
+  ##
+  # => This determines who the winner is based on the winning combo
+  ##
+  def winner
```
cernanb: Here is a potential refactor for this method:
```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
Sillhouette: What is the @board.taken?(combo[0] +1) doing?

I'm not sure what the reason for the +1 is...

lib/game.rb
Sillhouette: This seems to put me into an infinite loop when I use it, could be this refactor or the one for the won? method

I'm going to reach out to you on slack to talk a bit about it

PR #1129, opened 2018-06-09
Project submitted

eyosef:


lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def won?
+    WIN_COMBINATIONS.detect do |win_combination|
+      @board.cells[win_combination[0]] == @board.cells[win_combination[1]] && @board.cells[win_combination[0]] == @board.cells[win_combination[2]] && @board.cells[win_combination[1]] != " "
+    end
+  end
+
+  def winner
+  	board.cells[won?.first] if won?
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+	class Human < Player
+
+		def move(board)
+			puts "Please enter a number between 1 and 9"
+			input = gets.chomp
+		end
```
howardbdev: No need to create a variable `input` here - implicit return is enough, just `gets.chomp` for the last line in the method.

lib/board.rb
```diff
@@ -0,0 +1,66 @@
...
+		@cells[(input.to_i - 1)]
+	end
+
+	def full?
+		if @cells.include?(" ")
+			false
+		else
+			true
+		end
+	end
```
howardbdev: Avoid using the ternary operator or if/else to return true and false. Instead, rely on the implicit boolean return value of the expression.

cells.include?(" ")

This applies to `#taken?` also.  In addition, it's better practice to access instance variables and attributes through their getter and setter methods rather than directly, in this case, `cells` instead of `@cells`.

bin/tictactoe
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+Game.start
```

PR #1130, opened 2018-06-10
Done

ahotchkin:


bin/tictactoe
```diff
@@ -1,3 +1,9 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
+sleep(0.5)
+puts "How many players are there?"
+game = Game.new
+game.start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  def over?
+    self.won? || self.draw?
+  end
+
+  def winner
+    if self.won?
+      self.board.cells[self.won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,50 @@
...
+      [0,4,8], # left diagonal
+      [6,4,2] # right diagonal
+    ]
+
+    def move(board)
+      # Computer will first look to see if there any of the winning combinations can be completed on the next move.
+      # Didn't attempt to have computer choose to win vs. block. Computer will complete the first winning combination, regardless of token.
+
+      WIN_COMBINATIONS.detect do |combo|
+        # Why doesn't this work without "return"?
```
kkterai: Prefer to keep game-related logic in the Game class

kkterai:
Hi Allyson,

Great job on your TTT-with-AI project! You did a nice job of using the most semantic methods in your code and choosing the best iterators for the job. Next, I would like to see all "put's"/user-facing prompts in a CLI class maintain single-responsibility and separation of concerns. Again, good work on completing a difficult challenge. Keep crushing it!

K

lib/players/computer.rb
ahotchkin: I was trying to figure out the best way to access WIN_COMBINATIONS without adding it in the Computer class. Any suggestions on how to do that?

PR #1132, opened 2018-06-11
Done

kreopelle:


bin/tictactoe
```diff
@@ -1,3 +1,16 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+require 'pry'
+
+ TicTacToe.new
```
weezwo: Nice CLI class, although typically this will be named in a way that indicates its status as a CLI class. The logic that comes ofter this in the bin file will typically go in there, too.

PR #1133, opened 2018-06-12
Done

sweinstein27:


lib/board.rb
```diff
@@ -0,0 +1,68 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells, :board, :token
+
+  turn_count = 0
+
+  def initialize
+    self.cells = Array.new(9, " ")
```

lib/board.rb
```diff
@@ -0,0 +1,68 @@
...
+  puts " #{cells[6]}"" | ""#{cells[7]}"" | " "#{cells[8]} "
+end
+
+def position(input)
+  index = input.to_i - 1
+  cells[index]
+end
+
+def full?
+  if cells.any? { |cell| cell == " "}
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def over?
+    !!won? || draw?
+  end
+
+  def winner
+    board.cells[won?[0]] if won?
+  end
+
+  def turn
```

lib/game.rb
```diff
@@ -0,0 +1,102 @@
...
+  def draw?
+    board.full? && !won?
+  end
+
+  def over?
+    !!won? || draw?
+  end
+
+  def winner
+    board.cells[won?[0]] if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+class Human < Player
+
+  def move(board, timer = nil)
+    input = gets()
```
```
def move(board)
  gets.chomp
end
```

PR #1136, opened 2018-06-13
Done

Taylor-Williams:


bin/tictactoe
```diff
@@ -1,3 +1,14 @@
 #!/usr/bin/env ruby
-
 require_relative '../config/environment'
+
+puts "welcome to tic tac toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,59 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize (player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = nil)
+    board ? @board = board : @board = Board.new
+    @player_1 = player_1 if player_1
+    @player_2 = player_2 if player_2
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/players.rb
```diff
@@ -0,0 +1,20 @@
+module Players
+
+  class Human < Player
+    def move(board)
+      puts "give me a move (board position 1-9)"
+      position = gets.strip
```
weezwo: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.
```
def move(board)
  gets.chomp
end
```

PR #1143, opened 2018-06-16
Done

jouissances:


bin/tictactoe
```diff
@@ -1,3 +1,55 @@
...
+    game = Game.new(player_1 = Players::Computer.new("X"), player_2 = Players::Computer.new("O"))
+    100.times.collect {
+      game.play
+    }
+  else
+    start
+  end
+end
+
+start
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
+class Board
+  attr_accessor :board, :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
```
howardbdev: Get in the habit of accessing instance and class variables using getter and setter methods when you have them.  In this case, use `cells` instead of `@cells`.

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def full?
+    @cells.include?(" ") ? false : true
+  end
```
howardbdev: Avoid using the ternary operator or `if`/`else` to return `true` and `false`. Instead, rely on the implicit boolean return value of the expression. 
`def full?`
`  cells.include?(" ")`
`end`

This applies to `#taken?`, `#draw?`, `#valid_move?`, and `#over?` as well.

lib/board.rb
```diff
@@ -0,0 +1,46 @@
...
+    }
+  end
+
+  def taken?(input)
+    position(input) != " " ? true : false
+  end
+
+  def valid_move?(input)
+    taken?(input) == false && input.to_i.between?(1, 9) ? true : false
+  end
```
howardbdev: Also, no need to use the `==` against a boolean for a method that returns a boolean.  Here, `taken?(input)` is enough, no need for `taken?(input) == false`.

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+    }.compact
+
+    combo_o = board.cells.map.with_index { |o, i|
+        i if o == "O"
+    }.compact
+
+    WIN_COMBINATIONS.find { |combo|
+      (combo_x & combo).size == 3 || (combo_o & combo).size == 3
+    }
+  end
```
howardbdev: This method could be a little more efficient.  Just using `#find` (or `#detect`) will work:

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+     !won? && board.full? ? true : false
+  end
+
+  def over?
+    won? || draw? ? true : false
+  end
+
+  def winner
+    won? ? board.position(won?.last + 1) : nil
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,9 @@
+module Players
+  class Human < Player
+
+    def move(board)
+      input = gets.chomp
+    end
```
howardbdev: are you prompting the user here, such as `puts "Make your move"`?  Also, no need to assign a new variable called `input` here.  Using `gets.chomp` alone will implicitly return the user's input.

PR #1144, opened 2018-06-16
Done

nma143:


bin/tictactoe
```diff
@@ -1,3 +1,43 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+puts "Welcome to Tic Tac Toe!"
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize(cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "])
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    #@cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+    @cells = cells
+  end
+
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @board = board
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/players/human.rb
```diff
@@ -0,0 +1,13 @@
+module Players
+  class Human < Player
+    # your code here
+
+    def move(board)
+
+      puts "#{self.token}: Which cell number?"
+      cellnum = gets.strip
```
```
def move(board)
  gets.chomp
end
```

PR #1146, opened 2018-06-16
Done

iamdavidmichaelmoore:


lib/board.rb
```diff
@@ -0,0 +1,52 @@
+require 'pry'
+
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells =  Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+  def draw?
+    !board.cells.include?(" ") && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    won? ? board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,12 @@
+
+module Players
+  class Human < Player
+
+    def move(board=nil)
+      puts "Please enter a number 1-9"
+      input = gets.strip
+      input
```
```
def move(board=nil)
  puts "Please enter a number 1-9"
  gets.strip
end
```

PR #1147, opened 2018-06-18
Done

peterckim:


lib/board.rb
```diff
@@ -0,0 +1,74 @@
+class Board
+  attr_accessor :cells
+  
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+  
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+    puts "-----------"
+    puts " #{cells[6]} | #{cells[7]} | #{cells[8]} "
+  end
+  
+  def position(input)
+    cells[input.to_i - 1]
+  end
+  
+  def full?
+    cells.each do |position|
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+        
+      end
+    end
+    
+    return true
+    
+  end
+  
+  def turn_count
+    count = 0
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+    if cells[input.to_i - 1] == " " || cells[input.to_i - 1] == "" || cells[input.to_i - 1] == nil
+      return false
+    else
+      return true
+    end
+  end
+  
+  def valid_move?(input)
+    if input.to_i >= 1 && input.to_i <= 9
+      if taken?(input) == false
```
kkterai: You can simplify checking the valid integer range, and make it more semantic, by using [`Comparable#between?`](http://ruby-doc.org/core-2.2.3/Comparable.html#method-i-between-3F).
`position.between?(1,9)`

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+  def player_1
+    @player_1
+  end
+  
+  def player_2
+    @player_2
+  end
+  
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+      return @player_1
+    else
+      return @player_2
+    end
+  end
+  
+  def won?
+    WIN_COMBINATIONS.each do |combination|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,142 @@
...
+      return true
+    elsif won?
+      return true
+    else
+      return false
+    end
+  end
+  
+  def winner
+    if won?
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/players/human.rb
```diff
@@ -0,0 +1,10 @@
+module Players
+  
+  class Human < Player
+    def move(board)
+      input = gets.strip
```
```
def move(board)
  gets.strip
end
```

PR #1151, opened 2018-06-20
Done

cmlugoce:


lib/board.rb
```diff
@@ -0,0 +1,52 @@
+class Board 
+  
+attr_accessor :cells
+  
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/board.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+  
+  def reset!
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+
+ def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new )
+   @board = board
+   @player_1 = player_1
+   @player_2 = player_2
+   
+  end 
+  
+  def current_player
+    if @board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+      return @player_1
+    else
+      return @player_2
+    end
+  end
+  
+  def won?
+    WIN_COMBINATIONS.each do |i|
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,79 @@
...
+   def draw?
+    @board.full? && !won?
+  end 
+  
+  def over?
+    won? || draw?
+  end
+  
+  def winner
+    won? ? @board.cells[won?[0]] : nil
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1157, opened 2018-06-20
Project submitted

donnovan-lyons:


bin/tictactoe
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+Game.start
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def reset!
+    # can reset the state of the cells in the board
+    # sets the cells of the board to a 9 element array of " "
+    @cells = Array.new(9, " ")
+  end
+  def initialize
+    # sets the cells of the board to a 9 element array of " "
+    @cells = Array.new(9, " ")
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  def turn_count
+    # returns the amount of turns based on cell value
+    count = 0
+    @cells.each do |play|
+      if play == "O" || play == "X"
+        count +=1
+      end
+    end
+    count
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+  def position(index)
+    # takes in user input and returns the value of the board cell
+    idx = input_to_index(index)
+    @cells[idx]
+  end
```
howardbdev: It's good practice to use getter and setter methods to access instance and class variables when we have them.  In this case, we have `attr_accessor :cells` which gives us a getter and setter for the instance variable `@cells`.  In this case, and throughout the rest of the class, we can just reference the getter `#cells`, such as 
```
    idx = input_to_index(index)
    self.cells[idx]
```
or 
```
    idx = input_to_index(index)
    cells[idx]
```
The second way works since `self` is the implied receiver.

lib/board.rb
```diff
@@ -0,0 +1,69 @@
...
+  def taken?(index)
+    # returns true if the position is X or O
+    # returns false if the position is empty or blank
+    idx = input_to_index(index)
+    if @cells[idx] == "O" || @cells[idx] == "X"
+      true
+    else
+      false
+    end
+  end
```
howardbdev: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.  This applies to other methods as well, including `#valid_move?`, `Game#over?`, and `Game#draw?`.

lib/game.rb
```diff
@@ -0,0 +1,249 @@
...
+end
+
+  def current_player
+    # returns the correct player, X, for the third move
+    if turn_count.even?
+      player_1
+    else
+      player_2
+    end
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,249 @@
...
+      else
+        false
+      end
+    end
+    if winner != nil
+      winner
+    else
+      false
+    end
+  end
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,249 @@
...
+    # returns nil when no winner
+    WIN_COMBINATIONS.each do |combo|
+      if combo.all? {|idx| cells[idx] == "X"}
+        return "X"
+      elsif combo.all? {|idx| cells[idx] == "O"}
+        return "O"
+      end
+    end
+    nil
+  end
```
howardbdev: Use the `#won?` method you've already built.
```ruby
if won = won?
  board.cells[won.first]
end
```

howardbdev (commented):
Hey Donnovan!

Great job on your Tic Tac Toe with AI Project! I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can always email or slack me.

Regards,

Howard

PR #1158, opened 2018-06-21
Done

PPzhen000:


lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      turn
+    end
+    if won?
+      puts "Congratulations #{winner}!"
+    elsif draw?
+      puts "Cat's Game!"
+    end
+  end
+
+end
```
howardbdev: Looks like there's no executable to actually run this game.  Right now, although the tests can use your code, how would you actually run this from the terminal?  Create an executable file in your bin folder, something like `clitictactoe`, and it would only take a few lines of code to get it going:
```
 #!/usr/bin/env ruby
CLI.new.call
```
Where `CLI` is a new class responsible for input and output from a user.  Its responsibilities would include greeting a player, determine how many players and instantiate a new game, (once started, the game play would yield to the Game class), asking a user if another game is desired, and saying 'goodbye' when a user decides not to play again.

lib/board.rb
```diff
@@ -0,0 +1,57 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " ", " ", " ", " ", " ", " ", " "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+    @cells.all? {|ele| ele != " "}
+  end
+
+  def turn_count
+    counter = 9
+    @cells.each do |ele|
+      counter -= 1 if ele == " "
+    end
+    counter
+  end
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def taken?(input)
+    if (position(input) == " " || position(input) == nil)
+       false
+    else
+      true
+    end
+  end
```
howardbdev: Avoid returning boolean literals from `if` statements, relying instead on the implicit return value of the boolean expression.  This also applies to `#valid_move?`.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @board = board
+  end
+
+  def current_player
+    if board.turn_count % 2 == 0
+      player_1
+    elsif board.turn_count % 2 == 1
+      player_2
+    end
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    end
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1159, opened 2018-06-21
Done

hillevieklow:


bin/tictactoe
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+Game.new.start
```

lib/board.rb
```diff
@@ -0,0 +1,48 @@
...
+    @cells[input.to_i - 1]
+  end
+
+  def input_to_index(input)
+    input.to_i - 1
+  end
+
+  def turn_count
+    count = 9 - @cells.count(" ")
+  end
```
howardbdev: No need to declare the `count` variable here.

lib/game.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  end
+
+  def current_player
+    if board.turn_count.even?
+      @player_1
+    else
+      @player_2
+    end
+  end
```

PR #1160, opened 2018-06-22
Done

jedpeek:


bin/tictactoe
```diff
@@ -1,3 +1,23 @@
...
+  game.board.display
+  game.play
+elsif input == "0"
+  game = Game.new(Players::Computer.new("X"), Players::Computer.new("O"), Board.new)
+  game.board.display
+  game.play
+else
+  puts "Sorry, please enter the number of players. 1, 2, or 0 for AI"
+  input = gets.strip
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
+class Board
+attr_accessor :cells
+
+  def initialize
+    @cells = [" ", " ", " "," ", " ", " "," ", " ", " "]
+  end
```

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def cells
+    @cells
+  end
```
howardbdev: You already have this method thanks to `attr_accessor`.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+    self.cells[user_input.to_i - 1] != " " ? true : false
+  end
+
+  def valid_move?(user_input)
+    if user_input.to_i < 1 || user_input.to_i > 9
+      false
+    else
+   !taken?(user_input)
+    end
+  end
```
howardbdev: Avoid returning boolean literals from `if` statements or ternaries, relying instead on the implicit return value of the boolean expression.  This also applies to `#taken?`, `Game#full?` and `Game#over?`.

lib/board.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def update(user_input, player)
+  self.cells[user_input.to_i - 1] = player.token if  valid_move?(user_input)
+  end
```
howardbdev: Be careful to indent properly.  Seems like a little thing but it makes code much more readable.

lib/game.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def current_player
+    count = 0
+    self.board.cells.each do |board|
+      if board != " "
+        count += 1
+      end
+      end
+      @board.turn_count % 2 == 0 ? self.player_1 : self.player_2
+  end
```

lib/game.rb
```diff
@@ -0,0 +1,112 @@
...
+    pos_1 = self.board.cells[index_1]
+    pos_2 = self.board.cells[index_2]
+    pos_3 = self.board.cells[index_3]
+
+    if (pos_1 == "X" && pos_2 == "X" && pos_3 == "X") || (pos_1 == "O" && pos_2 == "O" && pos_3 == "O")
+      return win_combo
+    end
+    end
+    return false
+  end
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,112 @@
...
+  if won? != false
+    return true
+  end
+return draw?
+end
+
+def winner
+  win_combo = won?
+  won? ? self.board.cells[win_combo[0]] : nil
+end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,112 @@
...
+      @board
+    end
+
+    def player_1
+      @player_1
+    end
+    def player_2
+      @player_2
+    end
+
```
howardbdev: You already have these three getter methods thanks to attr_accessor.

howardbdev (commented):
Hi Jed,

Great job on your Tic Tac Toe with AI Project! I did a code review on Github, and left some comments for you. You should have received some notifications from Github. Feel free to go back through your code and update it using my comments. This is not required. If you have any questions you can contact me by email: howard@flatironschool.com or on slack: (@howard).

Regards,

Howard

PR #1161, opened 2018-06-23
Project submitted

GabrielValle91:


lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def position(input)
+    @cells[input.to_i - 1]
+  end
+
+  def full?
+    !@cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+    #binding.pry
+    !(self.position(position) == " ")
+  end
+
+  def valid_move?(position)
+    position.to_i.between?(1,9) && !self.taken?(position.to_i)
+  end
+
+  def turn_count
+    @cells.select {|c| c == "X" || c == "O"}.count
```

```ruby
def turn_count
end
```

lib/game.rb
```diff
@@ -0,0 +1,72 @@
...
+  def over?
+    self.draw? || self.won?
+  end
+
+  def winner
+    if !self.won?
+      return nil
+    else
+      return @board.cells[self.won?[0]]
+    end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/board.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def update(position, player)
+    position_adj = (position.to_i - 1)
+    @cells[position_adj] = player.token
+  end
+
+  def reset!
+    @cells.clear
+    @cells = Array.new(9, " ")
+  end
```
kkterai: `@cells.clear` is not needed since `@cells = Array.new(9, " ")` reassigns the value of `@cells`

lib/board.rb
```diff
@@ -0,0 +1,47 @@
+class Board
+
+  attr_accessor :cells
+
+  def initialize
+    @cells = Array.new(9, " ")
```

kkterai:
Good job in implementing model classes for this project! 

To get your TTT w/ AI program fully functional, you'll need to implement code in your bin executable which will start the program and encapsulate CLI logic into a class. For example, the code in your bin executable could be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

PR #1162, opened 2018-06-24
Done

pracniscate:


lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+  attr_accessor :board, :player_1, :player_2
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    @board.turn_count % 2 == 0 ? @player_1 : @player_2
```

lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def draw?
+    !won? && board.full?
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
```
kkterai: Nice job calling `Game#won?` once!

lib/players/computer.rb
```diff
@@ -0,0 +1,25 @@
...
+    def open_positions(board)
+      @open = []
+      board.cells.each_with_index { |cell, index| @open << (index + 1).to_s if cell == " "}
+      @open
+    end
+
+    def move(board)
+      positions = open_positions(board)
+      @move = positions[0]
+      @move
```
kkterai: You don't need to store your return value in a local variable; just utilize the implicit return value of the expression.

PR #1163, opened 2018-06-25
Done

nateshuster:


bin/tictactoe
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
+shu_game = Game.new
+shu_game.start_game
```
kkterai: How could you refactor these two lines of code into one?

lib/board.rb
```diff
@@ -0,0 +1,48 @@
+require 'pry'
+
+class Board
+  attr_accessor :cells
+  
+  def initialize
+    @cells = Array.new(9, " ")
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+    ]
+    
+    def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+      @player_1 = player_1
+      @player_2 = player_2
+      @board = board
+    end
+    
+    def current_player
+      if board.turn_count % 2 == 0
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+    
+    def won?
+      WIN_COMBINATIONS.each do |winning_line|
+        if board.cells[winning_line[0]] == "X" && board.cells[winning_line[1]] == "X" && board.cells[winning_line[2]] == "X"
+          return [winning_line[0], winning_line[1], winning_line[2]]
+          elsif board.cells[winning_line[0]] == "O" && board.cells[winning_line[1]] == "O" && board.cells[winning_line[2]] == "O"
+          return [winning_line[0], winning_line[1], winning_line[2]]
+        end
+      end
+      false
```

```ruby
  def won?
      @board.taken?(combo[0] + 1)
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+    end
+    
+    def over?
+      draw? || won?
+    end
+    
+    def winner
+      if won?
+        board.cells[won?[0]]
+      end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

lib/game.rb
```diff
@@ -0,0 +1,101 @@
...
+      
+      if won?
+        puts "Congratulations #{winner}!"
+      else
+        puts "Cat's Game!"
+      end
+    end
+    
+  def start_game
```
kkterai: We would prefer to see the CLI logic separated from the game logic in its own class (e.g. TicTacToeCLI). Splitting the interface from the implementation of the game allows for multiple interfaces (e.g. a web app) without requiring changes to the game code.

So for example, you could encapsulate CLI logic into a class and the only code in your bin executable would be:

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.start_game
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def start_game
   # Your game logic
  end
end
```

PR #1164, opened 2018-06-25
Done

DeedaG:


bin/tictactoe
```diff
@@ -1,3 +1,35 @@
...
+  Game.new(Player::Human.new("X"), Player::Human.new("O"),Board.new)
+else
+end
+
+
+
+
+
+  puts "Do you want to play again?"
+  play_again = gets.strip
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
+require 'pry'
+
+class Board
+
+attr_accessor :cells
+
+@cells = []
```
kkterai: Initializing `@cells` outside of an instance method context makes this a 'class instance variable' instead of a typical instance variable, which is probably not what you intended here. Class instance variables are only accessible through class methods. Instead, initialize your variable in the `initialize` or `reset` method.

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def initialize
+    reset!
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def position(user_input)
+    index = input_to_index(user_input)
+    @cells[index]
+  end
+
+  def full?
+     !@cells.include?(" ")
+  end
```

PR #1167, opened 2018-06-27
Done

sentenel:


lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def initialize(player_1 = Players::Human.new("X"), player_2 = Players::Human.new("O"), board = Board.new)
+    @player_1 = player_1
+    @player_2 = player_2
+    @board = board
+  end
+
+  def current_player
+    board.turn_count % 2 == 0 ? player_1 : player_2
+  end
```

bin/tictactoe
```diff
@@ -1,3 +1,46 @@
...
+    puts "This feature is not yet implemented."
+    puts "How about a nice game of Tic-Tac-Toe?"
+  when "global thermonuclear war"
+    puts "I think you have me confused with someone else."
+  when "q", "quit", "exit"
+    break
+  else
+    puts "I did not understand."
+  end
+end
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+    board.full? && !won?
+  end
+
+  def over?
+    won? || draw?
+  end
+
+  def winner
+    won? ? board.position(won?[0] + 1) : nil
+  end
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1181, opened 2018-07-02
Done

alexandrafren:


bin/tictactoe
```diff
@@ -1,3 +1,22 @@
 #!/usr/bin/env ruby
 
 require_relative '../config/environment'
+
```

```ruby
#!/usr/bin/env ruby
       
TicTacToeCLI.new.call
```

```ruby
class TicTacToeCLI
  def initialize
  end

  def call
   # Your game logic
  end
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def display
+    puts " #{@cells[0]} | #{@cells[1]} | #{@cells[2]} "
+    puts "-----------"
+    puts " #{@cells[3]} | #{@cells[4]} | #{@cells[5]} "
+    puts "-----------"
+    puts " #{@cells[6]} | #{@cells[7]} | #{@cells[8]} "
+  end
+
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
+class Board
+  attr_accessor :cells, :player
+
+  def initialize
+    @cells = [" "," "," "," "," "," "," "," "," "]
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  def reset!
+    @cells = [" "," "," "," "," "," "," "," "," "]
+  end
+
+  def position(space)
+    @cells[space.to_i - 1]
+  end
+
+  def full?
+   !@cells.include?(" ")
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def turn_count
+    counter = 0
+    @cells.each do |cell|
+      if cell != " "
+        counter += 1
+      end
+    end
+    counter
```

```ruby
def turn_count
end
```

lib/board.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def taken?(space)
+    position(space) != " "
+  end
+
+  def valid_move?(space)
+    space.to_i < 10 && space.to_i > 0 && taken?(space) == false
```
```
```

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+      @player_2
+    else
+      @player_1
+    end
+  end
+
+  def won?
+    WIN_COMBINATIONS.detect do |combo|
+      board.taken?(combo[1]+1) && board.cells[combo[0]] == board.cells[combo[1]] && board.cells[combo[1]] == board.cells[combo[2]]
+    end
```
kkterai: Nice job using `detect` in this method!

lib/game.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def over?
+    draw? || won?
+  end
+
+  def winner
+    if won?
+      board.cells[won?[0]]
+    else
```

```ruby
if won = won?
  board.cells[won.first]
end
```

PR #1182, opened 2018-07-02
Project submitted

haydenbetts:


lib/players/human.rb
```diff
@@ -0,0 +1,11 @@
...
+
+module Players
+    class Human < Player
+        def move(board)
+            puts "Enter your move: (1-9):"
+            player_move = gets.strip
+            return player_move
+        end
+    end
+end
```
```
def move(board)
  puts "Enter your move: (1-9):"
  gets.strip
end
```

lib/players/computer.rb
```diff
@@ -0,0 +1,20 @@
+require 'pry'
+
+module Players
+    class Computer < Player
+        # take the first available position
+        def move(board)
+            i = 0
+            player_move = ""
+
```
kkterai: What other logic might you add here for the AI?

